PR #1, opened 2017-07-12
Update DimensionDialog.cpp

borgdog:
clarification/corrections:

rearranged some to have in chronological order
split gottlieb em to reflect change in 76 (with overlap) from 41" to 42" 
removed wpc (through 1987) as wpc did not exist before 1990
added Williams EM (standard) to replace the wpc through 1987
removed Genie as own entry added Gottlieb System1 (widebody) instead as there was more than one
added WMS in front of the WPC entries as WMS was the company WPC was the processor


toxieainc:
Was already made in the old SVN

PR #3, opened 2021-03-14
7.1 Surround Sound Feedback (enhanced) mode

nkissebe:
Add a 7.1 SurroundSound mode tuned for SSF cabinet configurations.
This exaggerates the positional feel of the playfield sound effects when
played in a cabinet with exciter pairs positioned at each end of the cabinet.


toxieainc:
Great, thanks a lot!

PR #8, opened 2021-03-20
Inline convert mono .WAV samples to 2-channels when needed

nkissebe:
If running a system in 7.1 surround sound mode and play a mono sample via DirectSound, Windows will play the sample on the center surround sound channel. Most cabinets don't have this hooked up. This comes up from time to time and the usual fix is to use EqualizerAPO to redirect the center channel to the left/right speakers. We can avoid this situation for .WAV file samples (ie, when moved from playfield (3D) to backglass (2D)) by duplicating the mono channel and creating a 2 channel sample which will then get played on the left/right speakers. This isn't always a problem and seems to come and go as drivers and versions of windows update. Seems better to get it right inline than rely on an external fix. I think I avoid any possibility of the original sample getting corrupted or trying to convert an unsupported sample type. Its possible that someone with real center speaker would have preferred the original behavior, but I think that would be an extreme minority.  This only effects WAV samples. Other formats remain unchanged with the idea they are working as intended.


toxieainc:
Neat! Thanks..

PR #12, opened 2021-04-03
10.6.0

Ciceronic:
Idea: Make a "Quick-POV-export" function with assigned shortcut. That saves a POV-file with the exact name of the active table, no popups, no questions, no nothing. Like Ctrl+Shift+P or similar.


toxieainc:
There is already the command line option
-Pov [filename]         Load, export pov and close

Closing the PR as the actual pull is nonsense

Ciceronic:
Ok, sorry. My knowledge are some lightyears behind you.
I just open VPX file, adjust camera manually, and Export POV-file.

Anyway, thanks a lot for your impressive work!

toxieainc:
Ah, okay, for this workflow there is even something fancier now, just use on the command line:
-PovEdit [filename]     Load and run file in camera mode, then export new pov on exit

(exit via Start Key (save POV and quit) and Credit Key (only quit without saving))

You need a recent build like https://github.com/vpinball/vpinball/actions/runs/716328097 though

PR #17, opened 2021-07-14
added lower case filtering to layers. added checkbox for users to toggle case filtering. added mouse position converted units status.

horseyhorsey:


toxieainc:
Thanks!

PR #18, opened 2021-07-16
changed `SeqBlink` to accept a `TailLength` this fixes the sequencer …

horseyhorsey:
…getting stuck if value greater than 0 is sent. Any value higher than 0 will `Reverse`, where the lights will turn `off/on` instead of default `on/off`.


toxieainc:
Thanks!

PR #50, opened 2022-03-26
Additive blending for primitives, in play intensity for lights

vbousquet:
This PR adds the same options we have for flashers regarding additive blending to primitives. That is to say the ability to render primitive with depth test but without depth write, modulating their color and alpha channel, and performing additive blending.

The use case for this is the same as the one of the flashers: they are frequently use to apply baked lighting to flat surfaces (for example the lighting of a real flasher to the side blades). With this feature, this allows to apply these baked lighting to arbitrary surface (not only flat) for example directly on the table geometry.

The PR also contains the ability to query the 'in play' intensity of a light. The aim is the same as the 'in play state' added in 10.6: give acces and use VPX fader instead of creating a duplicated one in script.

This PR has been prepared a few monthes ago and tested since in a variety of situations by a variety of users. So I hope it is clean and safe to merge.

If you like the feature and merge it, I can also prepare the corresponding PR for VPVR (and organize the testing with the VPW team).

Here is a screenshot of the added properties at the bottom of the property panel of primitive properties:
image


primitive.cpp
```diff
@@ -414,6 +418,10 @@ void Primitive::WriteRegDefaults()
    SaveValueFloat(strKeyName, "Friction", m_d.m_friction);
    SaveValueFloat(strKeyName, "Scatter", m_d.m_scatter);
 
+   SaveValueBool(strKeyName, "AddBlend", m_d.m_addBlend);
+   SaveValueFloat(strKeyName, "Alpha", m_d.m_alpha);
```
toxieainc, line 422: The Load above loads 'Opacity', here it's 'Alpha'.

primitive.cpp
```diff
@@ -1746,6 +1771,8 @@ bool Primitive::LoadToken(const int id, BiffReader * const pbr)
 #endif
    case FID(PIDB): pbr->GetFloat(m_d.m_depthBias); break;
    case FID(OSNM): pbr->GetBool(m_d.m_objectSpaceNormalMap); break;
+   case FID(ADDB): pbr->GetBool(m_d.m_addBlend); break;
+   case FID(FALP): pbr->GetFloat(m_d.m_alpha); break;
```
toxieainc, line 1775: Loading of COLR is missing

toxieainc (commented):
Apart from the 2 mentioned details, i think it would also be better to just 'restore' SetFlasherCoreAlpha after changing it from 1,1,1,1 inside of primitive.cpp only, instead of setting it everywhere else.
Apart from that, everything looks cool, thanks!!

vbousquet:
The last PR fix these stupid bugs and perform uniform restore instead of setting it each time. I also had to set its default value after shader creation (in RenderDevice.cpp)

toxieainc:
Neat, thanks again!

PR #52, opened 2022-03-29
Reformat source with clang-format

jsm174:
This PR might be controversial but was hoping maybe it could drive some conversation.

I've been very interested in trying to make a port of VPX for a platform other than Windows. I know it is a crazy idea, considering how locked into Windows this is.  A few months ago I started VPXPP and made some good progress until it was time to actually render a table. Porting DX9 was way more challenging than I had anticipated.  

After finding out about and studying VPVR, I would like to take another attempt at this. 

Before that, I would like to help clean up and re-organizing code. A lot of time spent in VPXPP was trying to decode macros, determine classes from different named files, understanding single line `if` statements without braces intermixed in `#ifdef`s.

Since there was already a `.clang-format` file, I took the one from XBMC/Kodi (considering it is a well established C++ project) and made the following changes:

```
DerivePointerAlignment: false
SortIncludes: false
DeriveLineEnding: false
UseCRLF: false
```

`clang-format` was then run on all `cpp/h` files excluding the `inc` and `imgui` folders. 

`clang-format` version 15 was then run on the same files.  (version 15 has a new `InsertBraces` option)

The build from the CI was then tested, and no issues were noticed.

I realize these commits create massive diffs, but think it would be good for the project. If accepted, I can perform the same process on VPVR. 

(I have some more future updates planned, like fixing filenames' case, splitting out files with multiple classes, etc.)


toxieainc:
As much as i appreciate all what you do, this one i really want to deny. Just commiting global formatting changes is really really bad IMHO, and we also do avoid this in all of the professional projects i work on for various reasons.
Adapting formatting while really changing a file (features, fixes, etc) is perfectly fine IMHO, but everything else is just weird IMHO.

fuzzelhjb:
I'm with Toxie here....just reformating everything is something we shouldn't do. Especially the external libraries like win32xx and imgui. Win32xx has some changes we need in VPX that are not part of the original library project. If you reformat the code you can't tell what changes we made anymore if a new version of Win32xx comes out.
I added the .clang-format file as a template to reformat the code when you are developing/fixing something. So maybe we can reformat just the core VPX code and leave the rest untouched?

jsm174:
Understood. I appreciate the comments and figured this was a long shot! I had to try.
 
To address some comments:

> Just commiting global formatting changes is really really bad IMHO

Agreed.  I looked at other large projects, ex: https://wiki.blender.org/wiki/Tools/ClangFormat  
If clang-format didn't work so well, I wouldn't even have tried this. 

> Adapting formatting while really changing a file (features, fixes, etc) is perfectly fine IMHO,

Agreed. In a project this size, there will be files that probably will never get touched. I don't know how we could hit the entire codebase in a realistic amount of time.  

> So maybe we can reformat just the core VPX code and leave the rest untouched?

That's exactly what this PR did. It did not touch anything in `inc` or `imgui`:

```
find -E . -regex '.*\.(cpp|h)' -not -path "./inc/*" -not -path "./imgui/*" | sort > format.sh
```

jsm174:
Thank you for the comments. I really appreciate it!  TBH, I'm embarrassed as I thought I had considered all the negatives of what I proposed.  Tracking history via blame never crossed my mind as I've only ever used that a handful of times.  I'm also not a fan of code formatters -- I had it stuck in my head that a one-time level set for the code base would be good. 

That said, I agree with everything above! I am going to focus on VPVR and see what I can come up with. I definitely need a different strategy than `vpxpp` as it's just not maintainable being so different.

PR #53, opened 2022-04-09
Support HDR (.HDR/.EXR) image for primitive like VPX does for flashers

vbousquet:
This little PR just allows to use .hdr/.exr image for primitive. The use case is the same than for flashers (baked light effects with a range above 1).

I did test it with a few tables and did not notice any regression.

I also checked that the result is the intended one, meaning primitive behaving the same as flasher. Here is a screenshot of the comparison table (bottom are flashers, top are primitives with different comparison textures):
image

The test table for the above screenshot:
[HDR Prim Test.zip](https://github.com/vpinball/vpinball/files/8457244/HDR.Prim.Test.zip)

### DIFF
--- a/Shader.h
+++ b/Shader.h
@@ -30,7 +30,7 @@ enum shaderUniforms {
    SHADER_fDisableLighting_top_below, SHADER_backBoxSize, SHADER_quadOffsetScale, SHADER_quadOffsetScaleTex, SHADER_vColor_Intensity, SHADER_w_h_height, SHADER_alphaTestValueAB_filterMode_addBlend,
    SHADER_amount_blend_modulate_vs_add_hdrTexture01, SHADER_staticColor_Alpha, SHADER_width_height_rotated_flipLR, SHADER_vRes_Alpha_time, SHADER_mirrorFactor, SHADER_SSR_bumpHeight_fresnelRefl_scale_FS, SHADER_AO_scale_timeblur,
    //Integer and Bool
-   SHADER_ignoreStereo, SHADER_SRGBTexture, SHADER_hdrTexture0, SHADER_disableLighting, SHADER_lightSources, SHADER_doNormalMapping, SHADER_hdrEnvTextures, SHADER_is_metal, SHADER_color_grade, SHADER_do_bloom, SHADER_lightingOff, SHADER_objectSpaceNormalMap, SHADER_do_dither,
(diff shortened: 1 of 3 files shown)


toxieainc:
Thanks! What i don't get though is why you need the invgamma in the shader? This should actually lead to the wrong gamma being used in the end.
For me (NVIDIA GPU) it works as intended out-of-the-box (i.e. not needed).
Are you on Intel or AMD?

toxieainc:
If it actually makes a difference then i would rather suggest to do a `SetSamplerState(texUnit, D3DSAMP_SRGBTEXTURE, 0)` if it is a HDR texture, then afterwards restore it via `SetSamplerState(texUnit, D3DSAMP_SRGBTEXTURE, 1)`

vbousquet:
I put the InvGamma since this is what is used for flashers:

in flasher.cpp:
`const vec4 ab(..., hdrTex0 ? 1.f : 0.f, ...);
pd3dDevice->flasherShader->SetVector(SHADER_amount_blend_modulate_vs_add_hdrTexture01, &ab);`

And in flasher.hlsl:
`[branch] if ((flasherMode < 2.) && (amount_blend_modulate_vs_add_hdrTexture01.z == 0.)) // Mode 0 & 1
      pixel1.xyz = InvGamma(pixel1.xyz);`

Therefore, I thought that would be more logical to just do exactly the same for consistency between the render paths. I agree with you that it works as intended without the InvGamma on my NVidia card as well. If it's ok for you, I can propose a PR which just allows setting a HDR/EXR image on primitive without the shader changes (but the rendering should be slightly different from flashers as far as I understand, though, I just made a quick try and it is difficult to spot).

vbousquet:
Made more testing and checking and I was wrong. Sorry for that. The render path for flasher apply inverse gamma transform for **non HDR** textures. To be honest, I do not understand why. I will close this and propose a simple PR just allowing to set an HDR/EXR texture for primitive in the editor without touching the shader.

toxieainc:
Yes, i know. The whole non-HDR vs HDR texturing is still partially a mess.
'Back in the days', when not all cards or drivers supported sRGB (inv)gamma via DX9, i decided to do it all manually in the shader instead of the texture sampler. I corrected this by now for the basic/general shader, but for the rest its still like before.  :/
One of the many things on my todo list.

PR #54, opened 2022-04-09
Allow HDR/EXR texture for primitives

vbousquet:
This is a follow up to PR https://github.com/vpinball/vpinball/pull/53
It simply allows to set an HDR texture to a primitive.

### DIFF
--- a/primitive.cpp
+++ b/primitive.cpp
@@ -2063,11 +2063,6 @@ STDMETHODIMP Primitive::put_Image(BSTR newVal)
    char szImage[MAXTOKEN];
    WideCharToMultiByteNull(CP_ACP, 0, newVal, -1, szImage, MAXTOKEN, nullptr, nullptr);
    const Texture * const tex = m_ptable->GetImage(szImage);
-   if (tex && tex->IsHDR())
-   {
-       ShowError("Cannot use a HDR image (.exr/.hdr) here");
-       return E_FAIL;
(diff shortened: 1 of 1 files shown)


toxieainc:
Thanks! Note though that i'll have to insert some additional code to prevent 'abuse', i.e. to only allow HDR values >1 if it is used for blending. Otherwise all kinds of nonsense may happen.

toxieainc:
Can you please review my last commit? It clamps the texture input to 0..1, if lighting was not disabled completely.
Otherwise all kinds of nonsense may happen in the lighting code.

Is this okay? Cause i would assume that these new alpha blended primitives would always have their lighting disabled anyway, right?

vbousquet:
Your last commi seems perfectly clean to me. Beside, I'm not sure it is needed; you are adding some (very little) load on the pixel shader for something which is unlikely. Personally, I would just leave it as is, without the saturate.

Perhaps a way to have all these more clean, would be to add an 'unshaded' material (like we have default/metallic) with its own simple technique. If you think this would be interesting, I can propose a PR for this.

toxieainc:
The thing is less that its unlikely, but rather that artists/authors usually tend to abuse everything that is possible.
And to be able to progress forward and being able to improve VPX, i tend to disallow at least those obvious hacks that will cause trouble later-on.  ;)

PR #55, opened 2022-04-14
Allow playfield_mesh for visuals only

vbousquet:
This PR allows to split the visual part of the playfield from its physics part, like it is possible for all other objects. The use case for this is when the visuals for the playfield are not fitted for physics simulation (too high poly, cutholes for all the switches or even GI lights, some beveling,...). In this case, with this PR you would use a playfield_mesh like today (needed to get correct reflections) but disable its physics (mark it as a 'toy'), and add another collidable mesh with only the needed holes by the physics, or if not needed a full PF sized ramp without any hole.

To do so, it takes in account the 'toy' and 'collidable' attributes of the playfield mesh primitive.

This PR may introduce regression if a table creator has put a playfield_mesh, disabled its physics, relying on VPX to not take in account what is in the UI. This is not likely but with the amount of tables out there, it may happen. A table like this will need the physics of the primitive to be reenabled.

Below is a test table that shows it running, with reflection working, on a very highpoly playfield.
[Collidable PF.zip](https://github.com/vpinball/vpinball/files/8490843/Collidable.PF.zip)

### DIFF
--- a/pintable.cpp
+++ b/pintable.cpp
@@ -3641,6 +3641,14 @@ HRESULT PinTable::LoadGameFromStorage(IStorage *pstgRoot)
             for (size_t i = 0; i < m_materials.size(); ++i)
                m_materials[i]->m_fThickness = 0.05f;
 
+         if (loadfileversion < 1072) // playfield_mesh were always collidable until 10.72
+            for (size_t i = 0; i < m_vedit.size(); ++i)
+               if (m_vedit[i]->GetItemType() == ItemTypeEnum::eItemPrimitive && strcmp(m_vedit[i]->GetName(), "playfield_mesh") == 0)
+               {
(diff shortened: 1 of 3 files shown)


toxieainc:
I could add a fix for the latter issue, by increasing the table version number and then patching up all older tables if these feature a playfield mesh.

vbousquet:
In the latest commit, I added the check when loading with regards to version 10.72 (next one). I did put it in the pintable loading with the other backward compatibility checks to keep them all in one single place since it seems easier for readability/maintainability (vs dispatching it in each class with backward compatibility).

toxieainc:
Thanks!!

toxieainc:
I now bumped the file version, plus (in theory) allow to use 2 playfield mesh-es (so one for visuals, one for physics).

PR #56, opened 2022-04-17
HDR primitive: avoid saturating HDR textures

vbousquet:
Previous PR allowed for HDR texture for primitive, if lighting was disabled. Though, the diffuse color is saturated in the lighting shader, braking the initial intent. This PR disables the saturation if lighting is disabled.

Since stauration happens in the lighting, I think that the saturation in the base technique is not needed and could be removed (I did not removed it). https://github.com/vpinball/vpinball/blob/109634cafb16a98625da4384e7cddc5f5b602193/shader/BasicShader.hlsl#L239

I also included a little fix for primitive that were rendered even when set to additive with a 0 color or opacity (no-op rendering).

### DIFF
--- a/primitive.cpp
+++ b/primitive.cpp
@@ -1335,6 +1335,14 @@ void Primitive::RenderDynamic()
            return;
        if (m_ptable->m_reflectionEnabled && !m_d.m_reflectionEnabled)
            return;
+       if (m_d.m_addBlend)
+       {
+          const vec4 color = convertColor(m_d.m_color, m_d.m_alpha * (float)(1.0 / 100.0));
+          if (color.w == 0.f)
+             return;
+          if (color.x == 0.f && color.y == 0.f && color.z == 0.f)
+             return;
+       }
    }
 
    RenderObject();
--- a/shader/Material.fxh
+++ b/shader/Material.fxh
@@ -142,7 +142,7 @@ float3 lightLoop(const float3 pos, float3 N, const float3 V, float3 diffuse, flo
    const float glossyMax = max(glossy.x,max(glossy.y,glossy.z));
    const float specularMax = max(specular.x,max(specular.y,specular.z)); //!! not needed as 2nd layer only so far
    const float sum = diffuseMax + glossyMax /*+ specularMax*/;
-   if (sum > 1.0)
+   if (sum > 1.0 && fDisableLighting_top_below.x < 1.0)
    {
       const float invsum = 1.0/sum;
       diffuse  *= invsum;


toxieainc:
Thanks.. I want to revisit/review the saturation fix in more detail though.. So tomorrow i hope to wave it through..

vbousquet:
Thanks a lot for your time and patience!

I have tried to integrate in the existing code base, keeping my changes as minimal as possible for easier maintainability. If opened to wider changes, here is my feeling at the saturation and 'unshaded' path:
- regarding saturation of diffuse/glossy/specular, I do not understand at all why this is performed and especially in the pixel shader. Beside the forumla is somewhat cumbersome. I don't see how it can improve the rendering since it will depend on the actual pixel. I would be in favor of no saturation at all, making the shading more understandable by the table author (but I avoid touching things I do not fully understand).

- regarding the unshaded path, I use the default 'disable lighting' path. But in fact, it does not 'disable lighting': environment lighting is still applied, color is multiplied by 2 (since there are 2 lights), and geometric opacity is applied. The table author can have workaround for these but it makes the shading difficult to understand. A nicer way implementation would be an 'unshaded' check box, like the 'metal' check box of the material settings (graying out all the other material options, except active/opacity/base color), and then having a really unshaded rendering. That was the way I started at first for unshaded and additive primitive but it has some large impact on the codebase since it needs to change the way material are saved (biggest problem in my opinion with the backward compatibility) and to be taken in account for all shaders (easier one, since the unshaded technique is straightforward, but large change anyway). The nice point with this approach is that it is more readable and leads to better performances. I can propose a PR if you want to go down this path.

Anyway, these are just my 2 cents. The actual implementation with limited changes is ok as well!

vbousquet:
Since the unshaded rendering is directly linked to the additive (flasher) use case, another possibility would be to rename the primitive's 'additive' checkbox to 'unshaded additive' and adapt the rendering accordingly. If you prefer something more simple like this, I can prepare the PR as well. Just tell me. Regards

toxieainc:
Just a quick answer on 'regarding saturation of diffuse/glossy/specular, I do not understand at all why this is performed and especially in the pixel shader. Beside the formula is somewhat cumbersome. I don't see how it can improve the rendering since it will depend on the actual pixel. I would be in favor of no saturation at all, making the shading more understandable by the table author (but I avoid touching things I do not fully understand).'

This is needed for energy conservation, otherwise the material can generate energy. Its done per pixel, as the texture inputs are influencing this. Its pretty common to do this for physically based materials.

toxieainc:
I guess for now the PR as-is is fine. I tried to look up exactly why the disable-lighting works as it is, but failed. What i think is that some people wanted to avoid the specular highlights of the simple point lights, and thats why i added it like this (only disable the 2 point lights then), but i'm not 100% sure anymore.  :/

PR #63, opened 2022-05-29
Use RGB16F instead of RGB32F

vbousquet:
This PR proposes to use half float instead of full float for HDR textures, in order to improve performance and allows large use of HDR textures for flashers/primitives. The PR was built and then tested by a few members of the VPW group on different hardwares. No regression was identified while performance benefit was obtained.

Beside the overall performance gain, this PR allows older GPU to be able to play tables with lots of HDR flasher/primitives (GPU llike GTX 650,...) while for the time being they would either overflow due to the memory consumption or slow down due to memory bandwidth.

This is proposed as a direct change (not as an option) since the tests did not show any regression and always benefits. Though, this could be made an option in the video preferences (but it will make them a bit more complex and need a bit more support).

The test were made on the DX9 backend, not the SDL one (I'm not sure of its state at the moment), but I can perform further test if you want.

### DIFF
--- a/RenderDevice.cpp
+++ b/RenderDevice.cpp
@@ -1334,7 +1334,7 @@ void RenderDevice::CopyDepth(RenderTarget* dest, RenderTarget* src) {
 
 D3DTexture* RenderDevice::UploadTexture(BaseTexture* surf, int *pTexWidth, int *pTexHeight, const bool linearRGB, const bool clamptoedge)
 {
-   D3DTexture *tex = CreateTexture(surf->width(), surf->height(), 0, STATIC, surf->m_format == BaseTexture::RGB_FP ? RGB32F : RGBA, surf->m_data.data(), 0, clamptoedge);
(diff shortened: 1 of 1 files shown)


toxieainc:
Good idea. I will most likely merge it as-is later-on, and then still make it an option though (and wire it to the respective presets in the video preferences). The reason i picked fp32 is that 'real' HDRs easily feature way too large values for fp16 (=limited by 65504), but i guess that a) everybody just uses what is available on the net for free (i.e. HDRs that are mostly LDRs++) or b) these high peaks are averaged out anyway during mipmap computations (as the highest levels are not picked that often).

toxieainc:
..oh, and having the same for VPVR would be nice, too.

vbousquet:
Ok, nice. I will prepare the PR for VPVR.

toxieainc:
I just commited a variant of your changes that handles larger values automatically. Thanks!

toxieainc:
..oh, and please verify/test my variant again, thanks!

PR #65, opened 2022-06-08
Texture cleanup (port from VPVR)

vbousquet:
This PR ports the changes made in VPVR. This is an early PR to share the progress and get feedback since this cleanup has a fairly large impact on the codebase.

The first commit ports the changes from VPVR but the texture filtering, clamping and sRGB/RGB mode is still implemented in the hlsl file + setSamplerxx functions (like for clamp or linearRGB in the current code base which are functions argument to LoadTexture/SetTexture but these arguments are not used).

This first commit marks a test point to check for further modifications (or deciding not to go further) or be able to perform output comparison tests more easily.

To be precise, the changes included are:
- texture cleanup similar to VPVR:
-- include color space, alpha channel, and float depth in texture, 
-- make channel ordering to RGBA for all textures, 
-- use memory block corresponding to actual data format)
- improve multithreaded texture loading (again) by not resizing in low memory situations during multithreaded loading (do it on second, single threaded, try)
- improve the image dialog's format column by displaying when the texture has been downsized due to low memory


vbousquet:
No problem, no hurry. Thanks for taking the time to review and merge these changes!

I will port back your follow up modification to VPVR. I'm in the process of taking VPVR to the level of VPX. So I have a big PR on this side coming up. I have opened an "earlier PR" to share the progress (not to be merged yet). The idea is to try to have the exact same result on both side, and eventually to be able to build VPX as a build target of VPVR. Things are coming along nicely but I stumbled on lots of little bugs I had to solve, so this will take time as well. It will also allow to test an OpenGL ES on Angle to more easily run on other platforms (like jsm174 works on Macos), or test DX11/Vulkan on Window.

The only change I foresee for the VPX part of the code is regarding shader sampler state. For the time being this is somewhat complex since:
- they are referred by texture unit and not by shader's sampler name (I mean TEXTURE0 instead of 'texDepth' or 'texBloom') making the code less readable/portable,
- the filtering is defined either by the texture/sampler state, or in the effect file,
- the colorspace is defined in the effect file and in the texture object.
I will try to draft and propose something as simple as possible to make it simpler/cleaner.

toxieainc:
Sounds awesome!
Any help is appreciated on that front.. I already tried to make the code bases as similar as feasible, but of course had to spare the most complicated parts so far (what you mentioned above), in order to make both converge at some point again (and then later-on jsm's awesome porting work, too).
It would especially be awesome to have the 'desktop'/DT rendering of VPVR kinda compatible with VPX, especially as it doesn't seem to be much slower as the oldskool DX9/mixed-with-static-buffers rendering code. But so far there are still quite some differences (but then again, its also not THAT bad).

So just go ahead with the changes as you said, and good luck..  ;)
I will try to review as much as i can to catch potential regressions, too.

toxieainc:
Oh, and one more thing: As you mentioned that it would be nice to have VPX as build target of VPVR: Actually, in the end it would be the best to make everything converge within the 'main' VPX/vpinball, as it has the majority of the commit history (as Caligula unfortunately didn't 'properly' fork VPX back then). Thanks!!

Issue #66, opened 2022-06-18
Bitmap corruption by saving in 197


Wylted1:
**Describe the bug**
Corruption of Bitmap (.bmp) image files in saved table by doing nothing but saving.

**To Reproduce**
1. Open table in 10.7.1 197
2. Save file
3. Close VPX
4. Open table
5. Images corrupted

**Expected behavior**
No changes to included files

**Screenshots**
![10 7 1_Fresh](https://user-images.githubusercontent.com/25475953/174455757-053c78cc-386e-4383-a7b5-331939b841a9.png)
![10 7 1_Saved](https://user-images.githubusercontent.com/25475953/174455754-534d4df8-ef5c-4b14-af4f-dad964595909.png)
![10 6](https://user-images.githubusercontent.com/25475953/174456123-360818ad-f0e2-4193-8ea8-8db245e8fa9e.png)
![Saved in 10 7 1_197](https://user-images.githubusercontent.com/25475953/174456126-165282d2-f5e0-4071-8afe-058e78754533.png)

**Versions**
 - VPX 10.7.1 197
 - Windows 10 Pro 21H2

**Additional context**
Only tested on table "Cuphead Pro (D. Goblett & Co 2020)" https://vpuniverse.com/forums/files/file/6145-cuphead-pro-perdition-edition-dgoblett-co-2020/
Ten .bmp images, all affected

vbousquet:
This is a stupid bug I left in my previous commit. I will propose a PR to fix this

vbousquet:
The PR referenced above should solve this. I tested it with the table you linked and it works fine with this fix.

PR #68, opened 2022-06-19
Fix bmp saving bug

vbousquet:
Fix https://github.com/vpinball/vpinball/issues/66

To avoid duplicating code, this uses the data format conversion already used for the editor display (which is also BGRA).


toxieainc:
Great, thanks!

PR #69, opened 2022-06-19
Add Sampler & RenderTarget classes, use them instead of direct low level API

vbousquet:
This PR is a consequence of the work made on VPX/VPVR to have the 2 codebases converge. While working on it, it appeared that all the sampler and render target management code was spread along the codebases, with mixed in between these 2 different concepts. Therefore, I performed a full review and reimplemented this:
- by creating a Sampler class that hosts all the sampler methods (texture creation, update,...)
- by creating a Render Target class that hosts all the render target methods, especially display framebuffer, color render target management, color/depth buffer management and if needed copies to get readback access to depth buffer.
- adding clean access to render target buffers as samplers (they are exposed as a color / depth sampler)
- updating the code that depends on it.

This leads up to a quite extensive cleanup. Especially, since I did not kept the multiple unused methods for mixed in copies between DX9 surface/texture (they are fairly basic, and it seemed cleaner to keep them in the git history instead of the code as commented out/deprecated unused code).

Beside, I also made a full review of the depth buffer copies since there was a useless depth buffer copy (source and target being the same buffer) and checked it using the normal or alternative (NVidia API) depth buffer.

I did make quite a lot of tests on this (static AO, dynamic AO, different antialiasing, normal/alternative depth buffer,...) and did not find any issue, but, as always, since the change is quite large, I may have missed something.

This PR does not fix anything (beside the useless copy) or add any feature, but I think it is worth merging since it results in a cleaner/easier to maintain codebase, and helps to move VPVR and VPX toward a common ground.

### DIFF
--- a/RenderDevice.cpp
+++ b/RenderDevice.cpp
@@ -37,10 +37,6 @@
 #endif
 #pragma comment(lib, "dxerr.lib")       // TODO: put into build system
 
-static RenderTarget *srcr_cache = nullptr; //!! meh, for nvidia depth read only
-static D3DTexture *srct_cache = nullptr;
-static D3DTexture* dest_cache = nullptr;
-
 static bool IsWindowsVistaOr7()
 {
    OSVERSIONINFOEXW osvi = { sizeof(osvi), 0, 0, 0, 0,{ 0 }, 0, 0 };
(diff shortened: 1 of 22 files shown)


vbousquet:
I did update the initial commit by amending the needed changes to all the project files in order to pass the appveyor build test (I had forgot to update the one I do not use). That's the reason there are a few force-push reported after the initial commit.

toxieainc:
Sweet! Will try to review it later-on..

Issue #73, opened 2022-06-23
Refcount/Freeing of Rendertargets off by 1


toxieainc:
**Describe the bug**
@vbousquet Enabling DEBUG_REFCOUNT_TRIGGER now triggers some warnings for Rendertargets (and more)

toxieainc:
Unfortunately there is still some missed resources now, as
SAFE_RELEASE(m_pD3DDevice);
and
SAFE_RELEASE(m_pD3D);
both show a high refcount (23 & 2) for example for the CTRL+N new table for me.

toxieainc:
Seems to be resolved now, just verified, thanks!

PR #74, opened 2022-06-23
Fix back buffer ref leak, and wrong ref count alerts

vbousquet:
Fix https://github.com/vpinball/vpinball/issues/73

After checking DX9 docs, there were 2 problems here:
- when render target is back buffer, it was not cleanly released
- there was wrong alerts since the refcount is shared between a texture and its associated surface (they get down to zero ref after releasing the both, not only the first, needing to use the _NO_RCC variant as it was the case in the previous implementation)


toxieainc:
Great, will verify, too..

Issue #78, opened 2022-06-29
Broken pup apron and back wall screens


shaggy2dabs:
Last working rev is 187 anything after breaks pinupplayer apron and backwall screens The colors are wrong. Im am on rev 215. Red colors show up as Blue
!
![289811460_562932238704050_5444259171581298632_n](https://user-images.githubusercontent.com/29291907/176557929-09a4414c-885b-4bcf-aed8-270e9490cca4.jpg)

![289196836_1322989504895265_3207373582253402530_n](https://user-images.githubusercontent.com/29291907/176557942-c5f8fb7a-ee02-45d8-becd-c0cb66cdc362.jpg)

PR #79, opened 2022-07-03
Fix BGR/RGB in video capture

vbousquet:
Fix https://github.com/vpinball/vpinball/issues/78

This PR fixes inverted R and B channels in video capture flashers. This is a simple straight forward fix that should not impact performance that much.

Beside the fix, we could think of improving BGR management in VPX, either:
- by adding a fast SSE2 conversion code (I did put a reference to the Angle project implementation which could be just copied in and used anywhere we need it),
- or by adding sBGR and sABGR texture format to the list of formats supported by the BaseTexture class.
What do you think ?

### DIFF
--- a/flasher.cpp
+++ b/flasher.cpp
@@ -1067,9 +1067,13 @@ STDMETHODIMP Flasher::put_VideoCapUpdate(BSTR cWinTitle)
         const DWORD* const __restrict pCurrPixel = (DWORD*)lpbitmap;
         DWORD* const __restrict data = (DWORD*)m_videoCapTex->data();
 
-        //copy bitmap pixels to texture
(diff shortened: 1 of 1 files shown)


toxieainc:
I guess the SSE2 code should do the trick, why not.
But it also should not make a great perf difference, i guess. So whatever you prefer.
Thanks!

vbousquet:
Honestly, I would put this at the end of the TODO list since the impatc is expected to be really minimal. It is just that I usually try to avoid proposing commits which downgrade performance. So, I propose to keep it in mind (it is in the source tree comment) and review it later when the other changes are finished (who knows we may end up using bgfx like Mame :)).

PR #103, opened 2022-09-01
Fix Z artefacts in camera mode

vbousquet:
In camera mode, if you go too near the objects, the rendering will turn wrong. This is because enar plane is allowed to drop below 1 leading to z buffer imprecision problems.

Beside this fix, 2 small cleanups commits are included:
- in camera mode, the camera matrix update was done in the debug code, I propose to move it in the render method (where VR matrix update or BAM update is called in VPVR)
- a very minimal commit to limit the code diff between the codebase without any impact

This PR was tested (desktop mode, with a few tables) and should be safe to merge

### DIFF
--- a/CrashHandler.cpp
+++ b/CrashHandler.cpp
@@ -121,7 +121,11 @@ namespace
 
    void WriteHeader(FILE* f)
    {
+#ifdef ENABLE_SDL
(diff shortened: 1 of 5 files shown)


toxieainc:
I hope the near plane fix doesn't mess with stereo 3D (cause i remember something weird with that), but our testers will find out. ;)

PR #128, opened 2022-10-09
Refactor of playfield rendering & reflections

vbousquet:
This large PR contains a complete refactor of playfield rendering. I have tested it quite a lot and it worked without flaws on all my tests but it introduces changes that could have side effect, for example on tables that would rely on things under the playfield to be clipped (why would someone do that ? but who knows ?).

Since this PR is made against the current VPX version which includes changes from VPVR which are not fully ok, I did include it in the changelog for 10.7.3 (not the current 10.7.2) which is the version where the glitches imported from VPVR should be fixed. I did not state it in the changelog but there are also minor UI cleanups to ease user experience (disabling some UI fields for playfield mesh since they are not taken in account and overwritten by table settings at runtime).

Regarding performance, I did not make extensive tests but in the end (and to my surprise) on most of my tests, the performance would be the same or even better than existing rendering path. I also included a reflection mode without depth synchronisation that has little to no artefacts and very good performance for lower configuration users.

Regarding the changes:
- playfield is now processed like all the other parts instead of being rendered before them with under playfield clipping and blending limited to alpha clipping:
  - if the 'playfield_mesh' primitive has the 'Static rendering' disabled, it will be rendered with transparent parts (this results in 'playfield_mesh' to be blended like other primitives, no more need for alpha test or insert overlays)
  - parts under the playfield (like lower playfield or 3D inserts) can have 'Static rendering' checked and can have materials without the 'Active' checkbox (under playfield parts without these were clipped out before),
  - playfield texture can now be changed at runtime throught the script (see CommandReference.txt)

- playfield reflections are now rendered using a 'reflection probe' instead of a 'fullscreen mirror overlay':
  - performance may change a bit (better or worse) depending on the table and display configuration
  - reflection settings were unified between balls/playfield and for camera mode, and extended with new rendering modes:
    - Disabled: *NEW* remove all reflections
    - Balls: dynamic balls only
    - Static: reflection of static parts only (no change)
    - Static & Balls: reflection of static parts and dynamic balls (no change)
    - Unsynced dynamic: *NEW* reflection of static parts and dynamic parts without depth buffer synchronisation. Overlapping of dynamic over static and of static over dynamic reflections is incorrect. It has a limited performance cost. Added for lower end systems.
    - Synced dynamic: reflection of static parts and dynamic parts with depth buffer synchronisation. This corresponds to the way reflections were processed before. Overlapping of dynamic over static reflections is incorrect. It has a small performance cost.
    - Full dynamic: *NEW* dynamic reflection of all parts. It gives the best results without any overlapping artefact at the cost of more performances.

- reflections now work in Camera mode as well

Regarding alpha transparency, I chose to mask reflections depending on alpha. This leads to:
- fix situations were reflections would happen above holes (with current VPX, you can see ball reflections above scoop holes in many existing tables which do not have a playfield_mesh and use transparency for holes),
- but remove reflections above transparent parts likes VPW 3D inserts or when ball roll above a lower playfield glass.

One solution I did consider but did not implement was to add a 'playfield reflection' toggle to the primitive visual properties.  This would allow to have any primitive rendered with the reflection probe applied. In this situation, special uses of the transparency (lower playfield glass / VPW inserts) could have an additive primitive with playfield reflections on the offending parts.
The other solution I considered, would be to implement transparency as just transparency and not a mesh mask (so apply reflections over transparent playfield parts).
Since there are no perfect solution and the resulting artefacts seems bearable (at least to me). I just postponed the solution to this problem.

And here are screenshots to illustrate the rendering modes (default table, table reflection pushed to 100, moving the ball near the upper guide):
- Synced dynamic: dynamic reflection not occluding static one:
image
- Full dynamic: correctly occluding reflections between static & dynamic:
image
- For reference, VPX 10.6 up to 10.7.2RC would render like below. First image is without dynamic reflection, second is with dynamic reflection enabled (showing 2 artefacts of the current dynamic rendering: static reflection is doubled, dynamic reflection does not occlude static one):
image
image


toxieainc:
Hi! Sorry for breaking this PR, but i still had some unfinished changes/tweaks to your previous commits that i first wanted to get in. Please also review these, i hope i didn't screw something up.  :)

toxieainc:
Will hopefully have a closer look at this PR later today.

vbousquet:
No problem (in fact, thanks for tidying my mess :)). I will rebase and resolve the conflicting commits. 

Also, I have worked on the remaining open point in this PR (how to handle alpha for all situations) and came up with what I think looks like a good solution: take alpha in account when there is no playfield mesh (tables that use alpha for scoop and cutouts, so should not have reflections in it) and not for table that do have a playfield mesh (tables that use alpha for glass transparency, so should have reflections). I will work on it likely next week-end and come back with an updated PR.

vbousquet:
Regarding filtering, the way it is planned to be implemented is the following:
- sampler uniforms have a default filtering/clamping
- when binding a sampler, you may override it with an explicit filtering/clamping or bind it with UNDEFINED which then use the default for the sampler uniform.

This is done this way since most of the time, the sampler holds the meaning and the binding is just giving the image source. I think that's the reason why samplers are defined in the effect file in DX9. I think that setting the filtering/clamping while binding a sampler should only be performed for an explicit override (so fairly seldom).

For the time being, this is fully implemented for OpenGL and partly for DX9 (I did slow down on the DX9/GL merge since issues had appeared and I wanted to fix them first). So for the time being for VPX, the filtering/clamping is the one from the effect file + SetTextureFilter (like in DX7).

toxieainc:
Sounds great, thanks!
I will wait then with the review until your next changes are in.

vbousquet:
It is fairly good as it is right now (you can see it in action in latest VPVR) so it can be pushed. I was postponing since the more I dive into probes, the more I find it powerfrul and to be pushed further (at least to add some blur to reflections, support multiple probe for upper/lower pf or other mirrors, better manage playfield transparency, and also experiment refraction probe)

Anyway, just give me 5 minutes. I will clean up and push what I have pending on my side to have a first milestone.

toxieainc:
Yes, probes could be a good option for pinball as everything is happening just in this tiny controlled box. So one could think about just distributing a couple dozens/hundreds of probes (maybe not on a grid, but inbetween objects, etc, using the collision engine to properly place them??).
I was always scared of adding them as it didn't sound like fun with the DX9 code like it was, but now you modernized/cleaned up quite some, so maybe??

vbousquet:
Ok, this can be rebased/merged. Do you want to go for a review or can I go for the merge ?

Regarding probes, what I had in mind is to allow:
- blurry reflections (I think they often too much look like a perfect mirror)
- reflections for lower/upper playfields (at least static if it's causing performance issues)
- blurry, offseted (deformed) refractions
- tinted plastics (not possible at the moment since VPX use alpha blending, not tinting the receive light but replacing it)

The way I was thinking of was a basic implementation built up on what we have (to be added step by step):
- have a 'Plane Reflection Probe' concept:
  - defined by a plane (position/normal), blur level (downsampling amount), max static/dynamic render level (to be combined with user settings to have the actual level)
  - prerender each probe like what is done in this PR for the playfield

- have a 'Screen Space Probe' concept:
  - defined by a blur level (downsampling amount), max static/dynamic render level (to be combined with user settings to have the actual level)
  - during render, the first time a user of this probe is encountered, a copy of the render buffer is made taking in acocunt downsampling

- allow primitive to select an optional reflection prove for reflection with reflection amount, and an optional screen space probe for refraction
  - for reflection: if the primitive has an alpha channel, then render the reflection as an additive pass, if not just render applying it directly
  - for the refraction: use the primitive image has the tint/transmit amount, sample the screen space probe with a fake offset computed from the normal

- add some UI to setup probes (like image manager,...)
- all tables would have at least the default, undeletable playfield probe for reflections
- playfield_mesh would have a default setup using the default playfield reflection probe

toxieainc:
Just go for the merge then, you're the expert for this now anyway.  :)

As for the UI to setup probes: What do you have in mind there?

vbousquet:
For the UI, nothing fully clear for the time. The idea so far would be to have a dialog like for material/collection/image/... managers that would allow to set up the probes. Then have the possibility ot select them in the primitive (maybe wall as well) properties.

But before playing more with this, I think the cleanup should be finished first. There are still bugs to fix (JP's reported ramp bug and offset of the rendering/postprocess), and the merge to finish (the biggest part should be the shader state management, moving out of the state machine like it is right now to the way it is managed for VPVR where state is applied on draw call. It makes things easier to manage, allow render queue instead of static sorting, correct draw order for dynamic camera, easy frame debugging,...)

toxieainc:
Great job! I agree that finishing the cleanup/merge first is the best idea.

One nitpicking with the new reflection/mirror code: In the 'old' code i did some implicit blurring (2x2) via the texel lookup to make the reflections a bit less sharp, maybe you could do it the same way (unless its too complicated now)?

vbousquet:
This type of bluring can be done easily (just offset the sampling coordinates). I will have a look.
If a bit of bluring was a aim, why didn't you went for simple (and faster) downsampling (rendering at 0.75x or 0.5x, then using bilinear filtering, eventually add an intermediate blur pass) ? Did you tried ? Would it give good result ?

toxieainc:
As it was just a preprocessing pass i didn't care much for perf, but rather plain quality, plus it was easiest to just render at the original res.
So the implicit blur (2x2) via the bilerp lookup seemed like the easiest and highest quality tradeoff.

PR #134, opened 2022-10-19
Vpvr sync

vbousquet:


passionforpins:
Vincent I don't use VR and only run in desktop mode but rev 421 no longer accepts backdrop art. 410 does but also seems to pull in other jpgs in the table folder even without importing. I created a bug report on 410

PR #265, opened 2023-01-25
Fix analog nudge for non-DirectInput controllers (fixes #106)

### LINKED ISSUE #106: Analog nudging still broken, from mid 1.7.0 beta to present
All roads lead to UP

tested with Xinput device (dof nudge preset set to xinput API)
https://i.imgur.com/68DsKL5.png

Apron indicators as well as numeric display show correct axis inputs
but balls only react by going UP
and it does not matter how strong the nudge is either, the balls go up at the same force regardless
in 10.6.X if you set the nudge hard enough you can nearly drive the balls around the table
in 10.7.X all you can do by turning nudge up is tilt yourself faster, the balls still just blip UP a little

Video below will explain the issue fully

www.youtube.com/watch?v=8aS8zWe6jM0

if switched to dinput and switched to a dinput device, the nudging will work, but has to be totally recalibrated
but the feedback breaks
giving vpx a set of dinput dlls seems to do something to feedback (stick vibrates when table loads)
but then the nudging ceases working right

this was reported during 10.7 beta but at the time we didnt have github to log bugs

Chickenzilla:
It seems the standalone branch is tracking to replace a lot of this potentially but for now I figured I would offer this PR to fix #106 anyhow.  Feel free to ask questions, accept, or close at your leisure of course.  I've play-tested this under Wine on linux for several days now in both DirectInput and XInput modes and it seems to work well enough for me.  Also confirmed the standalone branch currently has the same bug.  A more detailed explanation of how this fixes the issue is in the linked issue as well.  Thanks for all your great work on this project!


toxieainc:
Thanks looking good!

PR #301, opened 2023-02-18
Misc fix

vbousquet:


vbousquet (commented):
@toxieainc I have just realized that this commit has some BGFX part in it. This was not intentional (I cherry-picked this commit for the rubber fix, without realizing that it also contained BGFX shaders) I will revert the BGFX part which is nothing but clean enough to be proposed. Sorry about that

PR #303, opened 2023-02-19
Add BAM menu

ravarcade:
Add BAM to VS 2022, 2019 project
Add tinyxml2 to project
Add BAM menu as last position under F11 menus


toxieainc:
Looking good, thanks!!

PR #315, opened 2023-03-05
Reworked IMGUI user interface & fix camera mode interactive light settings

vbousquet:


toxieainc:
@vbousquet Nice! Should we just delete the old non-ImgUI path then?

btw: Somehow the FPS stats graph scaling was broken in the past, did you also fix that one?

vbousquet:
Yes, I don't see the point of maintaining both UI path. I will clean up a bit the ImGui one (for the time being, it still lacks DPI awareness, a basic theme to be clean and the table info display) and remove the old info one.

I will also have a look at the keyboard shortcut since I don't think we still need 2 shortcuts (ESC & F11), ESC should be enough to open the ingame UI and eventually access the profiling section.

Regarding cleaning up, I think this could be pushed somewhat further. There are (at least) 2 other parts which are kept behind #ifdef for which we could take a decision and have a single codebase to maintain. I did not allow myself to touch these but they seem somewhat 'legacy' to me. Namely:
- SEPARATE_CLASSICLIGHTSHADER => just drop one of the options (separate or not). I would keep the current default and remove the separate one,
- ENABLE_BAM => just enable it for all builds (it does not depend on OpenGL) and show the UI control only if head tracking is enabled in the settings

And finally, we have 2 xml library: rapid_xml and tinyxml2. Perhaps we could decide for one of the boths (rapidxml seems somewhat declining while tinyxml2 is quite active on github, but I have no experience with those 2).

I did not look at the FPS scaling. It is just the code as is. I may have a look while I'm at it.

vbousquet:
I pushed a (large) update here https://github.com/vpinball/vpinball/pull/316

So now we have:
- DPI awareness (was really needed for HiDpi/4k screen)
- clean, splitted code between player and UI
- support for rotated display (limited, without mouse interaction, for ingame overlays and camera mode feedback)
- cleaner ImGui theming
- port of all existing information displayed
- single key shortcut to enter UI mode
- removed legacy UI

The only missing part is better keyboard shortcut handling.

I did not look to the FPS plots and did not go further with portable file dialog, nor did the other proposed cleanups.

vbousquet:
Large New batch of improvments merged in: https://github.com/vpinball/vpinball/pull/318

It happens that coding UI with ImGui is really really simple and fast, so I pushed it (largely) further than expected with full scene inspection and editing (but not persisted), keyboard shortcuts, material/light/camera/table properties editing (like in existing debugger dialog).

It happens to be quite fun, so i think I will continue to push it somewhat further (I would really love features like a free camera in 3D viewport, some more live editing,...). Let's see where this path leads...

toxieainc:
@jsm174 I guess its up to you to pick one..  :)  I have no clue which ones are considered sane and stable.
@vbousquet Insane.. Always wanted to go into that direction but never had enough time, great job!

toxieainc:
>     * SEPARATE_CLASSICLIGHTSHADER => just drop one of the options (separate or not). I would keep the current default and remove the separate one,

Yes, please. Nowadays the separate one does not make too much sense annymore.

>     * ENABLE_BAM => just enable it for all builds (it does not depend on OpenGL) and show the UI control only if head tracking is enabled in the settings

Yup, makes sense.

> And finally, we have 2 xml library: rapid_xml and tinyxml2. Perhaps we could decide for one of the boths (rapidxml seems somewhat declining while tinyxml2 is quite active on github, but I have no experience with those 2).

tinyxml2 seems like the better option going forward. Its also the one we use in RL projects.

fuzzelhjb:
Really cool. Nice job.
Does the UI also rotates in portrait view or just in landscape?
When I added it I thought about to use it in camera mode for cab users too.
But at that time ImGUI wasn't able to support this special portrait mode. I
had issues with the mouse events.

toxie ***@***.***> schrieb am Mo., 13. März 2023, 08:04:

vbousquet:
@fuzzelhjb The rendering can be rotated but mouse handling is a pain, so I limited the rotated use to overlays (camera mode, FPS,...). It's more or less a little hack, so i would like to limit its use to avoid problem and better maintainability.

@toxieainc Fully agree for these 3 and will provide the corresponding cleanups in the upcoming commits.

If you could have a look, I also have some more decision proposals in the last follow up PR here: https://github.com/vpinball/vpinball/pull/321
(this last PR let a few open bugs that I would like to close before too many early adopters stumble upon them)

PR #334, opened 2023-03-19
Fix for edge-case DSound enumeration crash

ScaryG:
If the registry entry for a DSound device is past the end of the enumerated array, the existing code will attempt to dereference anyway and crash. This PR fixes this case.


toxieainc:
Thanks!

Issue #442, opened 2023-05-08
No anaglyph 3D modes working in cabinet mode Fullscreen (FS)


gjupp:
**Describe the bug**
If playing table in cabinet mode, selecting any of the inbuilt anaglyph 3D modes results in no noticeable 3D effect, no visible colour separation. Note: 3D function DOES work in desktop mode.

I also suspect that LG 3D interlaced screen modes, side by side and top bottom may not work in cabinet mode although unable to test this.

Problem has existed since at least build 1002. Pressing F10 to toggle 3D effect off/on shows no noticeable difference.

**To Reproduce**
Switch 3D stereo output in Graphics options to any 3D mode
Play table in cabinet mode fullscreen (FS), (270 rotation tested with windows screen setup set to landscape for playfield, 1080o).
(Use Y axis also has no effect either on or off).

**Expected behavior**
Anaglyph 3D effect should have noticeable red/cyan separation visible even with no glasses (or whatever colour mode is chosen) 

**Versions & Settings**
 - VPinballX-10.8.0-1057-78f984e-Release-win-x86
 - Windows version 10
 - GPU HD4000
 - If this bug report is for VPX 10.8+, please join your setting file (locate 'VPinballX.ini' in '%AppData%\VPinballX', either zip it or change its extension to '.log', then drag'n drop it here)
 
**Additional context**
[VPinballX.log](https://github.com/vpinball/vpinball/files/11417056/VPinballX.log)

vbousquet:
I have just tested with your settings and I do get anaglyph separation both in desktop and cabinet mode. Not sure what the problem is. Could you share screenshots (in camera mode to have show the live settings) ?

gjupp:
Hmmm, that is odd. Here's the screenshots with red/cyan anaglyph mode activated for both cabinet (not working) and desktop (working although default 3D settings seem not great to me).

I should also mention that the inbuilt anaglyph 3D modes were previously working in cabinet mode in 10.7.3 although also not great compared to reshade/superdepth3d.

![3D_cabinet](https://user-images.githubusercontent.com/95056536/236811802-71c747b7-27f9-49eb-9f34-d9f7068bc190.jpg)

![3D_desktop](https://user-images.githubusercontent.com/95056536/236811820-a0f4a1e0-b663-4f2b-9d0e-01a44c25be2e.jpg)

vbousquet:
[...]

Issue #427, opened 2023-05-03
VPX 10.8 with not enough depth using reshade


emanoel299:
**Describe the bug**
VPX 10.8 using reshade 5.8 or any reshade version does not work like VPX 10.74 for the same settings.
There is not enough depth and looks like reshade GL version.
I am using reshade 5.8 with Superdepth 3D blue sky defender
SBS (side by side) stereoscopy, swap eyes and Y axis active.
TV 3D LG passive with passive glasses in cabinet mode.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Crash dump**
If it's about a crash, please attach `crash.dmp` as well.

**Versions**
 - VPX: [e.g. 1.7.0 Final Revision 367 (e242d7c), 32 bit - see *Help - About*]
 - Windows version
 - GPU and video settings if applicable

**Additional context**

gjupp:
Are you using the GL version of VPX or standard directx? Is there any 3D depth at all or is it just reduced? Perhaps post some screenshots of the greyscale depthmap comparison between 10.8 and 10.74 . I can't remember offhand what the setting is to get the greyscale map in superdepth3D shader but I know it is in there somewhere.

Also, maybe also mention this in the thread below to see what other superdepth3d users think https://www.vpforums.org/index.php?showtopic=48019 (I'm Gravy at VPF).

emanoel299:
Hi thank you for your reply

Are you using the GL version of VPX or standard directx?

I am using VPX standard directx version.
Any VPX GL version does not work good with reshade. There is depth but no 3D effect. It shows like a picture in the bottom

Is there any 3D depth at all or is it just reduced?

Just reduced depth
I tried ZPD adjustment but no effect

Perhaps post some screenshots of the greyscale depthmap comparison between 10.8 and 10.74 

Attached there is Greyscale screenshot (Superdepth3D - Depth map view)
Pirates of the Caribbean Siggis Mod 1.0

![VPX 10 74](https://user-images.githubusercontent.com/132472149/236106776-d610fd4a-af57-4dad-8df9-f8abed6eb056.png)
![VPX 10 80](https://user-images.githubusercontent.com/132472149/236106785-68903bd5-b60a-4109-97ec-14b8a191199a.png)

gjupp:
Hmmm, certainly a major difference there, I wonder if a change was made to better suit the inbuilt anaglyph/3D modes.

[...]

PR #451, opened 2023-05-08
Rewrite Camera matrices, add a new "absolute" mode

vbousquet:
This one is a large PR since diving in understanding and trying to fix https://github.com/vpinball/vpinball/issues/442 and https://github.com/vpinball/vpinball/issues/427 made me find a bunch of bugs/improvments.

I think there are a few problems with the way matrices are currently built:
- they are relative to a (partial) table bound, needing to be adjusted for eahc table,
- there is no feedback to something understandable, so it leads to setting the camera in very unrealistic ways (meters away,...),
- it uses a 'layback' transform which is not clean since it breaks the orthonormal property of the view,
- there were a few duplicates and bugs in the math classes, especially between OpenGL & DX versions.
All these make it very difficult to use the camera matrices (for anaglyph stereo, we need to generate the 2 eyes matrices).

So, I fixed what I could and I also create a new camera layout mode which try to improve all these:
- it is an "absolute" (not relative) positionning system
- the player position is relative to the bottom center of the table which should seem natural
- it only uses classic transform, keeping the view matrix orthonormal (no layback)
- it implements the layback feature using offcenter perspective projection which seems to work quite well

This also allowed a few general improvments:
- cleanups a bit everywhere in this part of the code
- in camera mode, add a key to define a default layout which is ok as a starting point for most table (thanks to absolute positionning)


gjupp:
Following this one as will hopefully lead to better 3D support.

vbousquet:
@gjupp  The anaglyph is quite better now (especially OpenGL one which will always be way better than the DX or Reshade option). This needs to setup the camera with the new camera layout mode (simply press credit key when in camera mode). In fact, the bug reports made me dive into it and understand why I would not be able to solve it with the 'legacy' camera (which is not a clean orthonormal view matrix + clean projection), then propose a new one. I still have a few improvments to push on this but it is ready to be tested with version 1068.

Here is an example of Flupper's TOTAN table in cab view with the latest build (to be looked at with anaglyph glass with a side angle), you will see that the depth is more realistic than in previous version, and that it is fully working for transparent parts (ramps for example):
TOTAN_anaglyph

gjupp:
That's pretty exciting news Vincent. A huge thankyou.

Is your image above from the GL or DX version (is it full stereo rendering?)? It's looking very good and your new method of positioning the 3D camera view from player's perspective seems to certainly help maintain correct separation from front to back. Something that was very difficult to achieve if not impossible in Superdepth whilst still retaining good depth.

I haven't been playing in 3D for the last few months as my discrete GPU died in my laptop and I'm having to hobble along with integrated HD4000 which is nowhere near fast enough for reshade/superdepth3D. A PC upgrade is something I will look at later in the year.

I'm about to turn 50 and noticed my vision starting to degrade a bit, this is more apparent for me in 3D with my left eye as the red lens cuts out a lot of light and is causing me some focus issues. I may have to look at trying to find an LG 42 passive 3D screen, a shame they are only 1080p though. Maybe a brighter screen or OLED may be better for 3D users whose eyesight is starting to fail a bit as my current screen is just an old 40 inch non 3D LG 1080p.

I've actually been using Trioviz inficolor 3D anaglyph glasses mostly which have magenta in the left eye and green on the right and I think are generally brighter and clearer for me now with better colour accuracy (especially in Inficolor mode in superdepth3D, something VPX inbuilt doesn't support unfortunately).

I'd love to see a 3D image of TOTAN in full colour mode with left eye magenta / right eye green if you have a spare minute to see how it compares to your red/cyan image. Hopefully I'll actually be able to test the new version in DX soon at least, as unfortunately the GL version wont run on my HD4000 gpu as you may remember.

Lastly, are you running your cab playfield display in windows portrait mode for your monitor setting? I ask this as your example image above is the right way up already, but perhaps you just rotated it after taking the screenshot? I ask this just in case it is related to the reason the 3D has not been working in cab mode for me as my cab playfield is set to landscape orientation in windows.

vbousquet:
The real stereo is only possible with the OpenGL version. DirectX 9 does not support this. DirectX9 and Reshade will always be very limited in terms of stereo rendering as they do not have the 2 eye renders (with different point of view, based on real eye geometry). That means that transparent parts will always be wrong, there will always be artefact near shape borders, render will always depend on depth buffer precision and depth reconstruction...

My cab orientation is portrait but this should not matter as with the new camera positionning, the cab orientation does not impact the result (it is applied like a 2d postprocess, not in the view matrix, but in the projection one).

Regarding the anaglyph colors, I'm thinking of improving that too (by making the color directly user adjustable with color intensity above 1 possible to account for all glasses which have color filter more or less balanced) but likely not in 10.8 since I would like to finish and release it as soon as possible (in fact, without these bugs, I would not have pushed the new camera in 10.8 in the first place, but I don't see a way to get correct results without it).

One of the thing I would like is to have good quality stereo 3D with headtracking. I think they will play really nicely together but that's for 10.9 or even 11, there's so many things that can be done with VPX!

gjupp:
So your image above is from GL version?

The lens colour customisation would be a great feature as there are certainly shade variations in different brands of glasses. Would be nice to tweak to prevent ghosting.

and the headtracking with 3D is the holy grail, were you thinking of a new option for this other than BAM?

I tried portrait orientation for playfield to see if it fixed the lack of 3D in cabinet mode, but still didnt work for me. Hopefully your upcoming changes fix that issue.

Thanks again for your brilliant efforts.

vbousquet:
Yes the screenshot above is from OpenGL.

The latest change will need you to update the camera layout to get correct results (run in F6 camera mode, press credit key, adjust but only inclination/field of view/view offset, not player position nor scale). The result with the previous view/projection matrices depends a huge lot on each table & user setting which happens to be often very unrealistic (things mathematically corresponding to player 3m away from the table playing with magnifying glasses :) ) making it nearly impossible to get right for all the tables, user setup, and both fake parallax and real stereo.

Regarding headtracking, BAM seems nice, Opentrack too. I just hadn't the time to really test this, and I wanted to get the base view/projection stuff clean first

dekay:
Open Track would be wonderful for the cross platform support it would offer.

PR #569, opened 2023-06-13
10.8.fixes

vbousquet:


vbousquet:
There are still 2 issues with the latest changes:
- SMAA is broken for stereo GL (ok in normal GL mode)
- Stereo is broken for DX

PR #622, opened 2023-07-07
Fix exception crash on Windows 7.

CapitaineSheridan:


vbousquet:
Thanks a lot!

(for reference https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-writefile states that lpNumberOfBytesWritten must be non null)

PR #692, opened 2023-08-25
Added build bat file to generate projects from workflow cmake files. …

droscoe:
Similar to the pinmame script I created for generating Windows project files from the workflow cmake scripts, I've ported it here to make it easier to build vpinball for Windows.  I also added a .gitignore to ensure the copied cmake scripts and build artifacts don't accidentally get checked in

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+.project
+build/
+/CMakeLists.txt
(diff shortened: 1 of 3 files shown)


toxieainc:
Nice!

Issue #736, opened 2023-09-03
Since rev1394, the tables do not start, VPX crashes without an error message.


workerb0y:
VPX does not start properly since rev1394. The editor loads the table, but when I press Play and the loading starts, it immediately freezes and exits to Windows without an error message, the complete VPX closes. The same is the case with the updates released since then (I tried with 1403 last time). Both the DX and GL versions produce this error.
I have attached the crash.txt.
[crash.txt](https://github.com/vpinball/vpinball/files/12506316/crash.txt)

My setup:
VPinMame 3.6 beta 488, Freezy 2.2.1 r11, b2s server 2.0.3, all in x64.

vbousquet:
Could you do it with a debug build (the ones with debug in their name) and share the crash.txt generated ?
This should produce a better crash log (sadly, the one you have shared does not have any information helping to find the bug)

vbousquet:
Using the latest would be better (just pushed some crash & bug fixes)

workerb0y:
> Using the latest would be better (just pushed some crash & bug fixes)

I tried, the error also appeared, but no new crash.txt file was created. I still see the same crash.txt file that I uploaded here. Maybe I need to look for it in a specific place?

Jayhay89:
Same, since latest build, tables crash on load
Will try debug and share crash log

Jayhay89:
@vbousquet 
This is all the debug build produces... same as non debug
[crash.txt](https://github.com/vpinball/vpinball/files/12506577/crash.txt)

workerb0y:
I turned on logging in the editor, the following vpinball.log was created after the error. 
Also, I've attached the dmddevice.log and flexdmd.log files, just in case...

[vpinball.log](https://github.com/vpinball/vpinball/files/12506581/vpinball.log)
[DmdDevice.log](https://github.com/vpinball/vpinball/files/12506582/DmdDevice.log)
[FlexDMD.log](https://github.com/vpinball/vpinball/files/12506584/FlexDMD.log)

PR #740, opened 2023-09-03
Fix crash when using anti strech ball

vbousquet:
Fix https://github.com/vpinball/vpinball/issues/736


workerb0y:
Thank you so much, it works properly now! <3

Issue #827, opened 2023-09-26
Window POV cm units only valid for a fixed screen size


Duglim:
**Long explaination:**
While playing with the new window POV settings and doing different tests with perspectivs I think I found, what makes the adjustments a bit odd:

- Let's assume you leave the table scale at 1:1:1 to simplyfy things.
- The "player Z", "Window Top Z Offet" and "Window Bottom Z Offset" are needed to adjust the general view perspective.
- All measurements are given in cm units, e.g. default "player Z" is 67 cm.
- If you have a big screen/monitor (e.g. 42 inch), the playfield comes close to a real pin playfield. If you look at an object on the back of the table, it would be about 120 cm in front of you and would look perspectivly correct at player height 67 cm.

BUT:
- If you have a smaller screen (e.g. 32 inch or like mine 28 inch) the same object is only 80 cm in front of you. Imagine a smaller real cab (hyopthetically), there you would have a view generelly more from the top, as you could never gain as much perspective depth. The playfield is simply not long enough. (Example: Medieval Madness: You would see the castle more from the top, if the cab would be just half the size).

I wonder, what would/should be my goal in window POV? 
Create a view like on a real fullzise cab, even if it is perspectivly wrong?
OR
Create a most realistic perspetive for your individual setup?
I think everyone need to find their own answer...

Calculation:
Knowing about this helped me, to calculate a rough factor multiplicator to find better perspective correct values for "Player Z" and both "Window Z". I need to multiply "Player Z by 1.5 to have a more correct top view (resulting in 100cm). "Window Z" needs to be multiplied by 0.66 to fit my 28'' screen cab.

**Short story:**
- "Player Z" and all other window POV settings with cm units only apply to a big screen (maybe 42 inch?). 
- If you have different screen sizes, you need to add a factor to the numbers to get perspectivly correct views.

**Ideas for improvement:**
- Maybe add some information to the POV desctiption here on Github.
- Maybe add a window POV setting for the screen size and calculate/apply the factor internally.
- Maybe get rid of the cm units and use values without units.

Or maybe just close this ticket, if I'm too stupid to see the whole window POV thing right. :) 
Nevertheless, thanks for all the fantastic work, all new options and implemented features are simple great. I would never citizes anyone here. :)

Thanks, Duglim

vbousquet:
[...]

Issue #817, opened 2023-09-25
View Layout Mode: Window  <--drastically reduces playfield quality (sharpness) compared to Legacy


redbone75:
Description:  Playfield quality (sharpness) is reduced when Playing a table with F5 vs viewing with F6

To Reproduce:  To easily see the quality degradation, press F6 and Zoom into table with both Legacy and Window Layout Modes. Save a screenshot in each of these modes.  Play the table with F5 in each of these modes. Save a screenshot while playing each mode.  Compare the screenshots.  It will be noticed that Legacy mode doesn't exhibit any degradation in quality of the playfield.

Expected behavior:  I would expect the quality to be the same in all Layout Modes.  I would expect the quality not to change in the same layout mode while playing F5 or adjusting modes with F6.  

Versions & Settings:
 - VPX:10.8 beta 5+
![Capture](https://github.com/vpinball/vpinball/assets/20893831/9ab33ded-840a-404c-bf5f-6bcdce8e1fdf)

vbousquet:
Could you fill in the bug report template ? With this information, I can not reproduce and therefore I can't fix anything.
What would be even better would be to provide the 2 version of the table with the setup that lead to this problem

PR #831, opened 2023-09-27
Window POV fixes (and other fixes)

vbousquet:
This https://github.com/vpinball/vpinball/issues/827 report allowed me to finally identify what felt wrong with the window POV:
Problem 1 : virtual vs real world sizes:
- Screen size and Player position are real world units
- Playfield size and Bottom/Top Z offsets are virtual world units (as well as other offsets which are directly applied to the playfield bounds)
- The matching between the 2 depends on physical size of screen
- Physical sizes must be scaled to virtual world
	
Problem 2 : window mode is performing double axis fitting:
- Fitting is done on the X and the Y axis (most constraining of the both), then applying aspect ratio, so the result is not entirely stable and makes other part of the algorithm (that expects a single fitting axis) wrong

Problem 3 : in the current implementation, inclination does not take in account y offsets, nor the scaling (which has an impact is y and z scaling are not equals) which is a bug

Problem 4 : position of "zero point depth":
- The table author does not know anything about the user setting. He should only define the glass position => use table glass top/bottom.
- The player may choose to place its screen rendering at any position relatively to the real playfield/glass volume (defined by the table author)
 
These commits fixes all this by:
- Fix inclination computation
- Only fit on the y axis => perform fitting on the center playfield axis
- Remove x wnd offsets => only use x offset and x scaling, in the end, it gives the same possibilities
- Remove y wnd offsets => only use y offset and y scaling, in the end, it gives the same possibilities
- Add top/bottom glass height to the table properties (not POV)
  - Rename existing glass height as top glass height
  - Add bottom glass height, initialized with the existing glass height for existing table
  - Adjust physics, input and UI accordingly
  - Add ui for it (using cm)
- Create a display physical size setting in video setting
- Compute real to virtual world scale (defaults to 1) and apply it to player position
- Adjust UI in order to have real world sizing when scale is at 100% (allow to easily check if everything is set up correctly)

This PR also contains other miscallenous fixes (most noticeable is the long time standing blurry playfield as reported here: https://github.com/vpinball/vpinball/issues/817, the other big one being a fix to ball reflection brightness)


SmedleyB:
@vbousquet First impression of this change is amazing. I reset the Windowed POV and the default it creates looks almost exactly like what I had dialed in myself in previous versions. Player X/Y/Z react like I would expect and I appreciate how the default view is placed - how the apron is cut off and the top of the table. Looking forward to getting a 21:9 monitor installed and trying this new feature on it as well. Looks like it will work great.

studlygoorite:
Question about the Dimensions Manager, if I load a table, say Bad Cats and go to the dimension manager and double click and apply the William standard size, I see it will then automagically create the POV for me. Is this correct, because it's damn close to perfect but part of my apron is cut off and I'm thinking that the measurement for my cabinet playfield that I entered into the VPX Graphics Options may not be exact and causing this cut off, then 1 cm-ish might make the difference? Also tried Cactus Canyon and it was a little too vertically stretched but it was a quick and easy fix. Is the plan to use this Dimension Manager now, or in the future, to create the POVs for us?

vbousquet:
@studlygoorite The dimension manager has nothing to do with the POV. It's just a helper when creating a table to start with the right dimensions (need to fix though, had a look and the guns'n roses size is wrong).

SmedleyB:
@vbousquet The latest version of Cue Ball Wizard is a good one to check out (https://vpuniverse.com/files/file/15354-cue-ball-wizard-gottlieb-1992-mikcab_mod/).
The generated default POV has a lot of black space at the top of the screen - I assume there is some setting or some object on that table that it is accounting for when building the POV. But it was still a good start to the POV and easy to fix with a couple quick changes.

DanLShane:
Well, the new POV has thrown off my red/cyan anaglyph view.  The new FATHOM playfield looks not horrible, but the convergence is way, way off -- a lot farther than before. It's especially noticeable on the apron cards.
PXL_20230928_155925425

PR #883, opened 2023-10-08
Improve SDL window overlaying

vbousquet:


cwscws:
With this "fix" the standalone player on MacOS X is ruined. The improvement seems to help Windows only.

Before this the windowed fullscreen mode made it possible to display all 3 windows (main player window, b2s and dmd) on one single screen. Now this is not possible anymore. The main player window always is in front and pushes the other windows to the back and as they are not windows with frame there is no possibility to manually arrange them in front.

Can you revert that? Or a better solution? Keep in mind this should be multi system compatible.

toxieainc:
@cwscws Can you please also file this as an issue so that it doesn't get lost? Thanks

cwscws:
Done. I hope I've done it the right way.

toxieainc:
Looks good, thanks!

francisdb:
see #887

Issue #1003, opened 2023-10-27
Window opens in Narnia. This time Image Manager


iaakki:
**Describe the bug**
I had changed my display configuration and I cannot access to image manager anymore as it opens to somewhere else. Probably to some other display that is not connected anymore. 

**To Reproduce**
1. Use VPX with 3 monitors
2. Unplug 2 monitors
3. Open VPX again
4. Cannot open image manager to current monitor

**Expected behavior**
One should be able to open image manager

I tried using "r
[crash.txt.log](https://github.com/vpinball/vpinball/files/13190921/crash.txt.log)
eset window positions" feature via settings, but it made the VPX to crash.

**Versions & Settings**
 - VPX: very latest 10.8. VPinballX64_1635 release DX
 - Windows 11
 - GPU and video settings if applicable
 - If this bug report is for VPX 10.8+, please join your setting file (locate 'VPinballX.ini' in '%AppData%\VPinballX', either zip it or change its extension to '.log', then drag'n drop it here)
 - If this bug report is for VPX 10.8+, please enable log in your settingd and join your log file

Oh that made me realize VPX is not using registry anymore. So here is the issue! Image manager truly is in narnia!
![image](https://github.com/vpinball/vpinball/assets/86254776/db0665db-24ac-4689-8423-08dbac0e16e1)

So... If it is possible that VPX can write huge values into those window positions, could the system actually check that it fits to current monitor? Or set ImageMngPosX to zero if value is past current monitor dimensions?

**Additional context**

iaakki:
Actually ImageMngPosX is in 2 places in my ini. Had to change it also from line 975 under [editor].

jsm174:
I'm pretty sure those huge values are expected. Since the ini stuff was all based off the registry and DWORD, that value actually represents a small negative value.

vbousquet:
Yeah, I think it's the porting to ini which does not handle those negative values encoded as DWORD. Will have a look when I'll be back

vbousquet:
Fixed in latest, also added persisting debugger position/size

iaakki:
Very nice! Thanks!

PR #1010, opened 2023-10-28
Avoid moving window outside display area, persist debugger window pos

vbousquet:


vbousquet:
Fix https://github.com/vpinball/vpinball/issues/1003

PR #1211, opened 2023-11-29
feat: window icon

francisdb:
see https://github.com/jsm174/vpx-standalone-scripts/issues/95

_This is my first c++ code so be kind ;-)_

This is what you used to get

image

With this applied

image


francisdb:
Proper Linux desktop integration like we have for MacOS now can be added next.

jsm174:
Very cool. I will check this out tonight to make sure we are all good on the iOS / android side before merging.

francisdb:
Should CI not be validating my PR?

francisdb:
Still need to check on linux if the png instead of ico works

francisdb:
Confirmed working

francisdb:
Seems to no longer do it's thing in Ubuntu 24.04 (Wayland)

PR #1216, opened 2023-11-30
docs(standalone): document wine vbscript TypeName bug

francisdb:
see https://bugs.winehq.org/show_bug.cgi?id=55969


jsm174:
Can you please move this entry down? I wanted to keep the list in numerical order, so we it documents when we found the issue. Thanks!

francisdb:
done

PR #1226, opened 2023-12-02
fix: empty imagename in texture cache

mpcarr:
When using LoadTexture, the image name is passed to Import Image. This means the texture name load be loaded from the filename. szName will be empty and the texturecache won't have the correct name.


vbousquet:
Thanks!

PR #1303, opened 2023-12-31
Remove duplicates in Default VPinballX.ini

toccata10:
Parsing this file (for batocera configgen system) fails because of duplicates (which is always bad). I'm just cleaning up.


toxieainc:
Great, thanks!

PR #1304, opened 2024-01-02
Initial Pup support

YellowLabrador:
This is the start of a PuP implementation. It takes over the existing Backglass and DMD screens when a PuP Pack is found, this way the user doesn't need to configure physical screens again.

There is basic support for Labels and Images. I started working on playing videos using ffmpeg. There is some code checked in but it's not done therefore videos are disabled for now.

Figuring out the correct semantics of the PuP API will be tricky. The documentation is sparse, I've read the scripts on a few tables to get started but it's anyone's guess what calls like SetBackground() does. 

I started on this project to get Munsters working. That's the main table I've been using to test the code. It does run but the DMD is kind of a mess. Background images don't switch correctly probably because calls like SetBackground() are not implemented. Sounds seem to play just fine. When I say "it runs" I mean it runs on an Arm Mac. On Linux SDL is giving an error and only the first frame of the first window renders correctly, not even the playfield shows anything. I'm sharing the code to see if anyone has an idea on how to get that working. My Linux box is on Wayland.

This implementation expects to find the pup pack for a table in the same directory the VPX is. So in the example below tablex looks for the Pup Pack content in a directory called `tablex_pupfolder`, therefore that directory must exist in the same place the VPX is.

```
/tables/tablex/table.vpx
/tables/tablex/table.ini
/tables/tablex/[tablex_pupfolder]/...
/tables/tablex/pinmame/roms/[gamename].zip
```


standalone/inc/pup/SDL_ttf.c
```diff
@@ -0,0 +1,4223 @@
+/*
```
francisdb, line 1: Any reason to not pull this in as a dependency?

jsm174:
This is fantastic. I am in the middle of a dmd refactor, so when I finish, I'll figure out how to get this in 10.8.1 and backported into standalone.

Thanks!

freezy:
Wait, does that mean we'll get open source pup now?

standalone/inc/pup/SDL_ttf.c
YellowLabrador: It's a single .c file, the build is a bit complex it looked simpler to just copy it.

YellowLabrador:
@freezy yes, emphasis on "will get". It's a long way from being fully functional.

francisdb:
Can you mark this commit as wanting to merge to `10.8.1` and rebase on `10.8.1`? `10.8.1` contains the latest changes and standalone has been merged into it.

Some of the commits on `10.8.1` are backported, squashed and force pushed to the `standalone` branch for existing 10.8.0 standalone users.

francisdb:
Would you mind updating the pull request description with an example on where to place the pup pack files to test this?

Is this compatible with the folder per table setup? Eg.
```
/tables/tablex/table.vpx
/tables/tablex/table.ini
/tables/tablex/table.directb2s
/tables/tablex/music/...
/tables/tablex/PUPVideos/[gamename]/...
/tables/tablex/pinmame/roms/[gamename].zip

/tables/tabley/...
```

standalone/inc/pup/SDL_ttf.c
francisdb: SDL_ttf used to be part of the build. But now I see it was as binary so also not ideal.

54d8b8400b5763f0e64d3edb3084c9c0be65eed9
224c661129b36cc90e218c42f63da22016319b15

jsm174: SDL_ttf wasn't being used by anything, but we removed it. Since this will use it, that code will come back. We do want everything as libs and so's.

francisdb: > SDL_ttf wasn't being used by anything, but we removed it. Since this will use it, that code will come back. We do want everything as `.dylib` and `.so`.

Not sure what you want to say here? There are 3 options:

* copy over code to the repo like this pr does.
* store binaries of libraries in the repo like used to be the case for SDL_ttf (so/dll/dylib)
* fetching dependencies as binary or  fetch code and compile like external.sh does

Which would be preferred here? I would think the third option is preferable?

jsm174: Sorry for not being clear enough. 

we will update this PR and build upon it to do the `external.sh` way.

YellowLabrador: I will look into it this weekend. After a bit of Googling the only reference about the problem I'm seeing on Linux is for someone else using SDL_ttf. I will try to build it without SDL_ttf and test it on Linux. If I don't see the issue anymore I might remove it all together and call FreeType directly, otherwise I will make it a dependency.

@jsm174, any thoughts on using cmake for dependency management instead of a shell file? I played a bit with FetchContent_Declare to download a zip/tar and build it. Using OVERRIDE_FIND_PACKAGE makes sure cmake doesn't grab a version that might be installed in the system.

jsm174: Yes, I thought about using cmake to fetch dependencies. But the differences between building for different platforms, (for example compiling libserialport), it just makes the cmake files too complicated, and they are already too complicated.

@vbousquet started moving things around into `third-party/runtime-libs` and `third-party/build-libs` and I think this is exactly what we needed.

If you look at the build processes for the following projects:

https://github.com/vpinball/libaltsound
https://github.com/jsm174/libdmdutil
https://github.com/ppuc/libzedmd/tree/libserialport

We tried to stick to the same approach, and it seems to work well.

I'm going to start slowly moving standalone to this as well.

YellowLabrador: I updated the PR and moved SDL_ttf to be built in external.sh along with other dependencies

jsm174:
@YellowLabrador - I just wanted to let you know I have not been ignoring this PR.  I'm getting close to pulling in this:

https://github.com/jsm174/vpinball/commits/libdmdutil/

Each component now has its own thread for rendering, so there will be some changes needed. Hopefully by the weekend, start tackling this.

YellowLabrador:
@jsm174 I rebased this request on your latest standalone and reimplemented it to follow the new design. There is now a single render thread for handling the Pup Pack content similar to what you did for B2S. I only tested it on a Arm Mac so I pushed the build changes only for that platform.

francisdb:
```
vpinball/standalone/inc/pup/PinUpPlayerLabel.h:4:13: fatal error: SDL_ttf.h: No such file or directory
    4 |    #include "SDL_ttf.h"
```

Can you maybe enable github actions on your side so we can see if the build should work or not? Still think it's annoying that pr's are not getting built on the vpinball side.

francisdb:
Looks like you did not add SDL_ttf to the linux build

jsm174:
> @jsm174 I rebased this request on your latest standalone and reimplemented it to follow the new design. There is now a single render thread for handling the Pup Pack content similar to what you did for B2S. I only tested it on a Arm Mac so I pushed the build changes only for that platform.

@YellowLabrador  Fantastic!  Fyi, I think maybe my threading updates are not 100% correct. It works great on Linux, but ClarkKent noticed significant stutter on MacOS (very noticable on Cyclone). 

SDL is not very clear on threading rules, but from the docs for [SDL_RenderPresent](https://wiki.libsdl.org/SDL2/SDL_RenderPresent#thread-safety) 

`
You may only call this function on the main thread. If this happens to work on a background thread on any given platform or backend, it's purely by luck and you should not rely on it to work next time.
`

So I need to revisit.

@francisdb - As for SDL_ttf, let me go back and add it to external.sh from what we had in the past and we can do one more rebase. This way we get mobile covered as well.

YellowLabrador:
> ```
> vpinball/standalone/inc/pup/PinUpPlayerLabel.h:4:13: fatal error: SDL_ttf.h: No such file or directory
>     4 |    #include "SDL_ttf.h"
> ```
> 
> Can you maybe enable github actions on your side so we can see if the build should work or not? Still think it's annoying that pr's are not getting built on the vpinball side.

@francisdb I did not update the Linux build scripts since moving SDLttf to external.sh. I did mention it in the comment. That said I shouldn't break the build. @jsm174 should I update the all the build files in my PR or are you doing that separately?

jsm174:
> @jsm174 should I update the all the build files in my PR or are you doing that separately?

I will do it if you don't mind.

YellowLabrador:
Go for it. This PR is already building SDLttf in external.sh for macos_arm64. Take a look at that script if it's helpful.

jsm174:
luckily I kept copies of standalone throughout the rebases

but for example, all the android, ios, tvos, etc was all done:

https://github.com/jsm174/vpinball/commit/4c33a655cf9a14aa3b326035c9b7a9366733a315#diff-4d80d684725a69f2d382c0dac42cd1cf8d86c47868ccc4074a6d697438b49b38R121

YellowLabrador:
Nice. Do let me know when I can rebase my branch again.

Are you ok with using ffmpeg for the video player? That's also currently in the build script. I can continue to work on the video player code if ffmpeg will be underlying library.

jsm174:
> Nice. Do let me know when I can rebase my branch again.
> 
> Are you ok with using ffmpeg for the video player? That's also currently in the build script. I can continue to work on the video player code if ffmpeg will be underlying library.

10.8.1 and standalone now have ttf added back for all platforms. 

Do you think we could rebase against 10.8.1 instead of standalone branch? Its easier to cherry-pick a 10.8.1 commit into standalone. 

Also, I really would like to look into [SDLKitchenSink](https://github.com/katajakasa/SDL_kitchensink), but I will dig into what you have now.

YellowLabrador:
> 10.8.1 and standalone now have ttf added back for all platforms.
> 
> Do you think we could rebase against 10.8.1 instead of standalone branch? Its easier to cherry-pick a 10.8.1 commit into standalone.
> 
> Also, I really would like to look into [SDLKitchenSink](https://github.com/katajakasa/SDL_kitchensink), but I will dig into what you have now.

@jsm174 I submitted another PR on top of 10.8.1 - https://github.com/vpinball/vpinball/pull/1366. I also remove the dependency on ffmpeg and deleted the video player.

jsm174:
superseded by https://github.com/vpinball/vpinball/pull/1690

PR #1388, opened 2024-01-30
Update standalone README.md

rh6423:
cmake didn't respect $PATH when I did a build, and used the wrong and old bison in /usr/bin. I've added a few lines about how to explicitly specify bison path to cmake command.

### DIFF
--- a/standalone/README.md
+++ b/standalone/README.md
@@ -145,7 +145,12 @@ cmake -DCMAKE_BUILD_TYPE=Release -B build
 cmake --build build -- -j$(sysctl -n hw.ncpu)
 ```
 
-*Note:* Make sure `SDL2`, `SDL2_image`, and `freeimage` are not installed by `brew`, otherwise you may have compile issues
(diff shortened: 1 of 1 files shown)


jsm174:
Hello. Thanks for the PR! 

Two things:

- Unfortunately, the standalone branch isn't taking PRs just because of how I've been rebasing. We can do PRs on the 10.8.1 branch.

- Also, this isn't necessary as we already have an entry to update the path for bison.

```
brew install autoconf automake cmake bison curl
export PATH="$(brew --prefix bison)/bin:$PATH"
git clone -b standalone https://github.com/vpinball/vpinball
.
.
.
```

I'm not sure why this isn't working on your system, but it's been working since we started the project.

We also do the same thing on the CI and no issues there as well:

https://github.com/vpinball/vpinball/blob/0d0dcff884f71eecab96111b63db6b5586ae4b2c/.github/workflows/vpinball.yml#L160-L162

PR #1410, opened 2024-02-04
standalone: add support for virtualdmd in libdmdutil

jsm174:


francisdb:
What exactly is virtualdmd?

jsm174:
It's probably not the greatest name, but libdmdutil now outputs separate memory buffers that represent an rgb24 and level data. This way standalone can request two of them, one for the in game dmd and one for an external window. Each can know independently when there is an update.

PR #1411, opened 2024-02-04
standalone: add RenderMode for external windows

jsm174:


src/assets/Default VPinballX.ini
```diff
@@ -18,13 +18,15 @@ HighDPI =
 VPRegPath = 
 PinMAMEPath = 
 PinMAMEWindow = 
+PinMAMEWindowRenderMode =
```
francisdb, line 21: Some docs on what options are available and what the default is would be welcome.

jsm174: Accepting PRs!

PR #1423, opened 2024-02-07
standalone: rework external rendering windows again

jsm174:
I shouldn't be fighting the system. 

The [docs](https://wiki.libsdl.org/SDL2/SDL_RenderPresent) clearly say:

```You may only call this function on the main thread. If this happens to work on a background thread on any given platform or backend, it's purely by luck and you should not rely on it to work next time.```

So this PR moves everything back into the main thread.  It also splits up the graphics classes into `SurfaceGraphics` and `RendererGraphics`

The window manager keeps track of all windows created in vbscript before the table is officially started.

After this message is logged:

`2024-02-07 13:33:43.769 INFO  [11514] [Player::Init@1837] Startup done`

We then create them.

I still have stutter in Batocera, but I will continue working to get performance acceptable. (no response needed, just keeping you informed: @toxieainc, @vbousquet)


francisdb:
Runs smooth on Ubuntu 23.10 Radeon xwayland full hd here. Will check on my cab how perf is there with nvidia 4k.

vbousquet:
I didn't have the time to look to this. For me the aim is (when enough free time is available, and likely 10.8/3.6 is out) to move to bgfx. When using it, you will get a clean support of multiple output (see here: https://github.com/bkaradzic/bgfx/tree/master/examples/22-windows). Supporting multiple outputs across multiple rendering backends and multiple platform is really a complex thing. I think it really needs a broadly tested library to get stable results

PR #1436, opened 2024-02-09
feat(standalone): ball control

### LINKED ISSUE #1281: Standalone: ball control not working
**Describe the bug**
Normally you can control the ball with the mouse if ini setting `BallControlAlwaysOn` is true, this is not working on standalone.

Probably because the related code is in a `#ifdef _WIN32`

https://github.com/vpinball/vpinball/blob/75d89c2abd72bdb54c0ab8f9b6c5b4f1c9d7f29f/pininput.cpp#L484-L505

francisdb:
fixes #1281 

To test this modify your ini file to contain the following values

```ini
[Editor]
ThrowBallsAlwaysOn = 0
BallControlAlwaysOn = 1
```

Mind the uppercase `E` in Editor, I had an old version of the ini with lowercase e causing the setting to not be applied.

Then drag your mouse with left mouse button on the playfield to have the ball follow you. For a quick move double click somewhere to have the ball teleport and fall down on that location.

### DIFF
--- a/pininput.cpp
+++ b/pininput.cpp
@@ -30,6 +30,11 @@
 #define INPUT_BUFFER_SIZE MAX_KEYQUEUE_SIZE
 #define BALLCONTROL_DOUBLECLICK_THRESHOLD_USEC (500 * 1000)
 
+constexpr DWORD D_MOUSE_DRAGGED_LEFT = 1;
+constexpr DWORD D_MOUSE_DRAGGED_RIGHT = 2;
+constexpr DWORD D_MOUSE_DRAGGED_MIDDLE = 3;
+constexpr DWORD D_MOUSE_MOVED_LEFT_DOWN = 4;
+
 
 PinInput::PinInput()
 {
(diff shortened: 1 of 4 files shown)


francisdb:
Still a bit confused on how the throw balls is supposed to work but for my part this is ready for review.

jsm174:
LGTM, but I don't have an easy way to test windows. Assuming windows uses old windows specific way anyway?

Would be nice to have someone check on android to make sure the touch stuff still works. I think I remember touch events might be wrangled into mouse events, but I'm not sure.

I'll defer to an additional review from @toxieainc.

toxieainc (approved):

toxieainc:
Windows stays as it is, yes. The rest looks good to me, assuming that it was also tested.  ;)

francisdb:
One day we'll have to make some clear separation instead of all those #ifdef.

jsm174:
> One day we'll have to make some clear separation instead of all those #ifdef.

I know it's a lot to take in, but until / if the windows ui gets abstracted out, this is actually pretty clean compared to what we could have done.

Then again, I've looked at the code base every day for the past 2 years.

Issue #1372, opened 2024-01-22
problem on 10.8 > rev 1347 on batman dark knight


TchouK62:
**Describe the bug**
When the scarecrow mission is launch, pushing right flipper validate the crane without contact with the ball.

**To Reproduce**
1. Go to scarecrow
2. Push right flipper button
3. Crane is validated and goes to the next point.

If the bug appears when playing a specific table, give a link to a place where the exact version of the table file can be downloaded.
Appears on NZ & TT version, and Bord mod too.

**Versions & Settings**
 - VPX: 10.8 > rev 1347
 - Windows version :11
 - GPU all tested (1660Ti and 6700XT)

**Additional context**

toxieainc:
Did you also update VPinMAME or which version do you use there?
And is this about 10.8 or 10.8.1?

TchouK62:
Vpinmame is up to date. I'm on vpx 10.8. I wasn't aware that 10.8.1 exists.

toxieainc:
So which exact versions do you test (see VPX about dialogue)? And VPM ist 3.5 or a 3.6 beta build (in case the latter, please also post the version)?

TchouK62:
Hey
All versions after 10.8 rev 1347 has the problem.
I'm using VPinMAME-3.6-720-e4df859-win-x64

toxieainc:
Okay, so inbetween beta5 and beta6 then, thanks!

TchouK62:
You're welcome. If you need more tests I can do, just ask

TchouK62:
hey @vbousquet I still have the bug in RC1. Is it a vpinmame problem instead of vpx ?

GroniPinball:
Oh Nevermind...Just found the updated Sam.vbs

TchouK62:
> Oh Nevermind...Just found the updated Sam.vbs

Where dis you fond that? That solved the bug?

GroniPinball:
Yes the updated Sam.vbs solved the Scarecrow Crane Bug. 

https://github.com/vpinball/vpinball/blob/master/scripts/sam.vbs

wiesshund:
@toxieainc @vbousquet 
Please undo this reversion, it is NOT a sam VBS error
It is a table script issue, the script needs updated
the SAM.VBS already has a contention for this


table script needs

Const cSingleLFlip = 0
Const cSingleRFlip = 0

that is all.

Technically only needs 
Const cSingleRFlip = 0

[...]

PR #1454, opened 2024-02-12
Disable staged flipper for SAM as it breaks table without them

vbousquet:
Fix https://github.com/vpinball/vpinball/issues/1372

### DIFF
--- a/scripts/sam.vbs
+++ b/scripts/sam.vbs
@@ -69,18 +69,8 @@ Function vpmKeyDown(ByVal keycode)
 	vpmKeyDown = True ' Assume we handle the key
 	With Controller
 		Select Case keycode
-			Case LeftFlipperKey
-				.Switch(swLLFlip) = True : vpmKeyDown = False : vpmFFlipsSam.FlipL True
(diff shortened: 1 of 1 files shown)


wiesshund:
@toxieainc @vbousquet 
Please **UNDO** this reversion, it is NOT a sam VBS error
It is a table script issue, the script needs updated
the SAM.VBS already has a contention for this


table script needs

Const cSingleLFlip = 0
Const cSingleRFlip = 0

that is all.

Technically only needs 
Const cSingleRFlip = 0

but the other hurts nothing since L isnt used either.

Hence me asking for details before this was done but no one would give any

wiesshund:
ALL tables that have any conflicts with staged flipping do to use of said circuits simply need the offending circuit (one or both flippers) set to be disregarded.

And not all SAM tables have any conflict.

Const cSingleLFlip = 0
Const cSingleRFlip = 0

Or whatever combination suits the table but you dont normally have to specify the value of 1 so you only need to set the ones you want deactivated.

just throw it in top of script after
Option Explicit
Randomize

PR #1462, opened 2024-02-17
STANDALONE: Added analog input event support

omigeot:
Added support for SDL_CONTROLLERAXISMOTION, should allow analog axis (plunger, nudges) to be used on standalone version.


omigeot:
From my last tests, nudges don't work, there's something else in their way.

PR #1489, opened 2024-02-25
standalone: add BGFX builds for Linux x64/aarch64 and Android

jsm174:
RPI and RK3588 moved into linux-aarch64
linux moved to linux-x64
android moved to android-arm64-v8a


francisdb:
Aha, so I can run the RPI version on Asahi linux on my laptop?

PR #1494, opened 2024-02-26
Better player close, log stutter frame

vbousquet:


wintimer.h
```diff
@@ -36,30 +36,69 @@ class FrameProfiler
...
          m_profileTotalData[i] = 0;
       }
+      // Clear worst frames
+      m_leastWorstFrameLength = 0;
+      for (int i = 0; i < N_WORST; i++)
+         memset(m_profileWorstData[i], 0, sizeof(m_profileWorstData[0]));
    }
 
-   void NewFrame()
+   void EnableWorstFrameLogging(bool enable) { m_logWorstFrame = true; }
```
francisdb, line 69: The `enable` argument is not used here.

PR #1501, opened 2024-02-27
Audit: report timer below framerate that will cause stutters

vbousquet:


src/parts/pintable.cpp
```diff
@@ -8200,8 +8200,8 @@ void PinTable::AuditTable() const
       case eItemGate: tdr = &((Gate *)part)->m_d.m_tdr; break;
       default: break;
       }
-      if (tdr && tdr->m_TimerEnabled && tdr->m_TimerInterval != -1 && tdr->m_TimerInterval != -2 && tdr->m_TimerInterval < 5)
-         ss << ". Warning: Part '" << part->GetName() << "' uses a timer with a very short period of " << tdr->m_TimerInterval << "ms. This will likely causes lag and the table will not support 'frame pacing'.\r\n";
+      if (tdr && tdr->m_TimerEnabled && tdr->m_TimerInterval != -1 && tdr->m_TimerInterval != -2 && tdr->m_TimerInterval < 17)
+         ss << ". Warning: Part '" << part->GetName() << "' uses a timer with a very short period of " << tdr->m_TimerInterval << "ms, below a 60FPS framerate. This will likely cause stutters and the table will not support 'frame pacing'.\r\n";
```
francisdb, line 8204: Would be nice if this were logged on warning level so it stands out in the logs.

vbousquet: This is meant for the editor version (not standalone) and is displayed in a dialog prompt so the user knows it exists

PR #1502, opened 2024-02-27
feat: improve audit log

francisdb:
- Correct line endings
- Log on warning level if warnings
- Enables audit on game start for standalone


vbousquet:
Thanks for the contrib, this cleanups things.
I'm not sure we should run AuditTable on startup like this. This is meant to be an helper for table author to improve and be aware of changes. Not something meaningful for a standalone run

vbousquet:
Also, I don't think this should be logged as warning since in fact these are just informations about the table, not real warnings about the application behavior

francisdb:
I think it's still interesting for ppl debugging tables on standalone as you have no editor.

Alternative options:
* `vpinball -audit foo.vpx`
* a general vpinball flag like `-debug` that enables this log when running a game
* adding this to the live editor as an option in the menu

vbousquet:
Just add a look and Windows UI does not handle `std::endl`, it needs the \r.
Regarding adding some UI edition capability to standalone, I think this would be nice, and the LiveUI is a seed for that but I really don't think we should bloat the code by adding things not cleanly. Calling AuditTable each time is really not what the function is designed for (this function is supposed to evolve, at some point it may take a long time to complete or generate long logs,...) and adding a commandline option does not seem as a good option, also knowing that no edition will be possible afterward. In the end, the table will be adjusted on a platform that supports editing.

PR #1503, opened 2024-02-27
fix: sdl input

francisdb:
was broken by 8b5554a00fbd6add2f5a8237d3088f45de9288de

On standalone/sdl any key/controller input was not being handled because a large part of the code was not triggered.

### DIFF
--- a/src/core/pininput.cpp
+++ b/src/core/pininput.cpp
@@ -842,6 +842,7 @@ void PinInput::HandleSDLEvents(DIDEVICEOBJECTDATA* didod)
             break;
          }
          break;
+      }
    #endif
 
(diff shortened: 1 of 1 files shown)


francisdb:
I seem to still have issue when pressing `esc` and then trying to go to table options with the `up`-`down` keys. Not sure if that is related.

francisdb:
@vbousquet mind merging this as currently keyboard input is completely broken?

francisdb:
@jsm174 mind reviewing this as currently keyboard input is completely broken on standalone?

francisdb:
Looks like this was fixed in a different way
see #1509

PR #1504, opened 2024-02-27
docs(standalone): fix linux path in readme

francisdb:
Also needs to be applied to 10.8.1 I guess


jsm174:
can you please switch this 10.8.1? It won't let me change the base branch.

jsm174:
merged manually into standalone branch.

PR #1527, opened 2024-03-03
Display playfield slope

### LINKED ISSUE #1507: Show 'phasics' playfield angle in Tweak menu (F12) when setting difficulty
Difficulty can be adjusted in the Tweak menu (F12) in the range between 0 and 100. This maps to the min and max playfield angle that is entered in VPX editor and used for physics calculation.
It would be nice to see the calculated playfield angle, based on the chosen difficulty (0 - 100) in the F12 menu.
At least, it would be nice to see the static min and max value for the angles in F12 menu. This way, the user can at least estimate what the resulting angle would be.

Complementary to this feature request is this other feature request: #1295 
When setting 'window top z offset' and 'window bottom z offset' and thus setting at what angle the table is shown on screen, it would be nice to show the 'visual' angle of the playfield as well in F12 menu. This can be calculated from the physical angle of the TV (monitor), which is entered in the Video options of VPX editor plus 'window top z offset' and 'window bottom z offset'.
This allows to compare the playfield angle to angle from the real world or to the angle, used for physics calculation.

Nicals:
Fixes #1507

This PR changes the display value of the Difficulty settings of TweakUI to show playfield slopes to the user.

playfield-slope

### DIFF
--- a/src/parts/pintable.cpp
+++ b/src/parts/pintable.cpp
@@ -8904,6 +8904,11 @@ void PinTable::SetHeight(const float value)
     m_bottom = value;
 }
 
+float PinTable::GetPlayfieldSlope() const
+{
+   return m_angletiltMin + (m_angletiltMax - m_angletiltMin) * m_globalDifficulty;
+}
+
 STDMETHODIMP PinTable::get_Height(float *pVal)
 {
    *pVal = GetHeight();
@@ -9554,8 +9559,7 @@ STDMETHODIMP PinTable::put_Gravity(float newVal)
       SetGravity(newVal);
       const float minSlope = (m_overridePhysics ? m_fOverrideMinSlope : m_angletiltMin);
       const float maxSlope = (m_overridePhysics ? m_fOverrideMaxSlope : m_angletiltMax);
-      const float slope = minSlope + (maxSlope - minSlope) * m_globalDifficulty;
-      g_pplayer->m_physics->SetGravity(slope, m_overridePhysics ? m_fOverrideGravityConstant : m_Gravity);
+      g_pplayer->m_physics->SetGravity(GetPlayfieldSlope(), m_overridePhysics ? m_fOverrideGravityConstant : m_Gravity);
    }
    else
    {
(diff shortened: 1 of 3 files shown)


toxieainc:
Why not both (difficulty and slope)?

vbousquet:
Yes, difficulty is not easy to understand as it impacts slope but also lots of other parts (kickers,...). Just showing slope could be misleading

Nicals:
Rolled back the changes and added slope info and difficulty definition in the settings label. It may help users understand what all this is about.

diff

I used C string functions as they seemed to be the norm in this project. Don't you prefer something more C++ like std::stringstream ?

toxieainc:
There is no real 'standard' as VPX is a mixture of up to 20+ years old code with about a dozen main devs over the years.
So the general rule is: Use whatever makes sense and fits the surrounding code i'd say.

toxieainc:
@Nicals So do you want to still change things, or should i merge?

Nicals:
Lets merge it. As you say, it fits the surrounding code.

src/parts/pintable.cpp
```diff
@@ -9557,8 +9562,7 @@ STDMETHODIMP PinTable::put_Gravity(float newVal)
       SetGravity(newVal);
       const float minSlope = (m_overridePhysics ? m_fOverrideMinSlope : m_angletiltMin);
       const float maxSlope = (m_overridePhysics ? m_fOverrideMaxSlope : m_angletiltMax);
-      const float slope = minSlope + (maxSlope - minSlope) * m_globalDifficulty;
-      g_pplayer->m_physics->SetGravity(slope, m_overridePhysics ? m_fOverrideGravityConstant : m_Gravity);
+      g_pplayer->m_physics->SetGravity(GetPlayfieldSlope(), m_overridePhysics ? m_fOverrideGravityConstant : m_Gravity);
```
toxieainc, line 9565: @Nicals Is this one here really correct? Cause seems like m_overridePhysics is ignored now.

Nicals: You are right, I did not knew this physics override settings. Should be fixed now, I clearly went too quickly on this one.

toxieainc: Thanks!

PR #1557, opened 2024-03-11
standalone: overhaul externals caching strategy

jsm174:
This PR allows for incremental updates when caching externals for caching, ie, instead of rebuilding everything if libpinmame changes, it will only build libpinmame.


francisdb:
Thanks Jason!

Issue #1409, opened 2024-02-04
Changing sounds from Table to Backglass in sound manager causes sounds to not be playable


gonzonia:
**Describe the bug**
Changing sounds from Table to Backglass causes sounds to not be playable. The sounds don't play when the table is played and they don't play in sound manager when selected and Play is pressed. 

**To Reproduce**
1. Open DAREDEVIL_1.2.vpx
2. Open Sound Manager
3. Scroll down to sounds with the name "Quote*"
4. Select all sounds with the name "Quote*"
5. Click on Position/Volume
6. Change to Backglass
7. Click OK
8. Save Table
9. Quit Visual Pinball

If the table is played at this point, the sounds do not play. 
After saving:
1. Open DAREDEVIL_1.2.vpx
2. Open Sound Manager
3. Scroll down to a sound with the name "Quote*" that is set to Backglass. 
4. Click Play and no sound plays.
5. Click on Position/Volume
6. Set to Table
7. Click OK
8. Select the same sound that is now set to Table,
9. Click Play
10. Sound will play through Table.
 
DAREDEVIL_1.2.vpx:
https://drive.google.com/file/d/18uEMuj0ZmTF720b2VWUrKZFG0YT2V5D4/view?usp=drive_link

**Expected behavior**
Sounds should play through backglass after being changed. 

**Versions & Settings**
 - VPX: v10.8.0 Beta (Rev. 1873 (bdd0673), windows OpenGL 64bits)
 - Windows 11 Pro (Version 22H2 (OS Build 22621.3007)
[VPinballX.log](https://github.com/vpinball/vpinball/files/14155205/VPinballX.log)
[vpinball.log](https://github.com/vpinball/vpinball/files/14155213/vpinball.log)

vbousquet:
Hi, just tested and done the sequence you propose and everything worked for me. Do you still have issues with latest build from master (available from github action, `master` labeled build) ?

vbousquet:
I have just pushed a fix for this. Could you confirm that it is now working as intended ?

gonzonia:
It’ll be a few days. But I’ll check as soon as I can

Issue #287, opened 2023-02-04
backglass .ogg files assigned to table for people using 2.1 Stereo setup


RusstyT:
There have been reports from people playing TimeLord2.0 & TimeLord 1.0 that all .ogg backglass files are being played from the table speakers. (See Crackers workaround below to reallocate the sounds to the backglass. This has proven to be successful stop gap measure). Common things between people with this issue is that multiple sound cards are being used and stereo 2.1 is selected from the sound manager. This can occur with people using Playfield speakers or usb sound cards.
This issue has been occurring with the VPX l0.7.2 Final (Revision366(f94da1e),32bit and in previous versions of 10.7.1 & 10.7.2 before that. People have also mentioned that they have had the same problem on IT and JP's wrath of olympus and dead pool but Crakkers (TimeLord table tester) mentioned the problem didn't occur in games using .oog files that Gedankekojote produced.

Crakkers work around
My workaround 'fixes' the issue with all of the sounds coming from the table (SSF speakers).
It involves moving all of Russty's .ogg files to the backglass. It seems that certain files have the power of veto and having just one of these set to "table" in the sound manager causes all sounds to come from the table. However, not all .ogg files seem to have this magical power, so rather than moving dozens & dozens of files individually to find the culprits, I just moved the lot.
Any .wav files are ok to stay on the table, plus any .ogg files beginning with C:\Users\Kai.... are also fine.......don't ask me why.
Anyone wanting a copy of my workaround can just send me a PM and I'll give you a download link.
I'm not sure if this workaround will help with the other audio issues being mentioned, but it certainly fixes the "all sounds coming from the table" issue.

thankyou for your work
regards
Russ Tolhurst (RusstyT)

vbousquet:
I have just pushed a fix for this. Could you confirm that it is now working as intended ?

RusstyT:
Can confirm that version 6L works fine according to Darthvito. Not everyone suffered from this issue of sounds being allocated from the backglass to the table but was quite devastating for people using SSF when this occurred.

Issue #64, opened 2022-06-04
Sounds assigned from table to backglass do not play


RichardL64:
If you assign a sound, using the sound manager, to backglass - e.g. move the knocker sound to play from the backglass rather than table - when run in F6 or F5 mode from the editor it will not play.

**To Reproduce**
Edit a sound in the sound manager - click on the 'backglass' option,
Run the table via F5 -  the sound will not play
Run the table via command line - i.e. a front end - the sound will play

**Expected behaviour**
Sounds assigned to the backglass should play in all modes.

**Versions**

VPX: 10.7.0 Final (Revision 367 (https://github.com/vpinball/vpinball/commit/e242d7c93fd8c427c4a3d4852277b7f088c883c3) 32bit)
Windows 10

sverrewl:
It's a long standing but. Save the table, EXIT completely; and it is just fine.

vbousquet:
I have just pushed a fix for this. Could you confirm that it is now working as intended ?

PR #1568, opened 2024-03-15
Reinitialize Sounds each time their output device change

vbousquet:
Fix https://github.com/vpinball/vpinball/issues/64
Fix https://github.com/vpinball/vpinball/issues/1409
Should also fix https://github.com/vpinball/vpinball/issues/287


vbousquet:
I have just pushed a fix for this. Could you confirm that it is now working as intended ?

RusstyT:
Thanks for doing this this Vincent. It doesn't cause an issue on my 7:1 set up so I have know way of checking directly. I'll ask some others who it has caused an issue.
Russ

Issue #1558, opened 2024-03-11
Add build number to the COM Version method of VPX


JockeJarre:
**Is your feature request related to a problem? Please describe.**
To help the flipper table creators, not having to tell in all forums which minimum version to use for VPX, VPM and B2S

**Describe the solution you'd like**
Currently the Version method **returns an integer** as 10800 for VPX 10.8, it would be an improvement if the build number would be added as decimal value like 10800.1923 (allowing 4 decimal characters), meaning one would have to change the method to **return a double** instead.

**Describe alternatives you've considered**
I have been thinking of using a separate Version method. But rather I would keep the same. 

**Additional context**
I have tested the current usage of VPBuildVersion and version in the local vbs files, and the same if statements still work perfect.

Some examples from "core.vbs":

1.  ElseIf VPBuildVersion >= 900 and VPBuildVersion <= 999 Then
2.  ElseIf VPBuildVersion >= 10000 Then
3.  Version >= 10700

Using this it will be possible to test using the build number as decimal:
1.  Version >= 10800.1911   is True  (where Version is a double with value 10800.1923)

The same also works in older versions
1.  Version >= 10800.1911  is True (where Version is an integer with value 10800)

francisdb:
Is it documented what 10700 represents? 10.70.0, 107.0.0 or 10.7.00. What happens when 10.10.0 is next? Floats can't represent all values and are a lossy format. Maybe it's time for something like semantic versioning?

JockeJarre:
the COM method Version and VPBuildVersion (get_Version in src/parts/pintable.cpp:

	*pVal = VP_VERSION_MAJOR * 1000 + VP_VERSION_MINOR * 100 + VP_VERSION_REV;

So you are right, VP_VERSION_MAJOR=10 , VP_VERSION_MINOR=8 , VP_VERSION_REV, means VP_VERSION_MINOR can only be one significant number. 

One could concider changing it, though it might mean a lot of changes needed. Or start to use the VP_VERSION_REV, to increase the size of the version universe?

toxieainc:
We don't need to. Roadmap-wise, the stuff Vince is planning will justify a move to 11 soonish anyway.

toxieainc:
Closing then here, right??

PR #1580, opened 2024-03-17
Add build number to the COM Version method of VPX #1558

JockeJarre:
Currently the Version method returns an integer as 10800 for VPX 10.8, it would be an improvement if the build number would be added as decimal value like 10800.1923 (allowing 4 decimal characters), meaning one would have to change the method to return a double instead.


JockeJarre:
We concluded that it isn't good to include the build number in the COM methods, as it would encourage more usage of the builds rather than the official releases.

PR #1612, opened 2024-03-28
10.8.1

vbousquet:


src/math/bbox.h
```diff
@@ -80,6 +80,8 @@ inline bool Intersect(const RECT &rc, const int width, const int height, const P
       return (p.x >= rc.top*width / 100 && p.x <= rc.bottom*width / 100 && p.y <= height - rc.left*height / 100 && p.y >= height - rc.right*height / 100);
 }
 
+/*
+// Disable as it does not gives the expected results as the non SSE implementation below
```
vbousquet, line 84: @toxieainc I stumbled on this bug while fixing LiveUI picking. I do not have the skills to implement Rect3D intersection with SSE (never used SSE instruction set). Still, it is a bit sad to disable it altogether so if you have some free time to give it a try, that would be nice :)

PR #1644, opened 2024-04-07
fix(standalone): DateDiff also accepts string as second arg

francisdb:
Found in Batman 66

ref https://github.com/jsm174/vpx-standalone-scripts/issues/284


francisdb:
```vbscript
bDebounceTime(idx)=DateDiff("s", "01/01/1970", Now()) + Timer + 0.500
```

francisdb:
@jsm174 I guess some more work needs to be done to make sure this is not overwritten in the next wine import?

PR #1645, opened 2024-04-08
Android - Controlling the web server via Settings screen

YellowLabrador:


jsm174:
This is great!

If possible, really would like to stick to `PLOG` for everything and not try to do `#define` swaps.

```
#ifdef __ANDROID__
#define LOG_TAG "WebServer"
#include "../../../src/utils/AndroidLog.h"
#define WEBLOGE ALOGE
#define WEBLOGI ALOGI
#else
#define WEBLOGE PLOGE.printf
#define WEBLOGI PLOGI.printf
#endif
```

YellowLabrador:
ok. I will have a PR for separating the PLOG initialization from main.cpp so it can be called from both the web server and main.cpp. That will allow for PLOG usage even when a table is not running, which is the case here. After that PR is done I will update this one to use PLOG directly.

toxieainc:
@jsm174 Is this still something that could be merged/ported, or outdated nowadays?

jsm174:
With the latest android version, this has been resolved.

PR #1660, opened 2024-04-16
fix(ui): go back to game instead of live editor

### LINKED ISSUE #1646: Pressing ESC twice no longer brings you back to the game
Linux 10.8.1 branch

Previously you could press `esc` to get the menu, pressing it a second time would take you back to the game. Now this brings you to the live editor.

francisdb:
fixes #1646

Not sure I understand all these flags but this fixes the issue I have where double `esc` brings you to the live editor instead of back to the game.

_There is still some weirdness in the live editor where you first need to select something before you can use `esc` to exit it._


vbousquet:
Cleanly fixed in latest

francisdb:
Thanks

PR #1662, opened 2024-04-16
Improve flex asset loading error message

francisdb:
before
```
2024-04-16 23:19:36.580 WARN  [115461] [AssetManager::Open@255] Asset not loaded: d_border
```

after
```
2024-04-16 23:19:36.580 WARN  [115461] [AssetManager::Open@255] Asset not loaded: d_border - Unsupported image format
```

### DIFF
--- a/standalone/inc/flexdmd/AssetManager.cpp
+++ b/standalone/inc/flexdmd/AssetManager.cpp
@@ -252,7 +252,7 @@ void* AssetManager::Open(AssetSrc* pSrc)
    }
 
    if (!pAsset) {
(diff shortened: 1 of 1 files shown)


jsm174:
I don't think we can assume the error is always `SDL_GetError()`. (It gets to that message from a few different paths)

Added an unsupported extension message, in https://github.com/vpinball/vpinball/commit/a189d5b18975023c5d9461439317a2bb18e60ca2#diff-c65c9ff4c59cba7bf244832b703e0c561c2699ee71bdebcec8dcefadba39aedfR130

PR #1664, opened 2024-04-17
feat(vbscript): create object error message

francisdb:
To improve debugging scripts on standalone

image

### DIFF
--- a/src/utils/def.cpp
+++ b/src/utils/def.cpp
@@ -677,6 +677,8 @@ HRESULT external_create_object(const WCHAR* progid, IClassFactory* cf, IUnknown*
 
    const char* const szT = MakeChar(progid);
    PLOGI.printf("progid=%s, hres=0x%08x", szT, hres);
+   if (hres == E_NOTIMPL)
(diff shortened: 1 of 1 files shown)


jsm174:
this is not cross platform. Even if it is working, it shouldn't. 

```
%S (uppercase) is used in Microsoft's C with the Windows API to print wchar_t* (wide characters). This usage is not standard in other C environments, where %S can have different meanings or might not be supported at all.
```

Please just refactor

https://github.com/vpinball/vpinball/blob/9b49a0582b03af8a84e6086cf500d312366acfb2/def.cpp#L655

Look at the `hres` and then `PLOGE` vs `PLOGI`

francisdb:
image

Do you think this should be logged on error?

jsm174:
> Do you think this should be logged on error?

Not sure I follow completely, but we can't really exit immediately here. Returning no object is a valid thing. ie, in windows a user might not have a COM object installed and the table could be smart enough to handle it.

jsm174:
oh warning vs error. sorry, now it makes sense.. warning is fine.

PR #1668, opened 2024-04-18
Remove non-ascii characters from WPC.vbs

### LINKED ISSUE #1666: Unable to open WPC.VBS ERR
here has been a problem where the following error appears randomly on a table that loads WPC.vbs.

_**"Unable to open WPC.VBS. Ensure the it is in the Scripts folder of Visual Pinball."**_

I don't know the reason, but I knew the solution and took care of it personally.
The cause is that inappropriate characters (garbled) are used in the comment on line 5 of the WPC.vbs script.
Please delete this comment line.

------
'Last Updated in VBS v3.56

'sw116 and sw118 are commonly used for UL and UR flippers(in most WPC pins),
'while in IJ (i think the only one that has it different?!) these are the ramp switches.
**_'Everytime you press the flipper keys those switches are 登n・ so one did use the cSingleLFlip variable to prevent that._**
'Nowadays cSingleLFlip has been replaced by the separate keyStagedFlipperL.

francisdb:
fixes #1666

These characters are not `UTF8`, `ASCII` compatible and seem to be coming from `Windows-1252`

This should also be applied to 10.8.1

### DIFF
--- a/scripts/WPC.vbs
+++ b/scripts/WPC.vbs
@@ -2,7 +2,7 @@
 
 'sw116 and sw118 are commonly used for UL and UR flippers(in most WPC pins),
 'while in IJ (i think the only one that has it different?!) these are the ramp switches.
-'Everytime you press the flipper keys those switches are “on”, so one did use the cSingleLFlip variable to prevent that.
+'Everytime you press the flipper keys those switches are "on", so one did use the cSingleLFlip variable to prevent that.
 'Nowadays cSingleLFlip has been replaced by the separate keyStagedFlipperL.
 
 Option Explicit


Panasony1008:
Is this non-ASCII character encoding issue not fixed in 10.8 master?
The following also contains inappropriate characters that may cause problems.

**[NudgePlugIn_blur2NoAccel.vbs 104th line, 75th character (75th byte)]
[NudgePlugIn_blurNoAccel.vbs 92nd line, 75th character (75th byte)]
[NudgePlugIn_mjrAccelAndTilt.vbs 104th line, 75th character (75th byte)]
[NudgePlugIn_robBlurNoAccel.vbs line 94, character 75 (75th byte)]
[NudgePlugIn_robNoAccel.vbs 94th line, 75th character (75th byte)]**

francisdb:
Those are correct utf8. Does windows not support that?

Panasony1008:
I don't know if it's a problem for Windows.
If you open it with a text editor, the characters will be garbled.
I thought it would be better to remove the anxiety factor.

sample : if IsEmpty(mForce) then exit sub 'prevent errors if vpmNudge.TiltObj isn**窶冲** set

francisdb:
You should use an editor that supports utf8. Or configure your editor for utf8.That's the most common encoding these days.

vbousquet:
I think the way VPX handle file is to load them, check if they are UTF8 compliant, if not try an ISO 8859-1 to UTF8 conversion, then uses the file. So , I would guess that UTF8 should be ok (on any platform)

Panasony1008:
thank you.
I understood it well.
No problem.

PR #1683, opened 2024-04-24
fix(input): only touch motion for motion event

francisdb:
6d2ea06e96ed3362a8036f1031376dafc62fb598 was modifying keyboard events causing shift / ctrl keys to no longer workon mac/linux


francisdb:
@vbousquet @jsm174 can you please have a look as keyboard input is currently totally broken.

vbousquet:
Cleanly fixed in latest

PR #1684, opened 2024-04-24
refactored plog so it can be initialized from multiple places

YellowLabrador:
Simply moved the plog initialization out of main.cpp and into utils/logger.cpp. I adjusted all references and CMake build files to use it.

This is necessary so the web server can initialize plog when running in the Android app. The Android app will start/stop the web server before any table is launched. The independent web server control in the Android app is in a pending PR waiting for this to be integrated.

### DIFF
--- a/build/CMakeLists_bgfx-win-x64.txt
+++ b/build/CMakeLists_bgfx-win-x64.txt
@@ -417,6 +417,8 @@ add_executable(vpinball WIN32
    src/utils/BlackBox.h
    src/utils/CrashHandler.cpp
    src/utils/CrashHandler.h
+   src/utils/logger.cpp
+   src/utils/logger.h
 
    src/ui/vpinball.rc
    vpinball.idl
--- a/build/CMakeLists_bgfx-win-x86.txt
+++ b/build/CMakeLists_bgfx-win-x86.txt
@@ -418,6 +418,8 @@ add_executable(vpinball WIN32
    src/utils/BlackBox.h
    src/utils/CrashHandler.cpp
    src/utils/CrashHandler.h
+   src/utils/logger.cpp
+   src/utils/logger.h
 
    src/ui/vpinball.rc
    vpinball.idl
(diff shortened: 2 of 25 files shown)


jsm174:
@YellowLabrador, thank you for this! This looks good to me. @vbousquet, @toxieainc - okay to merge?

jsm174:
@YellowLabrador, I had to revert this as it broke all the windows builds. Could you check this out please: https://github.com/vpinball/vpinball/actions/runs/8856875291

YellowLabrador:
@jsm174  I don't know much about Windows. I looked at the logs, the error message is quite generic:

```
d:\a\vpinball\vpinball\src\core\def.h(636): error C2039: 'tolower': is not a member of 'std' 
[D:\a\vpinball\vpinball\build\vpinball.vcxproj]
  C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.16.27023\include\sstream(14): note: see declaration of 'std'
D:\a\vpinball\vpinball\third-party\include\plog/WinApi.h(17): error C2632: 'int' followed by 'int' is illegal 
[D:\a\vpinball\vpinball\build\vpinball.vcxproj]
```

I did add two new files `src/utils/logger.cpp` and `src/utils/logger.cpp`. I see the errors are in the visual studio project. Could it be that I need to add those files to the VS project as well? I only added them to the cmake files.

jsm174:
Okay, I can try to look into windows some more. My guess is its probably the order things got pulled in.

YellowLabrador:
The other cpp files in `utils/` all `#include "core/stdafx.h"`. Googling stdafx.h seems to be Microsoft compiler thing. Could it be that it's required to build on Windows?

YellowLabrador:
@jsm174 I just pushed an update with the change I mentioned above. Will it trigger the build on all platforms?

francisdb:
I also find it annoying that there are no builds on pr's. Most open source projects have this and there's the option to let maintainers validate the pr before the build in case of fear of abuse.

@YellowLabrador you can set github actions on your fork as a workaround

YellowLabrador:
@francisdb it actually triggered a build. It took a while for GitHub to publish the reports so I incorrectly assumed it wasn't running a build.

PR #1685, opened 2024-04-26
Play a table if it's the only argument passed to the executable

WildCoder:
If the only parameter passed to the executable is the path of a vpx table we default to playing the table. This allows to assign .vpx extensions to VPinballX and launch a table in desktop mode simply by clicking on its icon.

### DIFF
--- a/src/core/main.cpp
+++ b/src/core/main.cpp
@@ -521,7 +521,16 @@ public:
          }
 
          //
-
+#ifdef __STANDALONE__
+         // If the only parameter passed is a vpx table we play it automatically.
+         const bool launchfile = (!valid_param) && (nArgs == 2) && (i==1) && strstr(szArglist[i], ".vpx") == (&szArglist[i][strlen(szArglist[i]) - 4]);
+         if(launchfile)
+         {
+            valid_param = true;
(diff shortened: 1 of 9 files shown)


toxieainc:
@jsm174 Okay with you?

jsm174:
If you're good with it, I am too!

jsm174:
@WildCoder - Actually I didn't realize this PR modified wine source code. 😥  I've tried to keep wine sources to an absolute minimum or at least with added __STANDALONE__, so I knew exactly what things changed, ie: https://github.com/vpinball/vpinball/blob/standalone/standalone/scripts/winepatch. It makes a ton easier when upgrading.

WildCoder:
Ah yes, my bad. I created two branches hoping to make 2 PRs but my two commits got mixed into the same PR and I didn't know how to undo it. I can easily go back and surround my includes with #ifdef _STANDALONE_ if you want.

francisdb:
Looks like this only works for absolute paths, starting with `/`. A path starting with `./` or `~/` will not be picked up.

WildCoder:
oh interesting I hadn't noticed, only using it to launch the table when dragging the vpx over a desktop file.

PR #1691, opened 2024-04-28
Fix window creation

Nicals:
I had the refresh rate configured as 60Hz in my VPinball.ini file. But I found the following log line when launching tables:

> [VPX::Window::Window@77] Requested fullscreen mode 3840x2160 at 32Hz is not available. Switching to windowed mode

This MR fixes two issues:

  + VideoMode was listed for the wrong display (67122625dcefd61923f9a1b851b033ea09985a8e)
  + Window constructor arguments were mixed on call (614b95917896d8474520a190dcd55577a6c04241)

### DIFF
--- a/src/core/player.cpp
+++ b/src/core/player.cpp
@@ -309,7 +309,7 @@ Player::Player(PinTable *const editor_table, PinTable *const live_table, const i
    #endif
 #endif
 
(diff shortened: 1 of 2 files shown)


vbousquet:
Thanks !

PR #1700, opened 2024-05-02
Standalone joystick

GruntMeister:
I have a joystick controller, where my buttons are attached to. It took me a long time to figure out, that the default linux builds are only working with game controllers. I was wondering why this has to be a compile time decision.

For my setup now, I just enabled both, and it seems to work fine.


jsm174:
hello. thanks.

a few things

- Please submit PRs to the 10.8.1 branch. If accepted, I can then cherry-pick them into standalone.
- GameController code was all experimental as I added it to support Xbox controls on iOS and Android. Thats why it was put behind a compiler flag
- we did add support for both in this commit: https://github.com/vpinball/vpinball/commit/8eac333aededce4baaeb39214380e17f071eccdf
- however we then reverted it. Please see: https://github.com/vpinball/vpinball/pull/1471

GruntMeister:
Hi Jason,

Thanks for the explanation. I was not sure about how to contribute to
standalone. 😉

But I have two questions left:
- did you really mean the 10.8.1 branch - and not master?
- how to solve the simultaneous joystick and gamepad issue? I think I
should try to auto detect what's connected, if that makes sense. Ok?

Cheers, Ingo

Jason Millard ***@***.***> schrieb am Fr., 3. Mai 2024, 01:36:

jsm174:
> did you really mean the 10.8.1 branch - and not master?

The master branch has no standalone code, so yes, please submit PRs to the 10.8.1 branch.

> how to solve the simultaneous joystick and gamepad issue? I think I
should try to auto detect what's connected, if that makes sense. Ok?

why is the `gamecontrollerdb.txt` route not acceptable?

GruntMeister:
Hi Jason,
I think joysticks and game controllers are two different things in SDL. I
have no game controller, but a joystick controller. So I have no valid
gamecontrollerdb.txt mapping for it.
The expectation is to have a simple Button 1, 2, 3, 4, 5, ... mapping.
That's it.

At first I didn't care anyway, and simply tried to find the correct
mappings for my buttons. But I was only able to get 4 or 5 of my 12 Buttons
mapped (with really weired numbers by the way ;) ).

So, after I read a bit about the differences of joysticks and game
controllers in SDL I really think, that the correct solution will be:

- remove the ENABLE_SDL_GAMECONTROLLER macro (similar to my PR)
- call SDL_InitSubSystem with "JOYSTICK" and "GAMECONTROLLER" (similar to
my PR)
- in SDL_[CONTROLLER/JOYSTICK]BUTTON[UP/DOWN] only do s.th. when a
controller or joystick was detected

The rest should stay and behave the same.
I can only imagine, that the last point on my list was missing in the old
implementation. Because sadly I also get SDL_CONTROLLERBUTTONDOWN events,
even when I have a joystick connected. So I can imagine, that it was also
the other way around.

Am Di., 7. Mai 2024 um 17:10 Uhr schrieb Jason Millard <
***@***.***>:

jsm174:
> - remove the ENABLE_SDL_GAMECONTROLLER macro (similar to my PR)

Thanks! 

I apologize but I'm still not following. Any joystick can be mapped to a game controller. That's how the `SDL_GAMECONTROLLER` subsystem works. 

If it didn't work this way then I would agree we would need to support both simultaneously. Or at least be one or the other based on a `VPinballX.ini` entry.
 
What tool were you using to make the mapping?

Maybe @toxieainc, @vbousquet might have some ideas.

toxieainc:
@GruntMeister As a lot of reconstruction happened since this PR, we are now also on SDL3.
Could you please revisit? Thanks!

GruntMeister:
@toxieainc Thanks for the hint. It is really working now! I don't know which of the changes did the trick, but all joystick buttons can now be mapped.

PR #1711, opened 2024-05-06
Update main.mm

kaicherry:
Adds NSOpenPanel to allow user to load a .vpx table* *Tables opened this way open with -DisableTrueFullscreen flag


jsm174:
Cool. 

Can you please close 1712 and add that to this PR, so it's one cohesive change? 

Also the formatting appears to be all over the place, and some commented out code.

PR #1713, opened 2024-05-07
Adds NSOpenPanel when app is double-clicked

kaicherry:
Simply adds a standard macOS "Open File:" panel when the app is launched by double-clicking the app's icon. Tables open this way, versus the command line, or double-clicking a .vpx file, open full screen, per the current default behavior(s).

### DIFF
--- a/standalone/macos-arm64/main.mm
+++ b/standalone/macos-arm64/main.mm
@@ -1,57 +1,48 @@
 #import <Cocoa/Cocoa.h>
-#import <Foundation/Foundation.h>
 
 extern "C" char **g_argv;
 extern "C" int g_argc;
 extern "C" int WinMain(void*, void*, void*, int);
 
 @interface AppDelegate : NSObject <NSApplicationDelegate>
+- (void)makeWindowExit;
 @end
 
 @implementation AppDelegate
 
 - (void)applicationDidFinishLaunching:(NSNotification *)notification
 {
-   if (g_argc == 1) {
-      NSAlert* alert = [[NSAlert alloc] init];
-      [alert setMessageText:@"VPinballX_GL\n\nThis app must be run using command line arguments or double clicking a \".vpx\" file."];
-      [alert addButtonWithTitle:@"OK"];
-      [alert setAlertStyle:NSAlertStyleWarning];
-      [alert runModal];
-   }
+    if (g_argc == 1) {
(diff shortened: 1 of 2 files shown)


jsm174:
was thinking, do we really want  `-DisableTrueFullscreen` when opening this way?

kaicherry:
From a (current) "FTUX"...I think, yes. The current experience now is, double-click a .vpx, be greeted with a "black screen" in TrueFullscreen, or open the app, and be told to use the Terminal(.app) to do so...which is not the "generally expected behavior" when launching an app. An "accessory view with a a checkbox" certainly *could* be added...however, this is an improvement as it stands over what is current.

francisdb:
> be greeted with a "black screen" in TrueFullscreen

You mean only for a few sec while the table is loading?

kaicherry:
I do not, no. M1 macs, multiple tables, different display setups, same result: double-click a vpx, launches, black display (you'll hear audio thou) unto you quit, which briefly displays what was heard.

francisdb:
Anything specific in your config?
Do you have multiple windows set up like backglass, dmd and pup windows?

kaicherry:
I have all of those things. I am unsure how or why those things matter as the PR is to offer a Standard Open Panel when the app is launched without a document.

francisdb:
Just sounds like a bug if you get a black screen when double clicking a table. And this pr is a workaround for that bug? Anyway, that conversation could indeed be moved to a different issue.

kaicherry:
It's not a workaround..it's a standard macOS behavior when a bundled app is launched without a document to...offer an opportunity to choose one to open. 

I'm actually a bit...confused by this level of "pushback"...over a standard os thing.

francisdb:
I'm m not pushing back, just wanted to understand what was going on with the black screen. Also I don't have commit rights so it's up to @jsm174

kaicherry:
Oh, ok misread; it happens. The "black screen issue" is unrelated to this; it was part of the "why that flag" option; if can launch fullscreen for all I care...the issue (and it's one I had with a 'real user'...who is also a software engineer) in that the behavior on macOS is non-standard (once it moved into an app bundle making it an "app" vs a pure command line invocation) so I fixed it because it's...standard :)

The source/reason* for the black screen issue is understood by me, and is unrelated to this.

francisdb:
?

PR #1714, opened 2024-05-09
Implements macOS NSOpenPanel 

kaicherry:
Displays a standard macOS Open Panel to allow a user to choose a vpx table to run if the app is started via the Finder/double-clicking app icon.

If a table is selected, it is launched with the "-EnableTrueFullscreen" flag.

Additionally, should the user cancel, an Alert is shown describing the different app launch methods.

### DIFF
--- a/standalone/macos-arm64/main.mm
+++ b/standalone/macos-arm64/main.mm
@@ -1,57 +1,99 @@
 #import <Cocoa/Cocoa.h>
-#import <Foundation/Foundation.h>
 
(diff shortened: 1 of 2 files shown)


jsm174:
I don't think we want:

```
            new_argv[1] = strdup("-EnableTrueFullscreen");
```       

This way we can still respect the ini file if a person changes it. And like you said, later add a real gui in the mac app with a checkbox, checkmark, etc.

PR #1727, opened 2024-05-18
Add 'Sync' button to Layers dialog to allow faster finding

stevejones72:
When toggled on, the Sync keeps the selected element in the main screen highlighted in the layer screen to allow quicker usage within the dialog.


toxieainc:
Thanks, will merge to 10.8.1 then!

PR #1775, opened 2024-06-12
pup: log warnings on duplicate playlist / screen

francisdb:
This happens when the old way of setting up playlists and screens is mixed with the screens.pup and playlists.pup.


francisdb:
@jsm174 what am I doing wrong here?

image

francisdb:
It's going to be a mess once you bring back the log level for all the normal logs as you'll have to not touch these.

PR #1780, opened 2024-06-15
Move SDL event loop to main game loop

vbousquet:


francisdb:
Any reason to not have builds on pr's? That would avoid merging broken code.

vbousquet:
This is no professional code (by far, just have a look at the codebase, it's more like 20+ years of hacks piled together). We do it on our free time to enjoy. So we agreed to have relaxed contribution rules. When things break, we just fix and the ship keeps sailing.

Your point is right in theory and we could have test driven, a sonarqube quality check, and so on, but I think if we do that, the remaining contributor would just get bored, leave and the project would just be dead by now.

toxieainc:
Knowing professional code bases, its also actually not bad.  ;)

I also hope to be more active again in the future (don't hold your breath though :/), and so i fully support Vinces statements above!

francisdb:
Almost all open source projects have builds on pr's. What does this have to do with enterprise software development? If you want to have more ppl contributing it helps to have builds and for example code formatting rules as that's less uncertainty for somebody jumping in.

francisdb:
I'm not saying that pr's should have a hard rule on build passing. It's just nice to have a build. Certainly for cross platform projects.

jsm174:
Not to pile on, but I feel the same way as @toxieainc and @vbousquet. Last night, I put several commits in and probably broke the build like 8 times. Yes, I felt bad about it, but it happens.

The CI takes forever to run because of all the caches, builds, and sbc builds.  If I had to wait 2 - 4 hours for everything to be perfect...

Anyway, big features, like PuP dev, I usually work in my fork for a while and then do a PR. However things like tweaking the Android `ApplicationID` or switching the CI to `macos-14`, I don't want to feel like I'm at work.

francisdb:
I really don't get the pushback. There is zero cost involved and it's a convenience.

vbousquet:
@francisdb I think I misunderstood your first comment. There are indeed PR builds. They happen on the committer fork. So for this PR, it's here: https://github.com/vbousquet/vpinball/actions

The problem here is the way I used it: this little PR is a leftover of the big one from yesterday (rewrite of game loop and synchronization). I made it before going to bed, was tired but pushed it anyway to let Jason have a look to the consequence on the other backends than windows (we are on the opposite side of the world). This morning I did a quick PR with the left over. The right way of doing this would have been to delay the initial commit and not have this left over. Or at least, I should have waited for the leftover PR build on my branch to finish. But I made it quickly before doing other personal stuff (free time project, not a job, so not standing in front of computer). When I saw, the build fail, I just fixed it. Lots of the contributions happen like this.

Regarding having a better codebase:
- we have PR builds on the commiter fork, so I would not change that as adding more build means more delay and in the end, they are not used,
- we are trying to clean up things, but this is a cleanup task, not a rewrite, so it will never be perfect and will take years
- it would be nice to have tools to limit regressions (at least tests and static analysis) but I don't think anyone has the time and the energy to implement them

francisdb:
This was not meant to be a comment on this specific PR. Just in general. But it was indeed triggered because I tried to build the latest changes this morning.

> They happen on the committer fork.

For my part that's the same as saying we don't do pr builds.

Builds give the PR submitter faster feedback and the reviewer can be sure the code at least builds on all platforms.

Again all this does not keep ppl with commit access from doing quick and dirty PR's, they could as well just push commits to the branches directly. It's more for external ppl that want to contribute and for reviewing/validating those contributions.

The only reason to not enable PR builds would be a security concern. Eg release signing tokens being provided to pr builds. 

> Due to the dangers inherent to automatic processing of PRs, GitHub’s standard pull_request workflow trigger by default prevents write permissions and secrets access to the target repository. 

And to be extra sure you can [require approval for public forks](https://docs.github.com/en/organizations/managing-organization-settings/disabling-or-limiting-github-actions-for-your-organization#configuring-required-approval-for-workflows-from-public-forks), but then I understand that part is more work for the maintainers.
https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions

Issue #1779, opened 2024-06-15
Ability to resize the line number column in script editor to accommodate larger font size


gjupp:
**Is your feature request related to a problem? Please describe.**
Script/code editor line number column only shows 3 character width line numbers for some users depending on font choice/size resulting in visually incorrect line numbering past line 999. The quick and dirty solution is to reduce the font point size for default text in the code viewer options however this can result in harder to read text in the script itself depending on monitor size/resolution/eyesight.

**Describe the solution you'd like**
Line number column should auto adjust it's width to accommodate larger font size.

**Describe alternatives you've considered**
Ability for user to manually adjust the line number column width

**Additional context**
I would also class this as somewhat of a bug as well as feature request. There is a thread related to this issue at https://www.vpforums.org/index.php?showtopic=51839

stevejones72:
![image](https://github.com/vpinball/vpinball/assets/45272774/fe66cf45-8545-4bba-aea3-56389343a801)

How about a 4 char wide width as a default?

gjupp:
[...]

PR #1794, opened 2024-06-18
Make code line number margin dynamic based on size of script loaded

stevejones72:
Make margin for line number dynamic based on script size.

Tried with 4 and 3  length script sizes, seems to work. i.e. (000's and 0000's) lines of scripts.


stevejones72:
https://github.com/vpinball/vpinball/issues/1779

toxieainc:
Thanks! Will merge to 10.8.1, too.

PR #1798, opened 2024-06-20
docs: document MaxFramerate

### LINKED ISSUE #1731: 10.8.1: No longer allows framerate higher than playfield monitor framerate
Testing on standalone which does not have frame pacing.

Previously below settings allowed unbound frame rate:

```ini
; Sync the frame rate with the refresh rate of your monitor
; 0 - None: No synchronization.
; 1 - Vertical Sync: Synchronize on video sync which avoids video tearing, but has higher input latency.
; 2 - Adaptive Sync: Synchronize on video sync, except for late frames (below target FPS), also has higher input latency.
; 3 - Frame Pacing (default): Targets real time simulation with low input- and video-latency (also dynamically adjusts framerate).
SyncMode = 0
MaxFramerate =
```

But since b520142df525e9c36966e35a4eb32c5c6a60ad49 the max framerate is always capped to the main monitor framerate:

First unset `MaxFramerate` (`-1`) is now display refresh rate
https://github.com/vpinball/vpinball/blob/1bb3104ceccc146aae4166079226b2ff129e34f0/src/core/player.cpp#L298-L299

Even if `MaxFramerate` is set it is capped to display refresh rate
https://github.com/vpinball/vpinball/blob/1bb3104ceccc146aae4166079226b2ff129e34f0/src/core/player.cpp#L321

It's also unclear what `RefreshRate` is for?

francisdb:
closes #1731

### DIFF
--- a/src/assets/Default_VPinballX.ini
+++ b/src/assets/Default_VPinballX.ini
@@ -251,6 +251,10 @@ RefreshRate =
 ; 2 - Adaptive Sync: Synchronize on video sync, except for late frames (below target FPS), also has higher input latency.
 ; 3 - Frame Pacing (default): Targets real time simulation with low input- and video-latency (also dynamically adjusts framerate).
 SyncMode = 
+; Maximum frame rate. Will influence input latency unless frame pacing is configured.
+; Defaults to the playfield display refresh rate.
+; Use 0 for unbound frame rate.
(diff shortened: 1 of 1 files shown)


toxieainc:
@vbousquet Opinions on this one?

vbousquet:
Thanks

PR #1801, opened 2024-06-21
build(standalone): Asahi linux

francisdb:
This promotes the BGFX build as it runs great. GL compiles and starts but does not render my test tables properly on Asahi Linux. It might however still be interesting to have for other ARM linux environments.

### DIFF
--- a/standalone/README.md
+++ b/standalone/README.md
@@ -256,6 +256,20 @@ cmake -DCMAKE_BUILD_TYPE=Release -B build
 cmake --build build -- -j$(nproc)
 ```
 
+### Linux (Fedora Asahi Remix 41)
+In a terminal execute the following:
+```
+sudo dnf install @development-tools
+sudo dnf install gcc-c++ autoconf automake libtool cmake nasm bison curl systemd-devel libX11-devel mesa-libGL-devel libXext-devel zlib-ng-compat-static zlib-ng-compat-devel wayland-devel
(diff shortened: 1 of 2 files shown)


francisdb:
@jsm174 The gl cmake file has some specifics for rpi that would need to be put under these flags.

francisdb:
@jsm174 do we want to create separate files for `gles2`?

jsm174:
I would look at 10.8.1 versions (even the bgfx lib one) and see if we could pass in options into cmake.

francisdb:
> I would look at 10.8.1 versions (even the bgfx lib one) and see if we could pass in options into cmake.

Not sure I follow. This is a 10.8.1 PR. It's just that the aarch64 cmake assumes `gles` where this is not always the case.

francisdb:
needs to be rebased once #2094 is merged

jsm174:
> Not sure I follow. This is a 10.8.1 PR. It's just that the aarch64 cmake assumes `gles` where this is not always the case.

I meant, make the cmake configurable with options like we do for pi and rk3588

https://github.com/vpinball/vpinball/blob/0951a93bfe15187861f8217bab12bb06f7864de0/standalone/cmake/CMakeLists_gl-linux-aarch64.txt#L3-L4

https://github.com/vpinball/vpinball/blob/0951a93bfe15187861f8217bab12bb06f7864de0/standalone/cmake/CMakeLists_gl-linux-aarch64.txt#L52-L60

PR #1836, opened 2024-07-10
Fix light overflow on iOS

vbousquet:


jsm174:
Thanks for all the help with this!!

For reference:

Screenshot 2024-07-10 at 4 05 22 PM

notes2

Before:

IMG_3983

After:


image

PR #1850, opened 2024-07-19
Change VPBuildVersion to DOUBLE and return GIT_REVISION as decimal #1558

JockeJarre:
I have now reactivated VPBuildVersion (it was stated to be deprecated) but have changed it to return a double instead of integer.
The git build number is now the decimal of the returned value.

### DIFF
--- a/src/parts/pintable.cpp
+++ b/src/parts/pintable.cpp
@@ -1168,9 +1168,9 @@ STDMETHODIMP ScriptGlobalTable::get_Version(int *pVal)
 	return S_OK;
 }
 
-STDMETHODIMP ScriptGlobalTable::get_VPBuildVersion(DOUBLE *pVal)
+STDMETHODIMP ScriptGlobalTable::get_VPBuildVersion(double *pVal)
 {
-   *pVal = VP_VERSION_MAJOR * 1000 + VP_VERSION_MINOR * 100 + VP_VERSION_REV + GIT_REVISION / 10000.0f;
+   *pVal = VP_VERSION_MAJOR * 1000 + VP_VERSION_MINOR * 100 + VP_VERSION_REV + GIT_REVISION / 10000.0;
 	return S_OK;
 }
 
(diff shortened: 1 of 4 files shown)


toxieainc:
@jsm174 @vbousquet Any issues with this? Looks fine to me like this..

vbousquet (changes requested):
I don't see any problem with the proposed change.
To be more consistent with other uses of float & double in IDL and implementation, I think this should use `double` instead of `DOUBLE`
Also, when computing the value, this should use double value and not float

src/parts/pintable.cpp
```diff
@@ -1168,9 +1168,9 @@ STDMETHODIMP ScriptGlobalTable::get_Version(int *pVal)
 	return S_OK;
 }
 
-STDMETHODIMP ScriptGlobalTable::get_VPBuildVersion(int *pVal)
+STDMETHODIMP ScriptGlobalTable::get_VPBuildVersion(DOUBLE *pVal)
 {
-	*pVal = VP_VERSION_MAJOR * 1000 + VP_VERSION_MINOR * 100 + VP_VERSION_REV;
+   *pVal = VP_VERSION_MAJOR * 1000 + VP_VERSION_MINOR * 100 + VP_VERSION_REV + GIT_REVISION / 10000.0f;
```
vbousquet, line 1173: Use `10000.0` instead of `10000.0f`

jsm174:
I'm guessing I will need to regen the IDL for standalone, so I can either tack on when this is ready, or just add a new commit.

toxieainc:
Thanks all!

toxieainc:
@JockeJarre One more thing: I would propose to divide by 100000, maybe even 1000000, instead of 10000. Seems to be more future proof for now.

jsm174:
@JockeJarre / @toxieainc ,

Was just thinking about this:

`VPBuildVersion - same as Version but with the github build number as decimal (GIT_REVISION / 10000.0)`

I know for myself, when building manually during dev, I never have `GIT_REVISION` set. Guessing this shouldn't be a issue.

toxieainc:
Yes, i don't think so. Unless one would test a very specific table that checks for this.

JockeJarre:
> One more thing: I would propose to divide by 100000, maybe even 1000000, instead of 10000. Seems to be more future proof for now.

@toxieainc Wouldn't it be better if we make the counter start from 1 for a new release? Normally 9999 would for sure be enough? We also have statements like `fprintf(f, "%04X", GIT_REVISION);` in the code. So we do not expect it to exceed 9999.


> I know for myself, when building manually during dev, I never have `GIT_REVISION` set. Guessing this shouldn't be a issue.

@jsm174 we could set the "default" value in the source to 9999 when you compile locally. When testing tables, your local build will always be newer than those checked in.

toxieainc:
I'd hope 9999 would be enough, too, but we're already pretty high right now..  :)

JockeJarre:
Yes, you are right. 20% is used soon. But we need to make 10.8.1 start from 1, which @jsm174 already has prepared for. We need to update the hash number in the build script

JockeJarre:
> Yes, you are right. 20% is used soon. But we need to make 10.8.1 start from 1, which @jsm174 already has prepared for. We need to update the hash number in the build script

But with 10.8.1 it is already to late, it is using the same counter as 10.8.0... It could be solved by moving to 10.8.2 soon and reset the counter hash (VERSION_START_SHA)

PR #1871, opened 2024-07-30
Update Default_VPinballX.ini with comments

Chaos99:
Added some descriptive comment to nudge settings

Will continue with additional commenting later.


toxieainc:
Nice. Just give us a hint when you're done.

francisdb:
Was it not worth merging this?

PR #1874, opened 2024-08-01
Set GIT_REVISION default value to 9999 for local compile

JockeJarre:
As in my last pull request, we discussed what happens when compiling locally with the GIT_REVISION set to zero.
Therefore I have set it to 9999 in git_version.h and updated the git action to replace 9999 with the git revision.

I have also added a comment to update VERSION_START_SHA in .github\workflows\vpinball.yml on version changes


toxieainc:
Thanks!

PR #1876, opened 2024-08-01
Added input device management for Win32 builds

ScaryG:
This PR is to address the issue, mostly encountered by non-cabinet PC users, where having multiple input devices (controllers, joysticks, steering wheels, etc) means they are all used for input to Visual Pinball, often hindering gameplay.

A new "Manage devices" button was added to the Key config dialog:

VP_Input_PR_1

Clicking it will take the user to a new dialog listing the devices currently attached to the PC, where they can be toggled on and off:

Screenshot 2024-07-30 212010

Implementation
------------------

By default, all devices are considered ON and will give valid inputs, which is maintaining the previous behaviour.

Input device preferences are saved to a new Settings section, ControllerDevices. Input is rejected in HandleInputDI() only if there's a specific setting that disables each device.

The feature is currently only implemented for the Win32 platform, and the new dialog resources are only in the English rc file. I'm not sure if the others are used - if they are let me know and I can put them in there too.


jsm174:
Any chance this can be rebased, so we don't have 30 extra commits?

ScaryG:
> Any chance this can be rebased, so we don't have 30 extra commits?

Done, made a new PR.

PR #1898, opened 2024-08-19
Nudge/plunger input model enhancements

mjrgh:
This update implements some enhancements to the pin cab I/O controller device input model that came out of work I'm doing on a Pico-based controller (a sequel the Pinscape KL25Z system).  The idea is to address some of the long-standing technical problems in the HID input model we're using today, which is a legacy of the earliest commercial I/O controllers.  All of the devices in the years since have just carried that old input model forward over the years, but it has some inherent problems that can be fixed through some relatively simple changes.  This update implements the changes that I think are needed to solve those problems.

The theory behind this is described here: [Improving the Virtual Pin Cab Input Model](http://mjrnet.org/pinscape/OpenPinballDevice/NewPinCabInput.htm).

I've implemented the device side of this experimentally in my new Pico project, which I haven't published yet, and in the KL25Z Pinscape project, which I have.  Anyone who has a KL25Z pin cab can try out the changes by installing the new KL25Z 2024-08-18 release.  See [Pinscape Release Builds](http://mjrnet.org/pinscape/swversions.php).

My testing with the KL25Z implementation has been quite encouraging.  The new model does seem to be a noticeable improvement.  It substantially improves the consistency of the plunger launch behavior for me, and the nudge response seems considerably more natural to my eye.  I don't think there are any downsides to the extensions, since they're **extensions**; they're basically new features alongside the existing ones that won't interfere with or affect existing setups.

This update **also** includes a closely related but somewhat orthogonal new feature, which is support for my proposed [Open Pinball Device HID](http://mjrnet.org/pinscape/OpenPinballDevice/OpenPinballDeviceHID.htm).  The idea behind this is to eventually replace the traditional joystick HID interface with a HID that's explicitly designated in its usage codes as a Pinball Device.  I recently discovered that the official USB HID Usage Tables actually set aside a code for a "Pinball Device" Application Collection (that's the USB term of art for a software model representing a physical box on your desktop, the same category used for keyboards, mice, and joysticks).  I'm pretty sure I'm the only person on Earth who knows it's there, because I don't think many people browse through those tables on their lunch hour, and I don't get the impression that the USB committee member who added the entry gave it more than about five minutes of thought at the time, so I fully expect they've forgotten about it by now.  At any rate, seeing it made me realize that we're doing it wrong by co-opting the HID Joystick CA for all of our devices.  The wrongness is *mostly* an academic matter of how USB was intended to be used at a conceptual level, but not entirely; I know from hearing from Pinscape users that the joystick model actually does create practical problems for some people, due to compatibility issues with non-pinball video games that believe that a HID device calling itself a joystick *really is* a joystick.  These sorts of issues can be solved by using HID as intended, by designating our devices under the functionally correct Pinball Device usage code.  This also has a few other potential advantages, particularly in terms of automatic configuration.  The rationale and technical details are all covered in depth in the linked proposal, so I won't go into further detail here.  The implementation is quite self-contained and has almost no impact on any of the existing input processing code, and it doesn't replace the joystick or any other existing input method, just adds a new one alongside them.

### DIFF
--- a/dialogs/KeysConfigDialog.cpp
+++ b/dialogs/KeysConfigDialog.cpp
@@ -317,6 +317,7 @@ void KeysConfigDialog::AddStringAxis(const string &name, const int idc, const in
    ::SendMessage(hwnd, CB_ADDSTRING, 0, (LPARAM)"rZ Axis");
    ::SendMessage(hwnd, CB_ADDSTRING, 0, (LPARAM)"Slider 1");
    ::SendMessage(hwnd, CB_ADDSTRING, 0, (LPARAM)"Slider 2");
+   ::SendMessage(hwnd, CB_ADDSTRING, 0, (LPARAM)"OpenPinDev");
    ::SendMessage(hwnd, CB_SETCURSEL, selected, 0);
    ::SendMessage(hwnd, WM_SETREDRAW, TRUE, 0);
 }
--- a/dialogs/KeysConfigDialog.cpp
+++ b/dialogs/KeysConfigDialog.cpp
@@ -353,6 +353,9 @@ BOOL KeysConfigDialog::OnInitDialog()
     int key = g_pvp->m_settings.LoadValueWithDefault(Settings::Player, "PBWRotationValue"s, 0);
     SetDlgItemInt( IDC_GLOBALROTATION, key, FALSE);
 
+    on = g_pvp->m_settings.LoadValueWithDefault(Settings::Player, "AccelVelocityInput"s, false);
+    SendDlgItemMessage(IDC_CBGLOBALACCVEL, BM_SETCHECK, on ? BST_CHECKED : BST_UNCHECKED, 0);
+
     on = g_pvp->m_settings.LoadValueWithDefault(Settings::Player, "TiltSensCB"s, false);
     SendDlgItemMessage(IDC_CBGLOBALTILT, BM_SETCHECK, on ? BST_CHECKED : BST_UNCHECKED, 0);
 
@@ -391,6 +394,9 @@ BOOL KeysConfigDialog::OnInitDialog()
 
     key = g_pvp->m_settings.LoadValueWithDefault(Settings::Player, "PBWAccelMaxY"s, 100);
     SetDlgItemInt( IDC_YMAX_EDIT, key, FALSE);
+
+    key = g_pvp->m_settings.LoadValueWithDefault(Settings::Player, "PlungerSpeedScale"s, 100);
+    SetDlgItemInt(IDC_PLUNGERSPEEDSCALE, key, FALSE);
 
     on = g_pvp->m_settings.LoadValueWithDefault(Settings::Player, "EnableMouseInPlayer"s, true);
     SendDlgItemMessage(IDC_ENABLE_MOUSE_PLAYER, BM_SETCHECK, on ? BST_CHECKED : BST_UNCHECKED, 0);
@@ -488,6 +494,7 @@ BOOL KeysConfigDialog::OnInitDialog()
     AddStringAxis("PlungerAxis"s, IDC_PLUNGERAXIS, 3); // assume Z Axis as standard
     AddStringAxis("LRAxis"s, IDC_LRAXISCOMBO, 1); // assume X Axis as standard
     AddStringAxis("UDAxis"s, IDC_UDAXISCOMBO, 2); // assume Y Axis as standard
+    AddStringAxis("PlungerSpeedAxis"s, IDC_PLUNGERSPEEDAXIS, 0);  // not assigned by default
 
     for (unsigned int i = 0; i < eCKeys; ++i)
       if (regkey_idc[i] != -1 && GetDlgItem(regkey_idc[i]) && GetDlgItem(regkey_idc[i]).IsWindow())
@@ -714,6 +721,7 @@ void KeysConfigDialog::OnOK()
     SetValue(IDC_JOYDEBUGGERCOMBO, Settings::Player, "JoyDebuggerKey"s);
     SetValue(IDC_JOYLOCKBARCOMBO, Settings::Player, "JoyLockbarKey"s);
     SetValue(IDC_PLUNGERAXIS, Settings::Player, "PlungerAxis"s);
+    SetValue(IDC_PLUNGERSPEEDAXIS, Settings::Player, "PlungerSpeedAxis"s);
     SetValue(IDC_LRAXISCOMBO, Settings::Player, "LRAxis"s);
     SetValue(IDC_UDAXISCOMBO, Settings::Player, "UDAxis"s);
     SetValue(IDC_JOYPAUSECOMBO, Settings::Player, "JoyPauseKey"s);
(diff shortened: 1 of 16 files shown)


toxieainc:
Looks good to me, thanks!
There is one additional caveat though, the main development already happens in the 10.8.1 branch, so could you please also port this there?

mjrgh:
> Looks good to me, thanks! There is one additional caveat though, the main development already happens in the 10.8.1 branch, so could you please also port this there?

Will do - I'll open a new PR for that.

jsm174:
@mjrgh, @toxieainc 

I just rebased standalone and will have to look into converting this to use libhidapi. (I know it works as I can talk to the pinscape with libdof using it)

Anyway, if you get compile errors on 10.8.1 when you copy this in, you'll have to `#ifndef __STANDALONE__` it. Here's a patch file just to show roughly where I had to work around it:

[pininput.txt](https://github.com/user-attachments/files/16738377/pininput.txt)

KrakenByteKraken:
Thank You!!! I'm sure this will bring a better pinball simulation and also fix many issues, like improve compatibility between pinball and other games.

toxieainc:
@mjrgh I just also ported this to the 10.7.2 branch, could you please verify that it works as expected? Thanks!

mjrgh:
@jsm174 
>I just rebased standalone and will have to look into converting this to use libhidapi. (I know it works as I can talk to the pinscape with libdof using it)

Is libhidapi something VP 10.8.1 is linking portably, or is that only platform-specific?  If it's universal, I can try to make the base 10.8.1 OpenPinDev implementation use libhidapi instead of the native Win32 USB APIs so that it doesn't need any platform-specific code.

mjrgh:
Great - let me know what I need to do to update my build environment for it once it's there.  I expect the API calls should translate pretty straightforwardly, since it doesn't need anything fancy, just basic HID discovery and then reading its input reports.

jsm174:
sure thing!

fwiw, here is your PinscapeDevice.cs ported to use it: 

https://github.com/jsm174/libdof/blob/master/src/cab/out/ps/PinscapeDevice.cpp

jsm174:
@mjrgh - so I have the hdiapi lib and dlll being built for 32 bit and 64 bit. It uses vpx 64-bit dll naming conventions (ie, hdiapi64.dll)

You can download them via the artifacts from the libdof CI:

https://github.com/jsm174/libdof/actions/runs/10802794992

You just need the hdiapi.h header from here: 

https://github.com/libusb/hidapi/blob/d0732cda906ad07b7e1ef93f1919035643620435/hidapi/hidapi.h

mjrgh:
@jsm174 - great, thanks, I'll give it a try.

mjrgh:
@jsm174 - do you have any proposal for where hidapi.h should go in the source tree?  I see there's already a copy called SDL_hidapi.h under third-party/include/SDL; perhaps there should be a single shared copy under third-party/include/hidapi?  Unless the SDL copy is a forked version with private mods or something.

mjrgh:
@jsm174 - also, I assume this is going to need some additions to the build scripts to pick up the new .lib in the linker commands and .dll in the deployment file copy?

mjrgh:
@jsm174 - well, crap, hidapi turns out to be a little too lightweight for our purposes.  It doesn't seem to have any access to the report descriptor data, which unfortunately is critical for this.  I need three things it doesn't seem to provide any access to - input report size, input report size, and the internals of the input report descriptor.  I'm not sure what the best way to proceed is.  Their enumerator struct looks like it wraps a native OS device path, so maybe the thing to do is use native Win32 API calls to get the extra information I need, leaving that as something that will have to be ported to corresponding native APIs on other platforms.  At least the rest of the code will be fully portable, so it's less porting work than if the whole thing were Win32 calls.

(I should warn you that I think this is going to throw a wrench into your libdof Pinscape driver.  The latest Pinscape firmware uses a larger report size than the previous ones, so the fixed report size being read is probably going to break.  It does on the Windows read call, at least; maybe they can handle it under the covers and it'll be okay.  With the Windows API, though, you have to ask for the exact report size the device sends, or the read fails.  I suspect that's what's going to happen with hidapi, since it looks like it's a pretty thin wrapper on the system APIs.  I have no idea how to fix that without the descriptor information.)

jsm174:
@mjrgh  - My knowledge on USB and HID is very limited.

So this function isn't helpful for anything?

```
	int HID_API_EXPORT_CALL hid_get_report_descriptor(hid_device *dev, unsigned char *buf, size_t buf_size);
```

mjrgh:
That should do the trick .  The version of the doc I was looking at pre-dated that so I missed it.  I see that it's in the header you pointed me to, so I'll give it a try.

mjrgh:
General question: the hidapi library has a shutdown function that they recommend calling before program exit.  Does anyone have an opinion about where that should go, or if we should bother calling it at all?  I notice that STANDALONE #ifdef's out some similar shutdown functions with comments mentioning crashes they cause, so I want to avoid creating a new one of those special cases. The library is only invoked during a runtime session, so it probably would be ideal to shut down hidapi after every run finishes, but that sounds like exactly the sort of thing that's creating the STANDALONE problems for other subsystems.

toxieainc:
> @mjrgh I just also ported this to the 10.7.2 branch, could you please verify that it works as expected? Thanks!

@mjrgh Could you please verify that one at some point?

mjrgh:
> > @mjrgh I just also ported this to the 10.7.2 branch, could you please verify that it works as expected? Thanks!
> 
> @mjrgh Could you please verify that one at some point?

Sorry for the delay on that - I'll take a look at that next.

mjrgh:
> > @mjrgh I just also ported this to the 10.7.2 branch, could you please verify that it works as expected? Thanks!
> 
> @mjrgh Could you please verify that one at some point?

@toxieainc Everything looks good.  I took a look at the code, built it, did a little testing on my cab, everything seems to be working well.  Thanks for taking care of that branch!  So the active branches at the moment are 10.8.1, 10.7.2, 9.9.5... main also, or is that dormant for now?

jsm174:
Main is 10.8.0
standalone is main + 1 giant commit , but I can take care of all that

toxieainc:
In principle 10.8.1 is the only active one. And then, if its worth it, we port not-so-risky things/fixes back to 10.8.0 and sometimes 10.7.2.
9.9.5 is extremely exceptional, but as you already had the work done, why not..

jsm174:
@mjrgh,

I noticed these <probably> useless warnings in that hidparser lib. Since you modified the source, maybe we can take care of these?

Screenshot 2024-09-15 at 2 26 57 PM

mjrgh:
@jsm174 - I just pushed a commit that should fix them (and tested that the fixes didn't break the parser, at least as far as my test device).  If you want to run them through your compiler to make sure it's satisfied now, the commit is here: https://github.com/mjrgh/vpinball/tree/hidparser-warning-fixes.  Let me know if they do the trick and I'll open a PR.

mjrgh:
> Main is 10.8.0 standalone is main + 1 giant commit , but I can take care of all that

It looks like 'master' doesn't have any of the ports - so that's all new in 10.8.1, or am I missing something?  If there aren't any other builds to update then it should be pretty quick for me to move the hidapi code over to it.

PR #1913, opened 2024-09-11
Nudge velocity, plunger speed, and Open Pinball Device for 10.8.1 branch

mjrgh:
This is the port of these features to 10.8.1.  I've done a little testing with it and all seems to be working.  I'm opening this as a draft for now because of the build impact (see below) and because I want to do a little more testing.

The Open Pinball Device implementation this time **mostly** uses the hidapi library, which in principle should make it portable to all platforms.  But hidapi is a little too limited for what this needs, so I had to break one sub-task into platform-specific code.  I created a portable abstraction for the task and implemented it for Win32, with a stub for all other platforms.  Other platforms can fill it in incrementally as desired, by adding an #elif branch at the bottom of PinInput.cpp and implementing the one function.  The abstract interface to it is thoroughly commented where it's defined, so hopefully it won't be too hard to implement on most platforms that have native HID support.

I had to rename the former "hid.cpp/h" files and the functions in them, because they created a naming collision with hidapi.  That's just as well, because naming them "hid_xxx" was kind of insane given that they were most definitely not "HID" code, other than they happened to use HID in the course of their work.  I renamed them "ushock_output_xxx" because that's what they really do.  This isn't exactly major surgery but it affects a few scattered lines of code elsewhere, so a few extra files got updated.

I need some guidance from someone who knows how the build setup works on how to properly add the hidapi library to the build configuration files.  I'm assuming they need to go in CMakeLists.txt, but it's not clear from inspection what the general formula for .lib/.dll combos is.  The others like it seem to have inconsistent treatment.  I don't want to break the build for everyone, so I'm marking this as Draft until I get some clarity on how to update the build.


src/core/pininput.cpp
```diff
@@ -2094,3 +2149,605 @@ LPDIRECTINPUTDEVICE PinInput::GetJoystick(int index)
...
+      }
+   }
+};
+
+// the USB implementation is in terms of the hidapi library
+#include "third-party/include/hidapi/hidapi.h"
+
+// NON-PORTABLE FUNCTION INTERFACE: Inspect a hidapi device's HID Input
+// Report Descriptor to determine if this is an Open Pinball Device.
+// hidapi doesn't provide access to the Report Descriptors, so we have
```
francisdb, line 2243: Is there an issue on the side of hidapi to track this missing feature? Might be a good idea to link it here.

mjrgh: I doubt it - I think it's just outside the scope of what hidapi sets out to do.

francisdb: There's a long conversation at https://github.com/libusb/hidapi/issues/249

jsm174:
The windows version of vpx has all the libraries directly in the repo. 

Standalone took a while to get where it is, so we made `external.sh` scripts to build all the libs. (The CI has logic to cache all the libraries since it takes a while.)

If we were to add all the standalone libraries, it would blow up the size of the repo and a fresh checkout would take forever. (It's already getting very slow)

Anyway back to your question:

`hidapi.lib` would go in `third-party/build-libs/windows-x86`
`hidapi.dll` would go in `third-party/runtime-libs/windows-x86`

`hidapi64.lib` would go in `third-party/build-libs/windows-x64`
`hidapi64.dll` would go in `third-party/runtime-libs/windows-x64`

`hidapi.h` would go in `third-party/include/hidapi/hidapi.h`

You would just need to add `hidapi` to `target_link_libraries` section in all the `CMakeLists_*.txt` files and you would have to add it to the `POST_BUILD` section.

Since the cmake files already have support for knowing about anything in `third-party/include` your source just needs to reference it at: 

`#include <hidapi/hidapi.h>`

As far as updating the Visual Studio project files, I'll defer to @toxieainc and @vbousquet as I only use cmake. 

As far as updating standalone, we really just need to update all the `external.sh` `LIBDOF_SHA` entries to `5c43c99ea28b44bb58b74554c4303a505e208148` and it will be automatically copied to where it needs to be.

Any other errors on the standalone side, I can help take care of.

src/core/pininput.cpp
mjrgh: It should be exactly what's needed, and it looks like the new facilities are in the library version that Jason's, so I'll look into switching over to that. If it works it should eliminate the extra platform-specific function.

mjrgh:
@jsm174 - perfect, thanks. I did put everything in those folder locations, but I wasn't clear on the CMake updates.  I'll add those couple of lines.

mjrgh:
@jsm174 - I see what you mean by the CMakeLists_*.txt files - I didn't realize there were a dozen of them.  How does the x64 library suffix work in the x64 files?  All of the other libraries in there just have the base name.  Is there a name munging rule that automatically adds the x64 suffix or something?

jsm174:
> @jsm174 - I see what you mean by the CMakeLists_*.txt files - I didn't realize there were a dozen of them. How does the x64 library suffix work in the x64 files? All of the other libraries in there just have the base name. Is there a name munging rule that automatically adds the x64 suffix or something?

It finds them because of the `target_link_directories`:

(x64)
https://github.com/vpinball/vpinball/blob/0cbc2817a5f881067195c54e311025092d998cd3/make/CMakeLists_bgfx-win-x64.txt#L597-L599

(x86)
https://github.com/vpinball/vpinball/blob/0cbc2817a5f881067195c54e311025092d998cd3/make/CMakeLists_bgfx-win-x86.txt#L598-L601

and then they get copied into the build folder manually:

(x64)
https://github.com/vpinball/vpinball/blob/0cbc2817a5f881067195c54e311025092d998cd3/make/CMakeLists_bgfx-win-x64.txt#L649-L653

(x86)
https://github.com/vpinball/vpinball/blob/0cbc2817a5f881067195c54e311025092d998cd3/make/CMakeLists_bgfx-win-x86.txt#L652-L656

mjrgh:
@jsm174 - so the DLLs are named namex64.dll, but it looks like all of the libraries are just name.lib.  I guess I need to rename hidapi64.lib to hidapi.lib in third-party/build-libs/windows-x64?

jsm174:
did you try `hidapi64.lib`? I really think it should work as we are doing that in libdof..

mjrgh:
> did you try hidapi64.lib? I really think it should work as we are doing that in libdof..

That's how it's currently named, and it builds fine that way with VS, but I haven't tried to generate any of this from cmake so I just assume it's all going to fail spectacularly when someone tries that.

jsm174:
lets leave it hidapi64.lib. 

https://github.com/jsm174/libdof/blob/5c43c99ea28b44bb58b74554c4303a505e208148/CMakeLists.txt#L137

I know it will work 😄

jsm174:
Also, the CI builds vpx in your repo, so you'll know when it starts working

https://github.com/mjrgh/vpinball/actions/runs/10818707752

jsm174:
any new files added or removed need to be updated in the list of cmake files as well:

https://github.com/mjrgh/vpinball/actions/runs/10818707752/job/30014951898#step:5:29

jsm174:
To get through that first macos error, somewhere in `hid-report-parser.h`, we could 

```
#ifdef __STANDALONE__
#ifdef BYTE_SIZE
#undef BYTE_SIZE
#endif
#endif
```

to get through the `_wtoi` maybe some helper in def.h/def.cpp?

```
inline int wtoi(const wchar_t *str)
{
  return (int)wcstol(str, 0, 10);
}
```


to get through `_countof` maybe:

`for (size_t i = 0; i < std::size(keyMap); ++i, ++m)`

jsm174:
cool.

I think on linux cmakes, we need `hidapi-hidraw` instead of `hidapi`.

https://github.com/jsm174/libdof/blob/master/CMakeLists.txt#L279

mjrgh:
> I think on linux cmakes, we need hidapi-hidraw instead of hidapi.

I'll give that a try.

mjrgh:
General question: the hidapi library has a shutdown function that they recommend calling before program exit. Does anyone have an opinion about where that should go, or if we should bother calling it at all? I notice that STANDALONE #ifdef's out some similar shutdown functions with comments mentioning crashes they cause, so I want to avoid creating a new one of those special cases. The library is only invoked during a runtime session, so it probably would be ideal to shut down hidapi after every run finishes, but that sounds like exactly the sort of thing that's creating the STANDALONE problems for other subsystems.

jsm174:
tbh, standalone never really had the concept of running multiple tables in the same process. You play a table, you exit, app closes.  (We do try to aim for no memory leaks and properly destroying objects. If things don't shut down right, for example, pinmame nvram might not save, etc.)

I've since had to revisit for the new iOS app as exiting the app programmatically really isn't allowed.

Is shutdown a one time thing, or can it be restarted? Ie, could just do where SDL init is at?

jsm174:
oh one other thing, I think for android and iOS we did have alot of this kind of stuff #ifdef out, and not compiled in via the cmake.  Android phones don't really give true usb access, and iOS I'd have to do research.

mjrgh:
> oh one other thing, I think for android and iOS we did have alot of this kind of stuff #ifdef out, and not compiled in via the cmake. Android phones don't really give true usb access, and iOS I'd have to do research.

Yeah, maybe I'll shuffle the OpenPinDev stuff into a new module and drop it from the mobile cmakefiles.

mjrgh:
Okay, great, looks like stubbing it on the mobiles did the trick.  It's all building now except for the two mobiles that look like they were already broken before ("I swear it was like that when I got here").

mjrgh:
Getting back to this...
> Is shutdown a one time thing, or can it be restarted? Ie, could just do where SDL init is at?

The documentation doesn't say, and it's hard to be sure when there are a dozen back-end implementations, but from the ones I surveyed it looks like they're careful to make it a mirror image of resource allocation on startup.  So yes, restarts should be okay, and it looks like they mean for that to be true.   So PinInput::UnInit() might be a good place for it, and it's definitely the efficient thing to do, as long as run mode is the only place it ever gets used.  My only hesitation is that this is a matter of statics inside hidapi, so if someone comes along later and wants to use hidapi in design mode, it would probably crash their code after exiting run mode.

jsm174:
> Okay, great, looks like stubbing it on the mobiles did the trick. It's all building now except for the two mobiles that look like they were already broken before ("I swear it was like that when I got here"). Still have some more details I want to work on but at least I think the build is sorted out, and no one will have to port Win32 HID calls, thank goodness.

I know whats up with the android bgfx builds. We should be good to try again this evening. I will keep you posted

jsm174:
@mjrgh - if you can rebase with the latest 10.8.1 branch, and then update:

BGFX_CMAKE_VERSION=1.128.8786-481
BGFX_PATCH_SHA=ee8994d080ac4a2acc9d94bfd33dac22e785ddd8

I believe that should fix the Android builds!

mjrgh:
@jsm174 - okay, did the rebase, where do the BGFX_CMAKE... changes go?

jsm174:
in the external.sh files. you can just do a search and replace.

mjrgh:
@jsm174 - hooray, that did the trick. Thanks!

toxieainc:
So is this ready to go then? Anything that would need to be ported back in addition to 9.9.5,10.7.4,10.8.0?

toxieainc:
Bonus points for reverting standalone/android-arm64-v8a/external.sh and the shader files..  :)

mjrgh:
> So is this ready to go then? 

I think it's ready as far as the initial functionality, but I was thinking I'd pursue the high-time-resolution flipper button idea a little further first.  That's still speculative, though, so maybe it would be better to call this done and start a new PR if the flipper stuff pans out.

> Anything that would need to be ported back in addition to 9.9.5,10.7.4,10.8.0?

The earlier PRs for 9.9.5 and 'master' had the same functionality, just without the portable hidapi implementation, so the only thing would be if anyone wanted the hidapi in those. 

> Bonus points for reverting standalone/android-arm64-v8a/external.sh and the shader files.. :)

Oh, yeah, I should get those back to their original state.  The external.sh was intentional, though - that was an update from jsm174 to get the android build working.

toxieainc:
Great! Thanks a lot..

vbousquet:
First, really thanks for all this. It's really great stuff!

I think your idea regarding high-time-resolution flipper button is really interesting. This is something I have been spending some time, so here are a few insights around the complete latency question I gathered along that eventually could help. First, when speaking of latency, I think there are 2 different perceived latency, one regarding gameplay (ability to do flipper tricks,...) and one visually perceived that I would define like this:
- gameplay latency = difference between expected behavior and simulated one if ones would have its eye closed,
- visual latency = the same but with the eyes opened :) and adds the rendering / framebuffer etc. latency to gameplay latency (won't go into this here)

Gameplay latency is difficult to measure but even to define since it implies multiple component. A classic sequence on a real pinball would be the following:
- move fingers, press button, mechanical contact, electric signal available to controller CPU (mostly instantaneous),
- Controller CPU strobe its switch matrix and update its internal copy in RAM,
- Controller CPU process the state change and toggle the flipper solenoid output,
- Solenoid coils current ramp up and flipper move (with varying power).
The switch contact to solenoid output change has been measured to be around 1ms on WPC fliptronnics.

For VPX, it becomes like this:
1. move fingers, press button, mechanical contact, electric signal available to pinball input device (mostly instantaneous),
2. pinball input device process, then communicates with computer through USB (with all that implies and you already described)
3. computer input API process and makes this data available to VPX
4. VPX process input and report it to script which updates a switch state table for PinMame
5. PinMame reads switch state table and emulate Controller CPU, in turns toggling a virtual solenoid output
6. VPX synchronize with PinMame and report solenoid state change to script which request physic engine to move flipper
7. VPX's physics engine actually move the flipper

This sequence happens with the following threading setup:
- All version of VPX excepted 10.8.1 BGFX variant are single threaded, mostly locked to display device. That is to say, most of the time, the VPX thread is locked waiting for the GPU to sync with display, leading to a latency pace directly linked to the display refresh rate.
- Until the TimeFence feature (which is recent and only in nightly builds), PinMame would perform its simulation on a 60Hz pace, on a separate thread than VPX without synchronization between them.

In the end, I would say that the gameplay latency should be something like this:
1. pinball device latency => likely around or less than 1ms
2. USB communication latency => varying depending on polling rate, I think I read from your posts that it's up to 8.3ms (average to half)
3. OS input API latency => don't know, I would expect it to be fairly low on modern input API but VPX is not using them
4. VPX latency => less than 1ms on BGFX 10.8.1, but up to the display rate on other builds (average to half)
5. PinMame latency => near real hardware if using TimeFence feature, up to 1/60 if not (average to half)
6. VPX latency again for the round trip to get the solenoid output
7. Impact of physical emulation (the solenoid simulation does not go into fast response behavior: we use an overall strength, but no current ramp up, no strength based on core position,...).

With 'FastFlips' steps 5 and 6 are skipped but all the others remain.

Using the numbers above, on a 60Hz display with a fast computer (which in fact increase latency since VPX and PinMame threads stall faster), this should result in the followings (disclaimer: I did not measure/verify most of these yet, so this may be wrong):
- VPX10.8 + PinMame without TimeFence without FastFlips: average of 29ms (1ms + 4ms + 0ms + 8ms + 8ms + 8ms) max around 57ms
- VPX10.8 + PinMame without TimeFence with FastFlips: average of 13ms (1ms + 4ms + 0ms + 8ms) max around 25ms
- VPX10.8.1 BGFX + Latest PinMame with TimeFence without FastFlips: average of  6ms (1ms + 4ms + 0ms + 0.1ms + 1ms + 0.1ms) max around 13ms
- VPX10.8.1 BGFX + Latest PinMame with TimeFence with FastFlips: average of  5ms (1ms + 4ms + 0ms + 0.1ms) max around 11ms

toxieainc:
Great writeup. My additional 2 cents:
I don't think that max/average latency is much of a problem nowadays (especially with your latest changes), BUT the variance in it.
Cause the higher the ms variance, the more inconsistent one can hit shots (brain can adapt to a bit of latency, but one cannot predict the latency variance for each individual shot obviously).

jsm174:
I can't add anything technical here, but @mjrgh, just wanted to say I can't thank you enough for supporting cross platform!!

If this get's into master / 10.8.0, I'll figure out how to get it in standalone branch.

mjrgh:
> I can't add anything technical here, but @mjrgh, just wanted to say I can't thank you enough for supporting cross platform!
> If this get's into master / 10.8.0, I'll figure out how to get it in the standalone branch.

@jsm174 I'm all for portable code, so I'm glad you raised the idea of using hidapi.  And the hidapi rendition of the VPX code is nicer anyway - the native Win32 HID APIs are so ugly.

Re the other branches, I actually did add it to the master branch already, with the original Win32 version.  I could probably redo that with the hidapi code pretty quickly since it should be almost identical to the 10.8.1 version.  Are there any other active branches where you'd want the portable version?  I'm guessing it's not relevant to 9.9.5, but maybe 10.7.2?

mjrgh:
@vbousquet All great observations on the latency situation.  The flipper button work I have in mind isn't for reducing latency, but I do want to make sure it doesn't add new latency, so it's definitely good to think about it in terms how the time sequencing goes.  

Has anyone ever tried to do any practical testing?  I'm thinking it would pretty easy to gather some actual data, since you you can take super-slo-mo video at 120 fps or even 240 fps with a decent phone camera these days.  The one special thing you'd probably want to do is wire LEDs to the flipper switches (in parallel with the key encoder input) so that there's an unambiguous indicator of the exact moment the switch engages.  Set up the camera with the LEDs and video screen in view, and then just count frames between LED activation and on-screen flipper response.

mjrgh:
@toxieainc Interesting point about variation in flipper response latency.  I hold out a little hope that the high-time-resolution flipper button input idea might actually help with that.  With the ~8 ms HID cycle, the latency between pressing the button and VP seeing depends on where you are in the HID polling cycle when you hit the switch - if you push the button right at the start of the cycle, the press won't go anywhere for the whole 8ms, whereas if you get lucky and press it just before the end of the cycle, it might only be a few hundred microseconds.  If the controller passes along information on the timing *within* the cycle, VP might be able to use that to make the injection into the simulator consistent, so that the latency is a more uniform 8ms no matter where you press the button in the cycle.  It's hard to know if that's good or bad; better consistency is good, but maybe the implicit longer average latency is bad.  I personally don't think the difference between 1ms and 8ms latency would be perceptible to a human, although if it takes the total latency from, say, 25ms to 35ms, that might be a different matter, since the threshold of perceptibility is probably somewhere in that range.

PR #1918, opened 2024-09-19
standalone: add support for libvpinball

jsm174:
To better support mobile, iOS and Android, we need a better way to manage app lifecycle.  This PR creates a new `libvpinball` shared library that aims to do this.

For some history:

- The reference implementation for iOS is really just using `SDL_main.c` and passing in some command line args to launch the example table.  
- We added an embedded webserver to make it easier to upload tables. We leveraged the Imgui/LiveUI to change which table loads on startup. To launch a new table, you have to manually kill the app.
- Although this works as a proof of concept, the user experience is terrible.
- We carried this same design over to the Android port.
- Android was recently updated to Kotlin. A list view was added to allow for easily selecting tables. When the table is exited, you can easily select a new table and start again. 
- Even with these improvements, it was realized how difficult it is to share simple things like logging and the web server.

Now that Apple has relaxed the guidelines on emulators, I would like to get Visual Pinball in the App Store. For the best chance of approval, the app has to be super easy to use and have a real user interface.

For the past few weeks I've been working on a new iOS app from the ground up. Bridging the SwiftUI and C/C++/SDL/BGFX world has been interesting. 

Here are some bullet points from this PR:

- new shared library`libvpinball` that can be used on iOS (and later Android, and others)
- the library is exposed as a namespaced C++ class or a C wrapper (Swift uses C)
- client only needs `libvpinball`, `libbass`, and one header file (`VPinballLib.h` or `VPinball.h`)
- all the functionality to build a minimal client application is available:
    - load/save ini values
    - load, play, stop, extract script
    - event callback (window data, rumble using haptics, progress status, etc)
    - get/set/reset table options
    - get/set/reset custom table options
    - get/set/reset view setup
    - toggling fps
    - controlling web server
    - resetting ini back to defaults
    - archive support (uncompress, compress)
    - log support (allows client to log to vpinball.log)
- based much of the shared instance/event/code style on the `VPXPlugin` code
- a new `__LIBVPINBALL__` preprocessor macro has been added to easily see where changes are at

On mobile, the LiveUI is pretty difficult to use because of input strategy and size of screen:

RPReplay_Final1726831768 2024-09-20 07_31_30

A new `LiveUIOverride` ini setting was added, so clients can provide their own UI. If disabled, everything works like it does now. If enabled, it's up to the client, to handle it. 

For example, in the iOS app, we use SwiftUI several `ScrollView`, `Slider` and `Picker` views:

RPReplay_Final1726831272 2024-09-20 07_23_50

This should make adjusting settings much easier for people who know nothing about VPX.

A few final topics for discussion:

- Duplication of View Setup, Table Options, and Custom Table Options code:
   - Reset and default logic seems to be inside of the LiveUI
   - Recreation of these structs in `VPinballLib.h` and `VPinball.h`.
       - Not sure how to easily share in a library without needing tons of other includes
   - LiveUI updates
       - Updates are queued because the client UI can't just update the structures in real time otherwise cause the renderer to crash
   - Custom Options
       - In order to share the options and have a consistent pointer to memory (`const char*`), a new struct value was added where all the literals are a tokenized string `LITERAL||LITERAL||LITERAL`

- Game Loop modification:
   - With the current implementation, if a user would take a screen shot, or receive a notification, iOS would become unresponsive.
   - SDL has a callback function specifically for this situation: `SDL_iPhoneSetAnimationCallback`.
   - To use this `MultithreadedGameLoop` will return immediately and only be called in `VPinballLib`
   - Hopefully this does not break other platforms

- Lifecycle:
  - Currently `g_pvp` needs to be deleted after each game. If not, starting a table will play the last game. (I must be missing something as I deleted `pActiveTable`).
  - Upon table exit, there are memory leaks, so this still needs to be look into.
  - `firstRun` had to be `#ifdef` out. Bass would complain about reinitialization and there would be no audio.
  
- Log file truncation:
   - On iOS, if the log file gets to big, or tables write tons of debug log, it will have a negative effect on game play 
   - We now truncate the log at the start of every game, which is controlled by the `ResetLogOnPlay` ini value
  
- Removal of the existing iOS app:
   - Once the library is integrated and a few last minute bugs worked out, the app will be submitted to the App Store
   - After approval, remove the old iOS CI and put the new iOS app in place
   - The plan would be to use the iOS app as a template for a new Android app
   - Remove web server and launch table code from LiveUI


toxieainc:
In principle looks good and makes sense. But could we maybe share some more code somehow that is now kinda copied (at least thats my impression) in VPinballLib?
So that we do not suffer from one component bitrotting (if somebody forgets to update things on both) over time?
Thanks!!

jsm174:
> In principle looks good and makes sense. But could we maybe share some more code somehow that is now kinda copied (at least thats my impression) in VPinballLib? So that we do not suffer from one component bitrotting (if somebody forgets to update things on both) over time? Thanks!!

@toxieainc, thanks. I ran out of gas last night, so I didn't get a chance to finish my summary. I've updated above. Your concern is exactly something I wanted to discuss!

jsm174:
I am going to merge this.

I would still like to see if anything could be done on sharing some view setup / table options code. I know there is on-going work to the plugin manager, so maybe another place for some shared code.

toxieainc:
I agree, maybe Vince has a good suggestion. I'm currently already multi-tasking too much in my spare time.  ;)
Thanks!

PR #1921, opened 2024-09-21
Pin sim controls (10.8.1)

mjrgh:
This is another "inter-op" proposal, this time for improving coordination between simulators like VP and front ends like PinballX, Popper, and PinballY.  The rationale and details for the proposal are here: http://mjrnet.org/pinscape/PinSimFrontEndControls/PinSimFrontEndControls.htm.

The VP implementation is simple and isolated, so I don't think there's much risk in adding this, but I'm marking this as a draft so that everyone can take a look at the scheme at the design level and decide if the design looks sound.  My main goal was to keep it simple to implement and use while solving a couple of key inter-op problems that all of the front ends have always suffered from and that I think are just intractable without SOME kind of explicit coordination between parties.

By the way, in the course of implementing this, I came across a little bug in VPinball::~VPinball() related to the way CS_CLOSE_APP is handled - the changes to vpinball.cpp are just a fix for that bug, not anything related to my additions (other than that the bug was preventing the additions from working, which is how I found it).  That can be fixed separately if the broader proposal here isn't to everyone's liking.

If this does seem reasonable, I'll submit additional PRs porting it to master, 10.7.2, and 9.9.5.

### DIFF
--- a/src/core/pininput_OpenPinDev.cpp
+++ b/src/core/pininput_OpenPinDev.cpp
@@ -342,11 +342,12 @@ void PinInput::ReadOpenPinballDevices(const U32 cur_time_msec)
    OpenPinballDeviceReport cr { 0 };
 
    // read input from each device
+   bool isNewReport = false;
    for (auto &p : m_OpenPinDevContext->m_openPinDevs)
    {
       // check for a new report
-      if (!p->ReadReport())
-         continue;
+      if (p->ReadReport())
+         isNewReport = true;
 
       // Merge the data into the combined struct.  For the accelerometer
       // and plunger analog quantities, just arbitrarily pick the last
@@ -407,6 +408,10 @@ void PinInput::ReadOpenPinballDevices(const U32 cur_time_msec)
       cr.genericButtons |= r.genericButtons;
       cr.pinballButtons |= r.pinballButtons;
    }
+
+   // if there were no reports, there's no need to update the player
+   if (!isNewReport)
+      return;
 
    // Axis scaling factor.  All Open Pinball Device analog axes are
    // INT16's (-32768..+32767).  The VP functional axes are designed
--- a/src/core/player.cpp
+++ b/src/core/player.cpp
@@ -819,6 +819,9 @@ Player::~Player()
       PLOGE << "Player::OnClose discarded since player is already closing (destructor called from 2 different places...)";
       return;
    }
+
+   // note if application exit was requested, and set the new closing state to CLOSED
+   bool appExitRequested = (m_closing == CS_CLOSE_APP);
    m_closing = CS_CLOSED;
    PLOGI << "Closing player...";
 
@@ -1016,7 +1019,7 @@ Player::~Player()
       m_progressDialog.Destroy();
 
    // Reactivate edited table or close application if requested
-   if (m_closing == CS_CLOSE_APP)
+   if (appExitRequested)
    {
       g_pvp->PostMessage(WM_CLOSE, 0, 0);
    }
--- a/src/core/vpinball.cpp
+++ b/src/core/vpinball.cpp
@@ -2042,12 +2042,32 @@ LRESULT VPinball::OnFrontEndControlsMsg(WPARAM wParam, LPARAM lParam)
 
    case 3:
       // GAME_TO_FOREGROUND
-      // If a player is running, bring it to the foreground.
+      // If a player is running, bring its main window to the foreground
       if (g_pplayer != nullptr && g_pplayer->m_playfieldWnd != nullptr)
-          ::SetForegroundWindow(g_pplayer->m_playfieldWnd->GetCore());
+      {
+         // bring the window to the foreground
+         HWND hwndPlayer = g_pplayer->m_playfieldWnd->GetCore();
+         ::BringWindowToTop(hwndPlayer);
+         ::SetForegroundWindow(hwndPlayer);
+
+         // Windows automatically minimizes full-screen-exclusive windows when they're
+         // moved to the background, but it doesn't automatically restore them when
+         // they're brought back in front, so we have to do this explicitly.
+         if (::IsIconic(hwndPlayer))
+         {
+            DWORD_PTR result;
+            ::SendMessageTimeout(hwndPlayer, WM_SYSCOMMAND, SC_RESTORE, 0, SMTO_ABORTIFHUNG, 100, &result);
+         }
+      }
 
       // handled
       return 1;
+
+   case 4:
+      // QUERY_GAME_HWND
+      // If a player is running, return its main window handle
+      return (g_pplayer != nullptr && g_pplayer->m_playfieldWnd != nullptr) ?
+         reinterpret_cast<DWORD_PTR>(g_pplayer->m_playfieldWnd->GetCore()) : 0;
    }
 #endif // __STANDALONE__
 
--- a/src/core/vpinball.cpp
+++ b/src/core/vpinball.cpp
@@ -2044,21 +2044,7 @@ LRESULT VPinball::OnFrontEndControlsMsg(WPARAM wParam, LPARAM lParam)
       // GAME_TO_FOREGROUND
       // If a player is running, bring its main window to the foreground
       if (g_pplayer != nullptr && g_pplayer->m_playfieldWnd != nullptr)
-      {
-         // bring the window to the foreground
-         HWND hwndPlayer = g_pplayer->m_playfieldWnd->GetCore();
-         ::BringWindowToTop(hwndPlayer);
-         ::SetForegroundWindow(hwndPlayer);
-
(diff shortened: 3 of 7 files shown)


francisdb:
How would this be implemented on standalone for *nix systems?

Where is this frontend closes vpx used? For automated screen recording?

mjrgh:
> How would this be implemented on standalone for *nix systems?

No ideas on that; the IPC mechanism that underlies it is Win32-specific.  Are there any Linux front ends to interface with?  If so, I think a parallel mechanism that suits their needs might be the best way to go.

> Where is this frontend closes vpx used? For automated screen recording?

No, just for normal user-initiated return from a game play session.  As outlined in the spec, the front ends on Windows provide a different sort of experience from the desktop, where the user exits the simulator through a *launcher* control rather than by operating the in-game menus or frame window controls in the simulator (i.e., VP).  Typically this is by pressing a pin cab button that's mapped to a launcher Kill Game command.  The point is to hide the normal "debug/pause/exit" screen and all of the window frame controls and anything else that looks like a normal Windows application UI, so that the pin cab feels like a standalone arcade machine.

vbousquet:
I think this is a great addition and that it will make things more clean, in line with latest cleanups in 10.8.1 (especially close sequence).

One feature that could be nice would be to add a 'capture command' (sent from frontend) that would tell VPX not to target realtime rendering (possible in 10.8.1 since VPX drives VPinMame master clock) but a fixed frame rate (part of the command) and send back each rendered frame. This would allow to capture videos at full framerate on any setup. And since 10.8.1 will likely have auxiliary outputs (backglass, DMD), a parameter of the command could be a mask of what to capture.

toxieainc:
Also looking good from my side...

mjrgh:
> One feature that could be nice would be to add a 'capture command' (sent from frontend) that would tell VPX not to target realtime rendering (possible in 10.8.1 since VPX drives VPinMame master clock) but a fixed frame rate

That seems worthwhile.  How about something like this:

- START_CAPTURE - WPARAM 5, LPARAM high byte contains desired frame rate in Hz, low byte has a mask of target window types (0x01 = main playfield, 0x02 = backglass, 0x04 = DMD, 0x08 = topper, 0x10 = instruction card, 0x20/0x40/0x80 reserved), other bytes reserved.  Return = 0 if not implemented or unable to satisfy request, otherwise a mask with the same meaning as the LPARAM with the frame rate and windows actually locked in
- END_CAPTURE - WPARAM 6, LPARAM 0 (reserved), return code 1 success, 0 failure/unimplemented.  Cancels frame rate lock and returns to normal operation

Is this something that VPX can currently implement? I'm not familiar enough with the DX/DXGI/SDL layer to know if it's possible.

If this needs more hashing out, I can always leave it out of the spec for now and add it once it's more settled.

mjrgh:
I updated the spec proposal at http://mjrnet.org/pinscape/PinSimFrontEndControls/PinSimFrontEndControls.htm with the begin/end capture commands outlined above, if you want to see something more fleshed out.

vbousquet:
> I updated the spec proposal at http://mjrnet.org/pinscape/PinSimFrontEndControls/PinSimFrontEndControls.htm with the begin/end capture commands outlined above, if you want to see something more fleshed out.

Just read it and it looks good to me, the only thing is that the process described tells VPX it is under capture but the capture itself is performed by an external program without synchronization. What I had in mind was to actually ask VPX to perform the frame capture and send back the frame. The benefit of this is that you get a synchronized video, at full frame rate even if the computer was not able to reach full frame rate while capturing.

To be more clear, in this mode, VPX would enter this loop:
- Render a frame
- Send frame back to requester
- Perform emulation for the requested FPS (so advance by 1/FPS)

The actual speed of this would not care as what is important is to have the frame at the right emulation time. Then the requester would assemble the frame into a video (maybe in a parallel task, or simply after receiving enough frames, performing compression all at once).

mjrgh:
> What I had in mind was to actually ask VPX to perform the frame capture and send back the frame. 

I thought you might be thinking along those lines from your original comment, since you did say "send back each rendered frame".  The thing is that that's a bit  beyond what we can do with a cross-process SendMessage(), since that can only pass back the LRESULT - it doesn't marshal pointers across process boundaries.  So we'd have to define a separate IPC mechanism for the frame transmission, like a pipe or shared memory block.

That's all doable, and I agree the video quality would be much improved by doing the frame capture on the VPX side, but this seems to increase the scope rather dramatically.  It would involve not only the extended IPC mechanism, but also adding this new capture mechanism to VPX, and adding consuming code to the front ends.  I don't think any of the existing capture tools would be able to use it as they exist now; PinballY and PBX Recorder just fire off FFMPEG child processes to read from the local graphics device.  I think they'd have to rework their designs around piping from VPX into FFMPEG, reformatting from raw video frames into some kind of streaming container format that FFMPEG can read, which sounds like a big task by itself.

And VPX doesn't currently have anything like a video capture mechanism, does it?  So this would also require designing and building that.  With regard to that, if we're going to have VPX do the frame capture, I think it might make more sense to use some model other than "get next frame".  In particular, it seems like it would be better to just delegate the entire capture process to VPX, letting it control the frame timing, so that it can maintain synchronization with real-time user input (rather than switching to a virtual frame clock provided by an outside program).  VPX could still pass back raw frames to the caller through an IPC sink or just a temp file, but it would stay in control of the frame timing.  That would change the cross-process API a bit, in that there wouldn't need to be a "get next frame" command, just a "start recording" and "stop recording" command.

Anyway, I'm thinking this would be better to detach into a separate project, and just leave anything about capture out of the current thing for now.  If it does end up fitting into the SendMessage() interface as a start recording/stop recording command pair or whatever else, that would be easy enough to add as a new rev of the protocol.

vbousquet:
I fully get your point. This should be implemented on the VPX side first. I will try to have a look at it when I get some time off. Perhaps not 'bloating' the API you designed with START_CAPTURE / END_CAPTURE would be best for the time being.

mjrgh:
Sounds like a plan.  I'll split off the part in the protocol doc to a separate future-plans to capture what we've talked about.

So I guess I'll mark this as ready to go, and I'll plan on creating PRs for 10.8.0, 10.7.2, 9.9.2, and main branch to follow.

I did a little looking around, and I see that ffmpeg will take raw video frames in various pixel formats as input.  So I think we can come up with something pretty simple end-to-end here.  The framework that's taking shape in my head is something like this: front end creates a pipe, creates a foreign-process duplicate of the pipe's write handle for the simulator, hands it to the simulator via the SendMessage scheme (a foreign-process pipe handle is something we can actually move across SendMessage, since it fits in an LPARAM); creates a dup of the pipe read handle to hand to an FFMPEG child process, with stdio redirected to the pipe as FFMPEG's input.  The simulator then just captures raw frames and writes them to the pipe.  That leaves some details still to work out, particularly how to negotiate the raw frame pixel format - presumably the "negotiation" is just that the simulator chooses, but it has to be able to tell the simulator so that it can set up the FFMPEG command line correctly.  Maybe it just writes a header struct that we agree on into the pipe as its first write, and the front end reads that out before launching FFMPEG.

mjrgh:
I updated the main protocol description document to remove the screen capture, and moved it to a separate "future directions" file.  I also fleshed out the new proposal outlined above with the pipe handle into a concrete set of messages and parameters.  I think it's actually approaching something that could work nicely and wouldn't even be a ton of work to implement.

http://mjrnet.org/pinscape/PinSimFrontEndControls/PinSimFrontEndControls.htm
http://mjrnet.org/pinscape/PinSimFrontEndControls/PinSimFrontEndControls-ScreenCapture.htm

jsm174:
So related / not related. On batocera I needed a way to capture videos for my cabinet, so I cheated and used ffmpeg in vpxds. It works surprisingly well and allowed me to capture any of the three screens. I have no idea how it works on windows though. 😄

https://github.com/jsm174/vpxds/blob/40dd4efef380b25c7eadd6403bdb33780955d98f/VPXDisplayServer.cpp#L220

mjrgh:
@jsm174 - that's pretty much exactly how all of the existing recorders work on Windows, they just substitute the local equivalent input device ("gdigrab" or "ddagrab").  They work fine; the only real difference we'd have here is that explicitly piping frames directly from the simulator would make the capture frame-exact. Capturing from the video buffer is less exact because ffmpeg and the simulator are running asynchronously, so the simulator might not generate exactly one frame for every frame that ffmpeg grabs out of the buffer.

toxieainc:
I didn't follow all details, but what should still be done here to get this in? Thanks all!!

mjrgh:
From my perspective it's all ready to go as-is, unless @vbousquet would prefer to defer the whole thing until it has the video-capture additions.  But I'm thinking those can be added as a future expansion so don't need to hold up this initial work.

vbousquet:
For me, it's ready to go in as well. The capture proposal will wait for a later update inside VPX before being exposed to this API.

PR #1926, opened 2024-09-27
Better plugins

vbousquet:


francisdb:
Is there a way to send messages from the plugin to vbscript and vice versa?

vbousquet:
Nope, not yet.

francisdb:
Thanks for the reply

PR #1932, opened 2024-10-02
Update default config for multiple window support

francisdb:
was missing in 9e1b841302e7fa6bfe6cb97c630bbff157b4ca70

Side note: A fallback from the `PlayfieldXyz` to `Xyz` would be great


vbousquet:
Thanks!

PR #1933, opened 2024-10-02
Use hidapi for OpenPinDev implementation

mjrgh:
Reworks the OpenPinDev implementation in master/10.8.0 to use hidapi.lib instead of Win32 native APIs, to make the STANDALONE merge easier.

Added hidapi.lib to the third-party libraries for the build.  This required (as in 10.8.1) renaming hid_init() in hid.h/cpp to resolve the conflict with the one in hidapi.lib, which has the better claim on the name, so I did the same complete renaming pass on those as in 10.8.1, renaming the files to ushock.h/cpp and changing all internal name prefixes from hid_ to ushock_.  The ushock_ naming is more descriptive of this code's purpose anyway.  

I updated the four Windows cmake scripts for these changes (the STANDALONE scripts will presumably need the same changes).


mjrgh:
Making into a draft - noticed some errant logic in the report processor that needs fixing.

toxieainc:
Is there any risk to this of breaking things for somebody (e.g. exotic setups)? Cause we wanted to stabilize 10.8.0 to finally get it out the door. Thanks!!

mjrgh:
I think it's pretty safe - it only affects the OpenPinDev stuff, which doesn't exist in any published devices yet anyway, so no one should even notice if that part isn't working.  Its only connection to the rest of the code is the couple of calls in pininput.cpp (to initialize it and poll it), and those are basically no-ops when no device is connected.  hidapi is a static-link library so it shouldn't cause any configuration issues.  

The only thing I can think of exotic-setup-wise is that hidapi could be dragging in a Windows system DLL that doesn't exist on an non-standard setup, but I don't even think that's much a risk, because the HID-related DLLs are already being dragged in from that old code UShock code that hand-links them.

On the other hand, the only benefit of this PR is to make life easier for the STANDALONE integration.  If that integration isn't part of this release, go ahead and defer this - STANDALONE can incorporate this into their merge instead.

jsm174:
Cool. I'll definitely rebase standalone if this gets in.

toxieainc:
Sounds good, will review and merge, thanks!

toxieainc:
@mjrgh Do you also have release notes for your last couple of changes, or is this all just transparent to the user?

mjrgh:
The switch to hidapi didn't change anything functionally (or hopefully didn't, anyway!), so there's no user impact from that.  I did add documentation for all the new HID features (the velocity inputs and the OpenPinDev stuff) in the docs/ folder.  But I just checked and it looks like I missed the OpenPinDev in the 'master' branch.  You can just copy the one from the 10.8.1 branch into master, or I can open a PR to do that if you'd prefer - it's at docs/Open Pinball Device User Guide.md.

toxieainc:
Will add it, no problem, thanks!!

mjrgh:
Thanks!

PR #1943, opened 2024-10-10
standalone: Pinscape has multiple usb product id's

francisdb:
I have 2 pinscapes, one was not recognized. They both report a different id.

`lsusb` will show this detail:

```
  idProduct          0x00f7 Pinscape Controller
  idProduct          0x00f0 Pinscape Controller
```

The product Id was updated 8 years ago but it seems that a version before that change is still popular.

https://github.com/mjrgh/Pinscape_Controller/commit/dcd8e614d09c68aac8eac2275a615f1dd8630fd1#diff-dba8f2421d03ea929e7f78c769eaa4046472804b124eff804aa9b0c0ecdb9603R136

image

### DIFF
--- a/src/assets/Default_gamecontrollerdb.txt
+++ b/src/assets/Default_gamecontrollerdb.txt
@@ -6,8 +6,10 @@
 # Mac OS
 03000000fafa0000f00000000a000000,mjrnet Pinscape Controller,leftx:a0,lefty:a1~,righty:a2,platform:Mac OS X,
 03000000fafa0000f70000000a000000,mjrnet Pinscape Controller,leftx:a0,lefty:a1~,righty:a2,platform:Mac OS X,
+03004c8009120000eaea000011010000,mjrnet Pinscape Controller,leftx:a0,lefty:a1~,righty:a2,platform:Mac OS X,
 
 # Linux
 03000000790000000600000010010000,DragonRise Inc.   Generic   USB  Joystick,a:b4,b:b1,y:b2,leftshoulder:b0,rightshoulder:b6,dpdown:b5,platform:Linux,
 03004c80fafa0000f000000011010000,mjrnet Pinscape Controller,leftx:a0,lefty:a1~,righty:a2,platform:Linux,
 03004c80fafa0000f700000011010000,mjrnet Pinscape Controller,leftx:a0,lefty:a1~,righty:a2,platform:Linux,
+03004c8009120000eaea000011010000,mjrnet Pinscape Controller,leftx:a0,lefty:a1~,righty:a2,platform:Linux,
--- a/src/assets/Default_gamecontrollerdb.txt
+++ b/src/assets/Default_gamecontrollerdb.txt
@@ -4,12 +4,44 @@
 0300f020790000000600000000000000,Generic USB Joystick,a:b4,b:b1,y:b2,leftshoulder:b0,rightshoulder:b6,dpdown:b5,lefttrigger:b7,righttrigger:b8,platform:Windows,
 
 # Mac OS
-03000000fafa0000f00000000a000000,mjrnet Pinscape Controller,leftx:a0,lefty:a1~,righty:a2,platform:Mac OS X,
-03000000fafa0000f70000000a000000,mjrnet Pinscape Controller,leftx:a0,lefty:a1~,righty:a2,platform:Mac OS X,
+#  Pinscape exclusive PID
 03004c8009120000eaea000011010000,mjrnet Pinscape Controller,leftx:a0,lefty:a1~,righty:a2,platform:Mac OS X,
+#  Pinscape emulating LedWiz
+03000000fafa0000f00000000a000000,mjrnet Pinscape Controller,leftx:a0,lefty:a1~,righty:a2,platform:Mac OS X,
+03000000fafa0000f10000000a000000,mjrnet Pinscape Controller,leftx:a0,lefty:a1~,righty:a2,platform:Mac OS X,
+03000000fafa0000f20000000a000000,mjrnet Pinscape Controller,leftx:a0,lefty:a1~,righty:a2,platform:Mac OS X,
+03000000fafa0000f30000000a000000,mjrnet Pinscape Controller,leftx:a0,lefty:a1~,righty:a2,platform:Mac OS X,
+03000000fafa0000f40000000a000000,mjrnet Pinscape Controller,leftx:a0,lefty:a1~,righty:a2,platform:Mac OS X,
+03000000fafa0000f50000000a000000,mjrnet Pinscape Controller,leftx:a0,lefty:a1~,righty:a2,platform:Mac OS X,
+03000000fafa0000f60000000a000000,mjrnet Pinscape Controller,leftx:a0,lefty:a1~,righty:a2,platform:Mac OS X,
+03000000fafa0000f70000000a000000,mjrnet Pinscape Controller,leftx:a0,lefty:a1~,righty:a2,platform:Mac OS X,
+03000000fafa0000f80000000a000000,mjrnet Pinscape Controller,leftx:a0,lefty:a1~,righty:a2,platform:Mac OS X,
+03000000fafa0000f90000000a000000,mjrnet Pinscape Controller,leftx:a0,lefty:a1~,righty:a2,platform:Mac OS X,
+03000000fafa0000fa0000000a000000,mjrnet Pinscape Controller,leftx:a0,lefty:a1~,righty:a2,platform:Mac OS X,
+03000000fafa0000fb0000000a000000,mjrnet Pinscape Controller,leftx:a0,lefty:a1~,righty:a2,platform:Mac OS X,
+03000000fafa0000fc0000000a000000,mjrnet Pinscape Controller,leftx:a0,lefty:a1~,righty:a2,platform:Mac OS X,
+03000000fafa0000fd0000000a000000,mjrnet Pinscape Controller,leftx:a0,lefty:a1~,righty:a2,platform:Mac OS X,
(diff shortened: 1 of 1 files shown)


francisdb:
@mjrgh mind confirming this?

mjrgh:
Both PIDs are correct, but the premise is wrong. :)  There isn't **A** Pinscape VID/PID; there are 17:

- The 16 VID/PID combinations that the LedWiz uses, when configured to emulate an LedWiz, namely FAFA/00F0 through FAFA/00FF.  The LedWiz uses the range of 16 PIDs as a simple zero-configuration way to distinguish up to 16 units attached to the same system.  At the user level, LedWiz-aware software labels the units with nominal Unit Numbers from 1 to 16, corresponding to PID 00F0 through 00FF.
- An additional VID/PID assigned exclusively to Pinscape via the quasi-official pid.codes registry 1209/EAEA.  This can be configured if LedWiz protocol compatibility isn't desired, which is sometimes useful because the manufacturer's LEDWIZ.DLL contains a crash bug that's triggered if any devices appear on FAFA/00Fx that don't use exactly the same HID report structure as the LedWiz.  If you don't want to replace the manufacturer's DLL (which is the better solution), you can reconfigure Pinscape with the private VID/PID so that the buggy LEDWIZ.DLL doesn't accidentally find it.

The line of code you quoted in the firmware source only sets the factory default.  You can select the VID/PID for a particular device via the Config Tool.  The reason for the change of default from FAFA/00F0 to FAFA/00F8 is that in the early days of Pinscape, many people also had a genuine LedWiz in their system, and a few people had more than one.  Those existing units would always be numbered sequentially from unit 1 (FAFA/00F0), so changing Pinscape's default to unit 8 (FAFA/00F7) got it safely out of the way of essentially everybody's existing collection.  

Whatever software you're using that is hard-coded to a single Pinscape VID/PID just needs to be updated to recognize the range of possible IDs.

mjrgh:
It might also be worth pointing out that Zebsboards once sold (maybe still sells, not sure) yet another LedWiz emulator, called ZB Output Control, that uses VID 20A0/PID 00F0-00FF, so that might be another block to add to the recognition list.

francisdb:
Some more questions:
* would the rest of the sdl2 input device ID stay the same for this ZB device / quasi official pinscape? **03004c80**fafa**0000**f000**000011010000**
* are the inputs the same for those devices? Eg default axis mappings for nudge and plunger.

The `gamecontrollerdb.txt` can be edited by the user so it might be ok to just handle the most common cases being what I already put in the PR?

mjrgh:
I have no idea what that SDL hex string is encoding (I can see the VID/PID in there, but I don't what the rest of it means).  My guess is that it's probably not the same.  Likewise on the ZB devices - I have no guess about what those are going to look like to SDL other than the VID/PID.

Input mappings on the Pinscape devices are somewhat configurable via the Config Tool, so no, you shouldn't assume a fixed axis layout.

If this is a config file that's editable by the user, maybe it would be better to leave it entirely blank and provide some tools for filling it from an automatic scan?  If it has to encode the USB IDs as well as the report layout, I think you're going to have thousands of possible combinations to deal with by the time you account for everything.

mjrgh:
I'm guessing this is all to re-implement the pininput.cpp joystick interface for STANDALONE mode?  If so, then you'd probably be better off using an approach more like what it's doing, of parsing the HID report descriptors to find the joystick fields by HID usage code (X, Y, Z, RX, RY, RZ, Slider, etc) and letting the user do the configuration in terms of the nominal axis names rather than assuming a fixed report structure.  That way you don't have to make a gigantic database of VID/PID values and corresponding pre-defined report structures, or make the user figure out things at that level of detail.  If SDL doesn't abstract HID at that level, you can look at hidapi (which I think is already in STANDALONE) and the hid-report-parser library in #1933.  The config file could then just have a list of the axis assignments, from HID usages to VP functions, basically the same thing you select with the Keys dialog in the Windows version.

francisdb:
I will also add the specific pinscape product id

@jsm174 what do you suggest? Should I add the whole range of id's?
Personally I would just keep the 2 defaults + specific pinscape id and let the users copy paste for their use case where needed.

toxieainc:
I leave this to you guys to figure things out, i have no opinion on this. :)

francisdb:
Added all ID's
But since the PinScape is very configurable we can only provide the more or less default config

As to why this file is in use and how it supposed to be used:
https://wiki.libsdl.org/SDL3/CategoryGamepad
https://github.com/mdqinc/SDL_GameControllerDB

I bought a [Virtuatilt](https://www.3dptronics.com/retro-arcade/virtuatilt-the-ultimate-virtual-pinball-controller-by-3dptronics) and was surprised it did not work out of the box like the pinscape in my cab.

If the current design in vpinball / vpinball standalone could be improved that is out of the scope of this PR, I'm not the guy to tackle that.

mjrgh:
> As to why this file is in use and how it supposed to be used:

Ah, I see, it's an SDL thing.  I don't know much about SDL, but maybe the static setup is deeply enough rooted that something more dynamic (i.e., with a run-time scan of actual live controllers, without the predefined list) isn't practical.  If SDL can accommodate dynamic controller additions, though, maybe at some point it would be nice to add (to VPX, not to SDL) a hidapi-based routine that scans for HID gamepads/joysticks the way Windows does, by Application Collection usage code.

jsm174:
Sorry, I wanted to write earlier about this but didn’t get a chance. (I apologize in advance if I’m repeating things already discussed.)

When I started working on standalone, I wanted to use an Xbox wireless remote since it was the only thing I had.

Using `SDL_Joystick`, I noticed that the button numbers were completely different depending on the device—for example, my iPhone behaved differently than the Apple TV, which behaved differently than the Mac.

After reading SDL’s docs, they recommend using the higher-level interface `SDL_GameController` (called `SDL_Gamepad` in SDL3). `SDL_GameController` is supposed to provide a "common" layout for game controllers and has automatic support for the most common ones. 

So, we added the `SDL_GameController` logic, put it behind a `#define`, and defaulted it to what I thought worked well on the Xbox layout:

[https://github.com/vpinball/vpinball/blob/30f571ca052466623f2370f899ed91a3078b70bd/src/core/pininput.cpp#L142-L155](https://github.com/vpinball/vpinball/blob/30f571ca052466623f2370f899ed91a3078b70bd/src/core/pininput.cpp#L142-L155)

After we got this running on Batocera, I wanted to use the Pinscape for nudging. Mine is set to the default mode, i.e., it behaves as a keyboard instead of a joystick.

Since the Pinscape isn’t really a gamepad, SDL allows you to map joysticks to game controllers via an environment variable or a `gamecontrollerdb.txt` file (see: [SDL3 Gamepad Documentation](https://wiki.libsdl.org/SDL3/CategoryGamepad), [SDL_AddGamepadMapping](https://wiki.libsdl.org/SDL3/SDL_AddGamepadMapping)).

To generate this environment variable or `gamecontrollerdb.txt` file, they provide a tool where you can press each button one at a time, etc.  
[https://github.com/libsdl-org/SDL/blob/main/test/testcontroller.c](https://github.com/libsdl-org/SDL/blob/main/test/testcontroller.c)

To further complicate things, each platform seems to generate its own string, which is why you see the Pinscape controller listed in multiple ways in our gamecontrollerdb:

[https://github.com/vpinball/vpinball/blob/30f571ca052466623f2370f899ed91a3078b70bd/src/assets/Default_gamecontrollerdb.txt#L1-L11](https://github.com/vpinball/vpinball/blob/30f571ca052466623f2370f899ed91a3078b70bd/src/assets/Default_gamecontrollerdb.txt#L1-L11)

---

**TL;DR:**  
I don’t think including all the infinite combinations of Pinscape controller mappings is feasible. I’m not sure if adding Game Controller support was the right approach for standalone. Maybe a built-in utility to help with mapping? 

This topic has been an issue with standalone since the beginning..

francisdb:
As I understand from the SDL docs. Gamepads report what buttons they have compared to some standard layout. Joysticks (like pinscape) do not, that's why we need these lines added to that file.

Can we merge? This is an improvement to what is currently set up and we can always change things later. It's not that 10.8.1 will be released soon.

mjrgh:
> As I understand from the SDL docs. Gamepads report what buttons they have compared to some standard layout. Joysticks (like pinscape) do not, that's why we need these lines added to that file.

I don't know about SDL, but there's no distinction like that at the USB HID level (and Windows likewise makes no such distinction, in DirectInput or any of the other numerous APIs they've added and deprecated over the years).  SDL might have its own reasons for treating them differently, though.

jsm174:
>  Gamepads report what buttons they have compared to some standard layout.

I don't think I full agree with this. Gamepads are just an abstract concept on-top of a joystick. Hence the repo https://github.com/mdqinc/SDL_GameControllerDB.

It looks like SDL has built in HID capabilities: https://github.com/libsdl-org/SDL/blob/b40ad5428bf8eb827875c6654a24b25a3d28c497/src/joystick/hidapi/SDL_hidapijoystick.c which is how they do all the standard controllers. (wondering if any of this could have been leveraged instead of libhid), 

Anyway, I'll merge, but I think we still need to think about this more.

mjrgh:
> Anyway, I'll merge, but I think we still need to think about this more.

Yeah, I have to imagine that the static controller database file could be eliminated or bypassed with a little HID report parsing work.  (Proof by example, in that the exact same functionality on Windows doesn't need such a database.)  And I'm sure that would be nicer for users.  But also just a nice-to-have; I expect that STANDALONE is aimed at more techie types of users anyway, and maybe the more "dot" files they have to deal with, the happier they are. :)

jsm174:
With standalone being include in Batocera, it really opens the door for non-techies -- especially with the latest multi-monitor support. Anyway I would love to make it easier since I feel like it's one on the last pieces of the puzzle. I just need to be told what to do instead of trying to make it up. 😄

PR #1949, opened 2024-10-15
misc: convert to SDL3

jsm174:
All compiling except for linux aarch64 (SDL_ttf / neon issues) which I'll work some more on tomorrow. 

- Areas to look into are commented with `// JSM174`
- Tested B2S, DMD, and PUP, macos, and iOS.


toxieainc:
Nice! Would this mean we could potentially start implementing HDR support now?  :)

src/renderer/Window.cpp
```diff
@@ -178,55 +180,67 @@ Window::Window(const string &title, const Settings::Section section, const strin
...
       if (m_fullscreen)
       {
-         SDL_GetWindowDisplayMode(m_nwnd, &mode);
-         m_refreshrate = mode.refresh_rate;
+         mode = SDL_GetWindowFullscreenMode(m_nwnd);
+         m_refreshrate = mode ? mode->refresh_rate : 0;
          m_screenwidth = m_width;
          m_screenheight = m_height;
-         if (fsRefreshRate > 0 && validatedFSRefreshRate != m_refreshrate) // Adjust refresh rate if needed
+         // JSM174 if (fsRefreshRate > 0 && validatedFSRefreshRate != m_refreshrate) // Adjust refresh rate if needed
```
toxieainc, line 225: No clue about this section, sorry..

src/renderer/Window.cpp
```diff
@@ -178,55 +180,67 @@ Window::Window(const string &title, const Settings::Section section, const strin
...
-                  m_refreshrate = validatedFSRefreshRate;
-                  found = true;
-                  break;
+
+            int num_modes = 0;
+            SDL_DisplayMode **modes = SDL_GetFullscreenDisplayModes(m_adapter, &num_modes);
+            if (modes) {
+               for (int index = 0; index < num_modes; ++index) {
+                 mode = modes[index];
+                 // JSM174 if ((mode->w == m_width) && (mode->h == m_height) && (mode->refresh_rate == validatedFSRefreshRate) && (mode->format == format))
```
toxieainc, line 235: What is unclear here? Cause this is definetly necessary to get the right resolution, etc for fullscreen, right?

src/renderer/Window.cpp
```diff
@@ -236,20 +250,24 @@ Window::Window(const string &title, const Settings::Section section, const strin
...
       }
-      PLOGI << "SDL display mode for window '" << m_settingsPrefix << "': " << mode.w << 'x' << mode.h << ' ' << mode.refresh_rate << "Hz " << SDL_GetPixelFormatName(mode.format);
 
       #if defined(__APPLE__)
          // FIXME remove when porting to SDL3: horrible hack to handle the (strange) way Apple apply DPI: user DPI is applied as other OS but HiDPI of Retina display is applied internally
          // FIXME this only solves the window size, not its position which is in HiDPI units while it should be in pixel units
-         SDL_GLContext sdl_context = SDL_GL_CreateContext(m_nwnd);
-         SDL_GL_MakeCurrent(m_nwnd, sdl_context);
+         // JSM174 SDL_GLContext sdl_context = SDL_GL_CreateContext(m_nwnd);
+         // JSM174 SDL_GL_MakeCurrent(m_nwnd, sdl_context);
```
toxieainc, line 267: Seems to be fine like this

toxieainc:
Looking good to me so far. Thanks!

toxieainc:
Regarding HDR/EDR: This is the best example i found so far: https://github.com/libsdl-org/SDL/blob/main/test/testcolorspace.c

toxieainc:
@jsm174 Do you own a device with HDR/EDR support?

vbousquet:
> Nice! Would this mean we could potentially start implementing HDR support now? :)

Yes, I'm pretty excited with all the features this move would open:
- finally HDR support !
- SDL_GPU which could save us the burden of BGFX (I would prefer SDL_GPU, but it is lacking layered rendering, so needs some contrib to main SDL project) and also perhaps allow to mix SDL_GPU with SDL_Renderer (which has a SDL_GPU backend)
- Support for clean multiple window on multiple display (work in progress in the BGFX but not fully satisfying for the time being, needed for advanced DMD/Alphanumeric/B2S rendering, including VR support),
- the new and better input API would allow to move away from legacy DX in a portable manner,
- the (more or less completely new) sound API would allow to move away from bass and have a fully portable solution (I think that for the time being, we have missing features on other platforms here),
- quite a lot of code cleanups if we decide to add a strong dependency for all the portable things: math, library loading,... (i.e. use SDL3 for all backends, but this likely goes with deprecating DX9)

toxieainc:
HDR will be as much of a game changer as >60fps monitors were, i'm pretty sure. But a lot of existing tables will just output insane brightness levels, too, i fear  ;)

SDL_GPU i personally would wait for some more, as its just so new. But then again you most likely will implement/port it, so if you think its already up for it.. ..

Input / Sound API: Definetly. There is simply no substantial movement anymore on the BASS side. And it was a false compromise right from the beginning.

DX9: Sure, lets bury it. I wonder though: Should we first release 10.8.0 and 10.8.1 'soon'? 10.8.0 at least seems really finished/stable, right? Is 10.8.1 still too much in flux (also no problem)?

As for the SDL_Renderer: How high/low-level is this? Is there a good overview page somewhere?

src/renderer/Window.cpp
jsm174: for some reason, `mode = SDL_GetWindowFullscreenMode(m_nwnd);` and `mode = SDL_GetDesktopDisplayMode(m_display);` always come back as `NULL`, so I was getting into asserts.

I figure maybe if you or @vbousquet attempt to try running on windows, whatever fixes are needed, will ultimately fix this.

jsm174:
> Yes, I'm pretty excited with all the features this move would open:

Any possibilities for `SDL_image` to replace any of `freeimage`?

jsm174:
@toxieainc, @vbousquet  - Linux aarch64 builds are now passing CI.  Feel free to tack on any commits to this PR.

vbousquet:
Just pushed a bunch of fixes. With them, everything seems good for me on Window (limited testing so far).
@jsm174 The commit also solves most parts you marked with // JSM174

The only remaining one is regarding Apple HiDpi that I can't test easily. I still had a look and things should be fairly cleaner since I think SDL3 now reports HiDPI display explicitely through the pixel density factor of the display mode, so using it should do the trick (see pixel_dendity here: https://wiki.libsdl.org/SDL3/SDL_DisplayMode)

vbousquet (approved):

vbousquet:
> HDR will be as much of a game changer as >60fps monitors were, i'm pretty sure. But a lot of existing tables will just output insane brightness levels, too, i fear ;)

Latest table builds are made from bakes so I hope they are likely nearly PBR compliant and should be great in HDR

> SDL_GPU i personally would wait for some more, as its just so new. But then again you most likely will implement/port it, so if you think its already up for it.. ..

I fully agree. The problem is that I can't find a 'perfect' way of having multiple window in BGFX and I'm really frustrated, so I would say that at some point, frustration will be too high and I will give a go at SDL, at least to know if it offers a solution for this.

> DX9: Sure, lets bury it. I wonder though: Should we first release 10.8.0 and 10.8.1 'soon'? 10.8.0 at least seems really finished/stable, right? Is 10.8.1 still too much in flux (also no problem)?

Regarding DX9, no need to bury it too fast. One first step would be to add SDL3 for all platform and leverage all the portability functions it offers (still keeping DX9 based on Win32 or perhaps SDL windowing). It is still the "reference" backend and before burying it, we should validate everything and that's not that easy with regards with the huge feature set VPX has.

Regarding releasing 10.8, I would say yes, just go ahead. It's not perfect but it seems fairly solid.
Perhaps we could document the remaining known bugs:
- SMAA not working for OpenGL (is it ok for DX ?)
- filtering of backglass for OpenGL (never really looked into this one)
- TweakUI in VR (there is a proposed hack for this, but it is not a 'fix' ; it won't work on all headset as it shift the rendering without using the real hardware values)

Regarding 10.8.1, obviously the main change is the standalone player on all platform which is extremely nice and would benefit from a fast release. In the same time, it's still a work in progress branch. Especially, the following things likely need more love:
- the plugin arch has matured a lot. Now, it allows DMD colorization, DMD virtual/physical output with fixed output support, pinup events, headtracking,... But still I would say that the API is really in pre-alpha state. The threading model is ultra restricted (no threading) and it needs more testing / plugins to be validated.
- new tonemapper are nice, but it's missing the AgX punchy
- BGFX would benefit from more testing (f.e. DX12 is not working at all on my side)
- I would say that additional Window output (DMD only for the time being) still have a too high performance impact when on different displays
- the new plumb is not documented and would need more testing.

> As for the SDL_Renderer: How high/low-level is this? Is there a good overview page somewhere?

Regarding SDL_Renderer, this is a 2D renderer somewhat limited. I don't think this is a long-term target but since Jason already implemented B2S and Pup with it, this could be one way to reuse this easily in a cross platform way. (for longer term, I think that the ideal solution would be to implement these in the core renderer)

PR #1955, opened 2024-10-17
fix hid_report_parser.cpp old-style C idiom warnings (10.8.1)

mjrgh:
Fix several warnings in hid_report_parser.cpp (about use of the error-prone old-style C idiom of using an assignment in an `if` without a comparison op).  I thought I had already fixed these a long time ago, but apparently not.


vbousquet:
Thanks !

jsm174:
Thanks @mjrgh !

Issue #1817, opened 2024-06-27
Tweak UI issues in VR


Panasony1008:
Quest3 users are having major issues with not being able to get the F12 tweakUI to display properly in VR.
It appears to look correct on Quest2, but this is likely due to the rendering center being different depending on the headset.
So a function to adjust parallax is needed.

I have created a temporary solution for VPX10.8RC4.
The key point is that in order to adjust parallax with LiveUI::Render(), we have added a new LiveUI::Render(int LR) for VR.
In order to distinguish between rendering for the right and left eye, RenderLiveUI(int LR) in player.cpp has been modified to distinguish between them.
The UI parallax and position are the PosX_diff and PosY_diff values ​​in LiveUI.cpp (L890 L891).

I wanted the diff value to be a parameter so that I could change it in the UI or in the .ini, but I couldn't figure out how to do that.
Please apply to 10.8.0.

[VR Tweak UI fix.zip](https://github.com/user-attachments/files/16010354/VR.Tweak.UI.fix.zip)

MikedaSpike:
I downloaded the master branch, replace the files in the zip above, compiled the gl version. The tweak menu was showing fine !.
I could even adjust  the distance between both eyes and move the menu more to the middle of the screen.
Will this be the last fix for getting 10.8 to a release ? 
@vbousquet did you see this, or you got another fix forvthe tweak menu ?

bbecausereasonss:
Can anyone please compile the fix for the latest build and RC?

MikedaSpike:
@Panasony1008 , why not fork the code,  ake your chang and do a pull request on your fix ?

Panasony1008:
> @Panasony1008 , why not fork the code, ake your chang and do a pull request on your fix ?
Unfortunately, I don't understand the github system, so I don't know how to do it.

MikedaSpike:
Thanks guys ! I really appreciate it!

BA5T1:
Great solution! Thanks for implementing.

PR #1971, opened 2024-10-29
Tweak menu fix for issue #1817 by Panasony

MikedaSpike:
This fix is from Panasony to show the Tweak menu correct within VR


vbousquet:
I will only merge on 10.8 branch, as I think this part of the code needs deeper refactoring (in progress in 10.8.1). Still this will make 10.8 release better. Thanks for the fix!

PR #2002, opened 2024-11-10
Fix relative include path

Nicals:
This MR fixes a path discrepancy I found while playing with the build system.


jsm174 (approved):

jsm174:
Thanks @Nicals !

PR #2010, opened 2024-11-13
fix: plugins building on non-windows os

### LINKED ISSUE #2008: plugins: issues on linux
**Describe the bug**

- [ ] `BOOL` is only available on MS platforms
   When trying to compile `VPXPlugin.h` on Linux I get the following error:
  ```
  VPXPlugin.h:97:44: error: unknown type name 'BOOL'
  ```
  As a workaround one can add a compiler arg `-DBOOL=char`
  (looks like BOOL is defined but only in `MsgPlugin.h`)
- [ ] `./CorePlugin.h:74:4: error: unknown type name 'uint8_t'`
- [x] Code switched from `__STANDALONE__` to `ENABLE_SDL` but `ENABLE_SDL` is nowhere defined in the cmake build files.
- [x] The SDL plugin loading code has not been properly updated to SDL3. Missing `(SDL_SharedObject*)` casts.
- [x] Plugins are no built on linux/macos as there is no build file for those environments

francisdb:
fixes #2008

Add below config to you `VPinballX.ini` to test the `hello.world` plugin. It will show a short `Hello World` message when the table starts.

Starting the serum plugin is currently broken: #2013 

```ini
[Plugin.hello.world]
enable = 1

[Plugin.serum.dmd]
enable = 1
```

Also tested with my rust based plugin https://github.com/francisdb/vpinball-plugin-rust

Pup plugin not enabled as it does not make sense without the closed-source windows dll.


francisdb:
@vbousquet please have a look

francisdb:
Still need to test on macos before this is merged.

jsm174:
Thanks for this. I haven't had a lot of time to really dig into it, so I will defer to @vbousquet. 

I am thinking more than likely this will probably require some changes as we don't prefix any dylibs or so's with arch, and then code signing/notarization and mobile will need to be considered too.

francisdb:
@jsm174 I guess this serum plugin somewhat conflicts with the standalone serum stuff. Anyway, it's disabled if not configured by the user.

vbousquet:
Thanks for having a look for this feature. This is really nice!

I had a first look and here are the little things I changed:
- ENABLE_SDL was an error in MsgPluginManager (it's legacy and has been replaced in VPX by a more precise ENABLE_SDL_xxx per subsystem) => I fixed it on 10.8.1 branch
- LoadPlugin is not supposed to return a value. The state is loaded after LoadPlugin and unload after UnloadPlugin. This makes the API cleaner and simpler. The test in MsgPluginManager was a left over. => I removed it from 10.8.1 branch
- it would be nice if you could squash the commits (merge them in a single commit) in the PR for better readability

The Serum plugin should not conflict with the standalone stuff for the time being as they are used in different places. In the end, the plugin would likely be usable for all uses, but this is a bit early as this feature is still pre-alpha.

PR #2021, opened 2024-11-17
OpenXR: prepare for other platform/grahic backend

vbousquet:


src/renderer/VRDevice.cpp
```diff
@@ -13,13 +13,107 @@
    #include "bx/math.h"
    #include "bx/os.h"
 
-   #define XR_USE_PLATFORM_WIN32
+   #ifdef BX_PLATFORM_WINDOWS
```
francisdb, line 16: This check is wrong. BX_PLATFORM_WINDOWS is always defined and should be `#if BX_PLATFORM_WINDOWS` as it is 0

vbousquet: Good catch. Fixed in latest

PR #2022, opened 2024-11-18
Ensure number of processors is strictly positive

Nicals:
If someone forgot to set the value of `VPinball::m_logicalNumberOfProcessors` (as I did), the default value will be -1.

This can cause a huge number of threads to be spawn due to ThreadPool [stop condition](https://github.com/vpinball/vpinball/blob/master/third-party/include/ThreadPool.h#L113) not taking in account possible negative numbers and spawning nearly infinite threads.

This PR ensures that any use of `VPinball::m_logicalNumberOfProcessors` returns a valid positive number of processors.
A log line will tell the user that something is wrong and the application will run with a very limited number of threads until VPinball class configuration is fixed.

Another solution would have been to raise an exception and crash the application if a not strictly positive.
I prefer the solution of having an error message and the application continue running in non-optimal state.
I can switch to raising an exception if you like it better.

**Before:**

before

**After:**

after

### DIFF
--- a/src/core/main.cpp
+++ b/src/core/main.cpp
@@ -448,9 +448,9 @@ public:
       //!! max(2u, std::thread::hardware_concurrency()) ??
       SYSTEM_INFO sysinfo;
       GetSystemInfo(&sysinfo);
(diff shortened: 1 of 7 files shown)


toxieainc:
Looks good, thanks!

PR #2029, opened 2024-11-27
Decouple save table to storage from win32 gui

Nicals:
This PR prepares the ground for integrating the save table feature to standalone.

Windows specific API has been decoupled by the visitor pattern introduced here.
This also lays the ground for potential gui progress bar to be easily implemented if needed without having to adapt the vpintable.cpp code.

I'm not really happy about putting this SaveTableWin32Visitor.cpp in the `src/parts/` directory.
Any suggestion for a better place ?

A note though: I dont have access to a Windows system just now.
This PR has been developed blindly and I could only check that the code compile properly.
I'll try to get my hands on a Windows system but won't be able to do so until at least a full week.


vbousquet:
I understand that the aim is to decouple the Win32 UI feedback. So I would say this should go in the UI folder.

Regarding the name, I think this should not be named a Visitor (as I don't think it is one) but simply something like `ProgressDialog` with a simplified API (constructor with title and max of progress bar / setProgress / destructor, error being part of this or handled separately).

I'm not easy enough with `standalone` code but maybe there is something similar for table loading. Perhaps @jsm174 can give some hints on that.

Regarding Windows build files, cpp and h files should be added in the corresponding section of the cmake files. These files should also be added to the vcxproj and filter files.

If possible, could you merge the commits together for cleaner history (I know, it's not the best part of the repo, but still it's a bit better) ?

Nicals:
A bit of context about this PR: I'm currently trying to implement an OS agnostic table editor using the Qt framework.
For this, I'm using a version of VPinball built as a library.
My goal with this PR is to be able to inject some feedback receiving service from outside this class.

My aim was to add any type of feedback in this method. My first implementation had a lot more methods then the one I implemented (ex: `imageHasBeenLoaded(std::string imageName)`). This would allow any external program calling this method to add any type of feedback it needed (simple progress bar, growing list of element being loaded, ...).
But I've decided to reduce its complexity since its not really needed.

Currently we have a feedback interface implemented as a progress bar on the current Windows implementation.
This is why I'm not really sure about forcing the interface to handle progress bars and having specific progress bar methods.
Does this makes sense to you ?

If so, I think we can go with a little bit of renaming:

- SaveTableVisitor: parts/VPXFileFeedback
- SaveTableWin32Visitor: ui/dialogs/VPXFileProgressBar

---

There is indeed something similar in the table load process.
All Windows UI related code have been disabled from other compile process with some `#ifndef __standalone__`.
I think the interface could be injected in those methods as well.

@jsm174 How do you feel about this ?
---

I've added the header files in the CMakeFiles.

Concerning merging commits, I also think its a good idea. I will do this as soon as we consider this PR ready to be merged.

toxieainc:
Looks good to me. One thing though: m_progressBar should be init'ed, too, otherwise the whole logic will fail (as it checks for it being nullptr).

jsm174:
Very cool stuff. I'm fine with this too!

toxieainc:
@Nicals Do you plan to extend this more, or is it ready to go?

Nicals:
I'm integrating this architecture on table loading side. This part ([here](https://github.com/vpinball/vpinball/blob/77a4aabc86e7e309b6c283e4251ab9037641e9cc/src/parts/pintable.cpp#L3842)) has two progress bars: one for win32 and one for libvpinball used for mobile build.
The mobile build has multiple progress bar for each kind of items loaded (images, fonts, ...) so I will need to change the public API of the VPXTableFeedback interface to be able to manage both of those UI.

I'm not currently at home and will try to finish this as soon as I'm back (hoping next week). I only need a free afternoon, a good motivation rush and a couple liter of coffee to have it done. So far it has been hard to got all those three requirement to fit together.

jsm174:

PR #2042, opened 2024-12-07
core pinmame callbacks: log errors

francisdb:
Currently errors are swallowed and the script writer has no clue

There might be more locations where this is useful.

### DIFF
--- a/scripts/core.vbs
+++ b/scripts/core.vbs
@@ -2571,17 +2571,54 @@ Sub PinMAMETimer_Timer
 				DMDColoredPixels = DMDp
 			End If
 		End If
-		If UseNVRAM Then
-			If isObject(NVRAMCallback) Then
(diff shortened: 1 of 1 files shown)


francisdb:
One negative is that this will flood the logs

francisdb:
No feedback on this one? Bad idea? Not an issue on windows?

toxieainc:
I wonder if this has any performance implications for endusers in combination with 'badly written tables'. I wonder if this could/should be tied to the table being run in 'play' mode vs from the UI/editor.

francisdb:
I was editing the script and starting the table using standalone. So if this requires a flag it would preferably be available on the command line.

francisdb:
On the other side I don't think some ifs will introduce a perf overhead.

francisdb:
Not very kind to close pr's without feedback.

jsm174:

PR #2048, opened 2024-12-09
standalone: add new Android app

jsm174:
This PR submits a brand new Android app that matches the experience on iOS app. 

I co-developed this with my friend @LeHaine.  Just wanted to give a huge shoutout, as this is my first android app, and I had a lot learn. This wouldn't have been possible without him!

Also some more shoutouts to the testers who gave us lots of feedback during development: @scottacus64, @MrFygar, @nailbuster, and especially @bietekwiet! 

There will be some followup PRs:

- fix bug where LiveUI + fps is not rendering on GLES
- fix bug where SDL3 looses context when leaving app (when you return, you see a black screen)
- finish docs
- submit to PlayStore

I wanted to get these fixed prior to releasing, but I think the app is good enough for people to start messing around with now.


francisdb:
Wondering if it would not be better to split off the android / ios app in their own respective repositories.

jsm174:
> Wondering if it would not be better to split off the android / ios app in their own respective repositories.

I'm fine with them being in this repo, but I will let the community decide.

francisdb:
> > Wondering if it would not be better to split off the android / ios app in their own respective repositories.
> 
> I'm fine with them being in this repo, but I will let the community decide.

Looks like the community is very fast in silently making the decision 😂

toxieainc:
I'm of the opinion that those who do the most work at any given point in time on a project get to decide how they do it fastest/best. My 2 cents.

PR #2077, opened 2024-12-23
android: mark app as game

francisdb:
https://developer.android.com/guide/topics/manifest/application-element

`isGame` is deprecated but seems to still be used a lot as a did a search on github.
https://github.com/search?q=android%3AappCategory%3D%22game%22&type=code&p=2

### DIFF
--- a/standalone/android/app/src/main/AndroidManifest.xml
+++ b/standalone/android/app/src/main/AndroidManifest.xml
@@ -75,7 +75,9 @@
             android:allowBackup="true"
             android:theme="@style/Theme.VPinball"
             android:hardwareAccelerated="true"
-            android:largeHeap="true">
+            android:largeHeap="true"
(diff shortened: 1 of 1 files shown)


jsm174:
Thanks for this. 

I *think* we might need to add some code to get this all working properly:

https://developer.android.com/games/optimize/adpf/gamemode/gamemode-api

something like:

```kotlin
val gameManager: GameManager? = context.getSystemService(Context.GAME_SERVICE) as GameManager?
gameManager?.gameMode = GameManager.GAME_MODE_PERFORMANCE
```

francisdb:
I thought that api was for if you want to make adjustments in the game for eg lowering fps when on battery saver mode. Don't think it's related to the categorisation.

jsm174:
From the docs, its supposed to be able to force a mode.

PR #2094, opened 2025-01-05
standalone: building on Fedora Asahi Remix

francisdb:
Extracted from  #1801 as bgfx works without build file modifications.


jsm174:

PR #2114, opened 2025-01-20
pup: include font style in name

francisdb:
The pup pack fonts dir has this font: `NeueAlteGrotesk-Bold.ttf`
The pup pack itself seems to reference `Neue Alte Grotesk Bold`

```
2025-01-19 16:44:21.919 ERROR [381750] [PUPLabel::PUPLabel@23] Font not found: label=PrevP2Combo, font=Neue Alte Grotesk Bold
```

The code tries to find the fonts by looking in the `m_fontMap` and the `m_fontFilenameMap`. File name matching obviously does not work here.

Normalized font name is `neue alte grotesk bold`

`m_fontMap` used to contain contains this:
Image

Looks like on windows the style name is also added in case it's not regular.

This PR fixes that.

### DIFF
--- a/standalone/inc/pup/PUPManager.cpp
+++ b/standalone/inc/pup/PUPManager.cpp
@@ -149,12 +149,16 @@ bool PUPManager::AddFont(TTF_Font* pFont, const string& szFilename)
       return false;
 
    string szFamilyName = string(TTF_GetFontFamilyName(pFont));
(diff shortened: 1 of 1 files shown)


francisdb:
@jsm174 Please check. I have no idea of anything needs to be freed in this code.

jsm174:
lgtm! Assuming you regression tested our usual pup tables?

PR #2115, opened 2025-01-20
standalone: implement SDL_ttf for textboxes

jsm174:
One thing standalone has always lacked was text for Textbox objects. A few people have wanted this for debugging purposes and a few tables actually use this.

I've tried to replicate the behavior by using SDL3_ttf.

It will try to load a ttf next to the table in the following:

```
<font name with spaces stripped>-BoldItalic.ttf
<font name with spaces stripped>-Bold.ttf
<font name with spaces stripped>-Italic.ttf
<font name with spaces stripped>-Regular.ttf
```

So for example, the example table would try to load:

```
2025-01-20 16:11:39.062 INFO  [14504270] [*Textbox::LoadFont@917] Font loaded: szPath=/Applications/VPinballX_BGFX.app/Contents/Resources/assets/LucidaSansUnicode-Bold.ttf
```

If the font can't be found, then it will display a warning in the log, and default to `LiberationSans-Regular` which is an open source font:

```
2025-01-20 16:09:13.539 WARN  [14500213] [*Textbox::LoadFont@925] Unable to locate font: szPath=/Applications/VPinballX_BGFX.app/Contents/Resources/assets/LucidaSansUnicode-Bold.ttf
2025-01-20 16:09:13.540 INFO  [14500213] [*Textbox::LoadFont@930] Default font loaded: szPath=/Applications/VPinballX_BGFX.app/Contents/Resources/assets/LiberationSans-Regular.ttf
```

Screenshot 2025-01-20 at 4 19 01 PM

It's not perfect, but I guess its something.

@toxieainc:

One thing I definitely am not understanding is:

https://github.com/vpinball/vpinball/blob/7cc481542512b55ebbeefa74dd09699bc1dc7d2b/src/parts/textbox.cpp#L422-L424

This code seems to generate a radically high number, even if I try to divide the font size by 10000. [See here](https://learn.microsoft.com/en-us/windows/win32/api/olectl/ns-olectl-fontdesc).


toxieainc:
@jsm174 Phew, no clue about that one. If i ever changed that part, i don't remember why.  ;)

toxieainc (approved):

vbousquet:
Regarding https://github.com/vpinball/vpinball/blob/7cc481542512b55ebbeefa74dd09699bc1dc7d2b/src/parts/textbox.cpp#L422-L424

I don't know either and did stumble against it without understanding. I t looks like a (clumsy) way of adapting to display size.

Regarding moving to SDL_TTF, I think this is great. I would prefer not doing it through `#ifdef __STANDALONE__` but instead replace the Windows code. One of the aim of 10.8.1 is to integrate as much as possible with standalone toward a common portable code base, dropping most of the Windows dependency when we can. Adding SDL as a dependency on all builds (including DX9) is not a problem and would even be better since it would help remove some of the #ifdef (we have so many of them that testing with reasonable coverage is fairly difficult). I did not yet had time to compare your proposal against existing Windows rendering and check if they (more or less) match. If possible, this would be great. If not possible to do it right now, I'm also ok with the PR as it is atm.

vbousquet (approved):

jsm174:
Thanks. I think for now I will merge this, but I will definitely revisit someday for windows. 

I think we need some kind of FontManager class so we can scan windows fonts to get their paths. I'm not sure how else we could load them with `TTF_OpenFont`. 

- Maybe we can use DirectWrite / IDWriteFontFile?
- Maybe we can load the font binary data from windows and use `TTF_OpenFontRW`?

PR #2122, opened 2025-01-27
Flexdmd plugin

vbousquet:


francisdb:
Nice!  Is this in use by standalone?

vbousquet:
Not yet, but that's the idea

PR #2134, opened 2025-02-07
10.8.1

Le-Syl21:
Fix not working Joystick
Fix double Joystick/Gamepad discovery
Add an option to activate Joystick (which disable Gamepad)
Updated StandAlone Linux Build Doc to give user choice

### DIFF
--- a/src/core/pininput.cpp
+++ b/src/core/pininput.cpp
@@ -311,7 +311,7 @@ void PinInput::RefreshSDLJoystick()
    SDL_JoystickID* joystick_ids = SDL_GetJoysticks(&joystick_count);
    if(joystick_count > 0) {
       for (int idx = 0; idx < joystick_count; ++idx) {
-         m_pSDLJoystick = SDL_OpenJoystick(idx);
+         m_pSDLJoystick = SDL_OpenJoystick(joystick_ids[idx]);
          if (m_pSDLJoystick) {
             m_num_joy = 1;
             if (SDL_IsJoystickHaptic(m_pSDLJoystick)) {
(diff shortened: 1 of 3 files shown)


jsm174:
Thanks, this is really good, something I've been thinking about for a while. 

I'd like to carry it over to other platforms than just Linux, but I really wish we could make this more runtime vs compile time and use an ini value.

Thoughts?

Le-Syl21:
To be honest if you implement Joystick lowlevel I don't understand why keeping Gempad too. And to be really honnest SDL work everyware so it can be great to remove windows code too

vbousquet:
> To be honest if you implement Joystick lowlevel I don't understand why keeping Gempad too. And to be really honnest SDL work everyware so it can be great to remove windows code too

I second that. I think the current state can be explained by the step by step we had over the years building it, but if we can clean up, it would be really nice to have a single & clean path whatever the platform. (note that input is not my skill area, so this would need someone else to review)

jsm174:
As standalone evolved, the only gamepad I had was an XBox One controller.  When using SDL_joystick, if I paired the controller on my iPhone, I got different button numbers than when I paired on my Mac, and then on Android, and Linux.

So after reading all the documentation, SDL_GameController (now SDL_GamePad), was supposed to the recommended way for consistent results.

However, I do see this has caused countless issues on all platforms including Batocera.

I do think we should support both though, because on mobile, out of the box defaults, would support the same controller on multiple devices.

jsm174:
Sorry about this getting closed, if you can re-open it against master instead of 10.8.1.

PR #2145, opened 2025-02-12
standalone: upgrade to toxieainc freeimage fork

jsm174:
I believe this PR is pretty much ready to go. I finally can get a successful FreeImage 3.19.0 build on all platforms. 

I've left in draft, because there is one problem small issue. (maybe something might jump out @toxieainc)

It seems like the colors are flipped:
 
Screenshot 2025-02-11 at 11 13 46 PM

FWIW, we always used a patched `FreeImage.h` in third-party because standalone would get redefined typedef errors. This always bothered me, so I figured I would make a custom `FreeImage.h` with only what is needed.

IMG_638920FBC32F-1


toxieainc:
Hmmm.. Maybe some ifdef is not triggered that flips (or not flips) R & B components?

jsm174:
Yup, you're right.  In the`FreeImage.h` wrapper I made, I forgot to define: 

```
#define FI_RGBA_RED			2
#define FI_RGBA_GREEN			1
#define FI_RGBA_BLUE			0
#define FI_RGBA_ALPHA			3
```

`Texture.cpp` has `#ifdef's` for these. 

Do you think we could get the `CMakeLists.txt` addition and `PluginEXR.cpp` patches in your fork?

toxieainc:
> Do you think we could get the `CMakeLists.txt` addition and `PluginEXR.cpp` patches in your fork?

@jsm174 No prob, where exactly should these go to then?

toxieainc:
I commited the pluginexr, but which cmakelists of the PR should go where?

jsm174:
> I commited the pluginexr, but which cmakelists of the PR should go where?

I ended up putting a PR on your repo, but you can close it. The CMakeLists.txt should go in the root. We just need to make sure that if any new files are added/removed we keep this file up to date. Thanks!

PR #2148, opened 2025-02-12
typo in ubuntu build instructions

francisdb:


jsm174:
fixed in https://github.com/vpinball/vpinball/pull/2145

PR #2153, opened 2025-02-12
core pinmame callbacks: log errors

francisdb:
Currently errors are swallowed and the script writer has no clue. There might be more locations where this is useful.

see previously closed #2042


francisdb:
Probably something that should be hidden between some kind of superstrict flag table authors can enable

francisdb:
🪦 Sadly nobody seems to care about table developer UX

Issue #2119, opened 2025-01-22
standalone: 10.8.1 DMD windows no longer rendered / ZeDMD empty


francisdb:
With the latest `10.8.1` branch commits those windows don't appear and my ZeDMD shows nothing.

Just went back to 86d260387ffa374ed4cdf579303f8f65f65998f3 and there everything works as expected. ZeDMD shows FlexDMD and FlexDMD window visible.

Linux
BGFX build
Happens both for FlexDMD and PinMAMEDMD based tables

francisdb:
Could be related to #2095?

jsm174:
We've been steadily upgrading SDL3 and 3.2.0 was recently released.

@superhac first discovered this issue and we worked backwards.

This is the commit that changed things:

https://github.com/libsdl-org/SDL/commit/b43b333d2521b2de2c55598065ef851c9a37aa81

Can you please try forcing a different renderer?

```
export SDL_RENDER_DRIVER="gpu"
```

https://github.com/libsdl-org/SDL/blob/17625e20dfbcb156ae6a7ea464579d103e3486ce/include/SDL3/SDL_hints.h#L2882-L2914C33

Le-Syl21:
> We've been steadily upgrading SDL3 and 3.2.0 was recently released.
> 
> [@superhac](https://github.com/superhac) first discovered this issue and we worked backwards.
> 
> This is the commit that changed things:
> 
> [libsdl-org/SDL@b43b333](https://github.com/libsdl-org/SDL/commit/b43b333d2521b2de2c55598065ef851c9a37aa81)
> 
> Can you please try forcing a different renderer?
> 
> ```
> export SDL_RENDER_DRIVER="gpu"
> ```
> 
> https://github.com/libsdl-org/SDL/blob/17625e20dfbcb156ae6a7ea464579d103e3486ce/include/SDL3/SDL_hints.h#L2882-L2914C33

I haven't ZeDMD but a B2SDMD
It doesn't work after update.

```
export SDL_RENDER_DRIVER="gpu"
```

Work's for me on Linux :-D thx

jsm174:
It’s not an sdl issue. We’ve just been lucky in the past.

jsm174:
> ZeDMD 4.0.2 still does not work though. Shows the logo when the table boots but then nothing.

@mkalkbrenner is currently working on some updates. When they are done, we'll upgrade libdmdutil and libzedmd. 

[...]

PR #2156, opened 2025-02-13
listres: Show error if sdl init fails

francisdb:
An example case is when compiling sdl on linux with some wayland dependencies missing and then using SDL_VIDEODRIVER=wayland the init will fail.

### DIFF
--- a/src/core/main.cpp
+++ b/src/core/main.cpp
@@ -861,6 +861,12 @@ public:
          std::filesystem::create_directory(PATH_USER);
 
       if (m_listRes) {
+         if (!SDL_Init(SDL_INIT_VIDEO))
+         {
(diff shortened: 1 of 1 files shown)


toxieainc:
Shouldn't this be rather done for the already existing https://github.com/vpinball/vpinball/blob/be3c0add3912317ea6237c1234c8954720051fc5/src/core/main.cpp#L1206 ?

francisdb:
Possibly, I find the huge main very hard to follow with the mix of cli parsing and all these directives. That's why I specifically only fixed `-listres`.

All these `SDL_xxx` commands can fail but errors are mostly ignored in the codebase. There's more cases in the issues like #2119

toxieainc:
Please verify.. Thanks!

francisdb:
> Please verify.. Thanks!

Verified and does not log errors.

PR #2158, opened 2025-02-13
Added new feature 'Where Used' button and dialog to Image Manager

jarringmars-vpx:
New Feature:    Where Used Dialog
How to Access;   New button 'Where Used' on the Image Manager dialog.
Description:
One of the challenges I've found when modding tables is hunting down all the table objects that reference a particular image that I want to update with a new (better one).  That image might be used by many different table objects.  To make this easier I've added a new button labeled 'Where Used' to the Image Manager dialog.  When clicked it pops a new dialog with a listview of all the images in use and the table objects that are referencing them.  You can select any row of the listview and click a button on the dialog labeled 'Edit Control' and it will place that table object into the 'Properties' editor dialog for you.  Then it's a simple matter to update the image property and move onto the next.  No hunting and searching in the layers dialog needed.  Attached is a screen shot that demonstrates this.  
image


toxieainc:
Neat idea and execution, thanks!!

jarringmars-vpx:
Thx...I'm looking at doing something similar with the Materials Manager next.

jarringmars-vpx:
Hi Toxie.  I haven't worked in VS for a while and was fighting the editor a little when I worked on that code.  That's why there were some extra spaces.  I found the settings for tabs/indentation (see below).  Can you share the settings you use.  Thx.  Clay
image

toxieainc:
Don't worry about such details. In principle, as the code base is not the same everywhere (some places have 3, some 4 space indent, some tabs), we just match whatever is locally used.

The 'new norm' should be 3 spaces.

But, as said, don't worry about this. We should think about features, not coding style guides.

PR #2164, opened 2025-02-17
Fix not working Joystick

Le-Syl21:
Easier for buttons than GamePad Mapping


toxieainc:
Looks good to me.. Thanks!

PR #2165, opened 2025-02-18
Add build number and SHA7 to the file properties (#1535)

JockeJarre:
This was an old issue which happened to disappear in the 10.8.1 builds last year.
I have "only" reworked the file properties to show the build and SHA7 on windows.

image


toxieainc:
Great, thanks!

PR #2166, opened 2025-02-19
Added 'Where Used' button and dialog to 'Material Manager'. 

jarringmars-vpx:
Added the 'Where Used' button and dialog to 'Material Manager'.  Moved the WhereUsedDialog code into it's own cpp and header files.

Description:  In the 'Material Manager' dialog there is a new button labeled 'Where Used'.  If clicked the 'Where Used' dialog appears showing a listview of all materials and the table objects referencing them.  Users can select any row and then click the 'Edit Object' button to put that table object into the table 'Properties' editor.  Optionally users can edit a table object by either double-clicking a row or selecting a row and pressing the <enter> key.

The attached screen shot illustrateshow to use this new feature:
image


src/ui/dialogs/ImageDialog.cpp
```diff
@@ -159,6 +168,7 @@ INT_PTR ImageDialog::DialogProc(UINT uMsg, WPARAM wParam, LPARAM lParam)
                    ListView_SortItems(SortData.hwndList, MyCompProc, &SortData);
             }
          }
+         break;
```
toxieainc, line 171: This seems to be wrong? Cause then the code below cannot be reached

jarringmars-vpx:
I might have added a break where I shouldn’t have.

src/ui/dialogs/ImageDialog.cpp
jarringmars-vpx: The break I added on line 171 was definitely a goof on my part.  What's the process to resolve...Should I fix that one line and submit a new pull request?

toxieainc: You can just fix it in your branch and it should pop up here automatically

toxieainc:
Thanks!

toxieainc:
@jarringmars-vpx Oh, i just noticed that you didn't add the 2 new files..

francisdb:
Probably repeating myself but github actions builds on pr's would have avoided merging broken code.

jsm174:
It's all good! I'm wondering if you still need to add the files to the cmakefiles. You can enable actions in your repo, and see if it builds successfully.

jarringmars-vpx:
jsm..is this what needs updated.
image

jsm174:
Should be, you'll need to do it to all flavors of the windows cmake files.

jarringmars-vpx:
Ok,  Thx.  It looks like toxie fixed this up.  Much thx to all and for their patience with me.  I'm new to c++ and visual studio.  Been working with java for the past 7 years.

PR #2168, opened 2025-02-21
add external.sh support for windows builds

jsm174:
~~Lot's to discuss here, so will leave in draft and update description when @jarringmars-vpx fixes windows build issues.~~

This PR attempts to unify the build process across all platforms.

The idea would be to stay consistent, so the scripts expect to be run in a bash shell.  (The github windows runners support this with no additional software.)

```
git clone https://github.com/vpinball/vpinball
cd vpinball
./platforms/windows-x64/external.sh
cp make/CMakeLists_bgfx-windows-x64.txt CMakeLists.txt
cmake -G "Visual Studio 17 2022" -A x64 -B build
cmake --build build --config Release
```

making debug builds is the same:

```
git clone https://github.com/vpinball/vpinball
cd vpinball
BUILD_TYPE=Debug ./platforms/windows-x64/external.sh
cp make/CMakeLists_bgfx-windows-x64.txt CMakeLists.txt
cmake -G "Visual Studio 17 2022" -A x64 -B build
cmake --build build --config Debug
```

`external.sh` generates a directory at `external/<platform>-<arch>/<Release|Debug>` 
`external.sh` also will only build what is necessary.

Some other things:

- Moved all the standalone cmake files into make
- Redesigned all external.sh caching logic to be simpler and only copy specifically what is needed into 

```
third-party/include
third-party/build-libs/<platform>-<arch>
third-party/runtime-libs/<platform>-<arch>
```

- Removed framework logic from iOS cmake and moved into iOS Xcode project
- Removed duplicate macos-x64|arm64 `Info.plist`/`main.mm`
- Fetch all libraries (even for plugins) from third-party folder
- Removed most binaries from repo
- Almost all libs and dlls for window are prefixed with 64
- Serum plugin now uses libserum and libserum64
- Moved pole from standalone into third-party to be shared by plugins
- Build DX versions using visual studio 2022, by installing deprecated v141_xp deprecated components
- Updated all plugin extensions for linux, windows, and macos. (Removed x86 from linux and macos)
- External SHAs moved into `./platforms/config.sh`. No more changing the SHAs in multiple files.

So the biggest issue with this PR is, people will no longer be able to checkout the repo and just compile the app.

Because of this, there are a lot of libraries to compile, and maybe challenging on Windows. I've added additional artifacts to the ci, which is just a copy of the third-party directory. A user could overlay the third-party folder with these. 

Screenshot 2025-02-21 at 12 56 48 PM

(FWIW, making artifacts directly from external is not practical as they are over a gig each.)

~~Finally, I have not updated any of the vs project files in this PR. The biggest changes would be fixing the 64 lib names, etc./~~


jarringmars-vpx:
Hi jsm...what are the windows build issues.  I'm not aware.  The only build issue I run into is when building x86 debug or rls versions.  The plugin-pinmame.vxcproj fails but I haven't been working in that project.

jsm174:
I think your PR forgot to add some files, see:

https://github.com/vpinball/vpinball/pull/2166#issuecomment-2673861961


```
D:\a\vpinball\vpinball\src\ui\dialogs\ImageDialog.cpp(8,1): error C1083: Cannot open include file: 'WhereUsedDialog.h': No such file or directory [D:\a\vpinball\vpinball\build\vpinball.vcxproj]
  (compiling source file '../src/ui/dialogs/ImageDialog.cpp')
  
  LayersListDialog.cpp
  MaterialDialog.cpp
D:\a\vpinball\vpinball\src\ui\dialogs\MaterialDialog.cpp(7,1): error C1083: Cannot open include file: 'WhereUsedDialog.h': No such file or directory [D:\a\vpinball\vpinball\build\vpinball.vcxproj]
  (compiling source file '../src/ui/dialogs/MaterialDialog.cpp')
  ```

jarringmars-vpx:
Ok.  Thx.  Let me take a look.

jsm174:
> Ok. Thx. Let me take a look.

Cool. Thanks. You can also see the results from the builds in the actions tab: 

https://github.com/vpinball/vpinball/actions/runs/13452260202

francisdb:
This is great, all we now need is a script that finds out what os we are on to launch the correct external build. And maybe the same cmake if that is possible? (no more cp)

Is this `.sh_` some kind of convention? Never seen that.

jsm174:
> This is great, all we now need is a script that finds out what os we are on to launch the correct external build. And maybe the same cmake if that is possible? (no more cp)
> 
> Is this `.sh_` some kind of convention? Never seen that.

extra files, by accident. I redesigned the caching and forgot to remove them. will do a force push after discussion.

francisdb:
make/README.md could also use an update showing how to build for each system?

jsm174:
> `make/README.md could also use an update showing how to build for each system?`

Volunteering?

francisdb:
> > `make/README.md could also use an update showing how to build for each system?`
> 
> Volunteering?

I can update it after the merge. Moving over some stuff from he standalone readme?

jarringmars-vpx:
Ok...I just did a new pull request that should include the 2 missing files (WhereUsedDialog.cpp and WhereUsedDialog.h).

jarringmars-vpx:
I created a new branch from master and I do get a build error on the vpinball project but not sure what the issue is.  I don't show any tracked files that changed.  Would it be something that's normally untracked that changed and needs to be added?

jsm174:
Yes, One of SDL_ttf external deps changed, and since our CI is cached not seeing the issue here. But it will be fixed by this PR. Sorry about that.

jsm174:
Updated initial comment: https://github.com/vpinball/vpinball/pull/2168#issue-2869021952

Please let me know your thoughts.

vbousquet (approved):
This is quite a large and needed change, thanks for preparing it!

I did read and tested a full build using it. The following steps were needed:
- install a bash script (failed with w64devkit but succeed with git-bash)
- install CMake
- install Perl (Strawberry)
- install build tools v142 for libdmdutil (while others use the latest v143)
And everything worked perfectly for building the dependencies or the app (both through CMake or using the updated MSVC project files).

On windows, the burden of this changes seems quite limited, that is to say the following:
- for 'simple' build (most users), download the debug/release version of the dependencies and use them (and switch when changing the MSVC build target between Release/Debug)
- for complete build, install the packages listed above, then update and switch between release/debug by a single command

On the other hand, the benefits are quite large:
- single maintainable build approach
- finally a way to easily switch all the dependencies between debug/release (only BGFX had it) !
- far easier way to update libs
- easier way to actually know each dependency version

So, I would say, it can be merged as is.

There are a few nice to have improvements that could be tackled before or after the merge:
- avoid the need of v142 build tools and have all dependencies use the same version
- adjust the README
- maybe remove the MSVC project file in favor of MSVC CMake project support (`launch.vs.json`, `CMakePresets.json`, ...)

toxieainc:
I don't have time to fully test/review this at the moment, so i just second Vincents final paragraph on the improvements.
Is there also a chance to get rid of Perl?

toxieainc:
@jsm174 That would be great! Then i think dependencies would be fine.

Cause git-bash may be installed anyway, and CMake is the default for many projects out there nowadays.

toxieainc:
Looking at the libserialport stuff i think the vc142 dependency is artificial.

vbousquet:
Following our discussion, the update to remove Perl dependency, and the update to libdmdutil to remove v142 use, I've merged the PR. The `README` still need a bit of overall rewrite.

jsm174:
I'm really glad we could get this done, and glad things like bash are more acceptable on windows now.

I think we could still go further, but to answer your question, the SHAs are all in [config.sh](https://github.com/vpinball/vpinball/blob/master/platforms/config.sh).

If for some reason, we needed a different version for a specific platform, then we can override it in the platform specific `external.sh`. I will say though, with all the work for standalone, the SHAs have always been the same across all platforms.

As of right now, here is where we are at:

| Component        | Version                         | Windows | Mobile | Linux/macOS | URL                                         |
|------------------|---------------------------------|---------|--------|-------------|---------------------------------------------|
| SDL              | SHA based on 3.2.4               | ✅      | ✅     | ✅          | [SDL](https://github.com/libsdl-org/SDL)     |
| SDL_Image        | SHA based on 3.2.0               | ✅      | ✅     | ✅          | [SDL_Image](https://github.com/libsdl-org/SDL_image) |
| SDL_TTF          | SHA based on ~master             | ✅      | ✅     | ✅          | [SDL_TTF](https://github.com/libsdl-org/SDL_ttf)     |
| SDL_Mixer        | SHA based on ~master             | ✅      | ✅     | ✅          | [SDL_Mixer](https://github.com/libsdl-org/SDL_mixer) |
| FreeImage        | SHA based on toxieainc 3.19.0    | ✅      | ✅     | ✅          | [FreeImage](https://github.com/toxieainc/freeimage) |
| BGFX CMake       | 1.129.8863-490                   | ✅      | ✅     | ✅          | [BGFX CMake](https://github.com/bkaradzic/bgfx.cmake) |
| BGFX Patch       | SHA based on vbousquet fork      | ✅      | ✅     | ✅          | [BGFX Patch](https://github.com/vbousquet/bgfx)     |
| PinMAME          | SHA based on vbousquet getstate fork | ✅      | ✅     | ✅          | [PinMAME](https://github.com/vbousquet/pinmame)      |
| OpenXR           | SHA based on 1.1.45              | ✅      | ✅     | ✅          | [OpenXR](https://github.com/KhronosGroup/OpenXR-SDK-Source) |
| LibDMDUtil       | SHA based on master              | ✅      | ✅     | ✅          | [LibDMDUtil](https://github.com/vpinball/libdmdutil) |
| LibAltSound      | SHA based on master              | ❌      | ✅     | ✅          | [LibAltSound](https://github.com/vpinball/libaltsound) |
| LibDOF           | SHA based on master              | ❌      | ✅     | ✅          | [LibDOF](https://github.com/jsm174/libdof)          |
| FFMPEG           | SHA based on n7.1                | ✅      | ✅     | ✅          | [FFMPEG](https://github.com/FFmpeg/FFmpeg)           |


Notes:

- FFMPEG is coming soon, as I have this done in vpinball-libs, but needs to be cross compiled on linux
- libdmdutil provides the builds for cargs, sockpp, libserialport, libzedmd, and libserum
- libaltsound can be built for windows, but with the @superhac's SDL3 audio stuff, was waiting
- If I can ever get back to libdof, would provide the build for hidapi. (It did provide the build for the libs for @mjrgh recent commits)

vbousquet:
As things are progressing on the plugin side, I'm now pretty satisfied with how segment works (and PWM segments were really needed even for not really PWM hardware like S11). So now that DMD, display, segments, PinMame and scripting are (mostly) good, I will likely tackle integrating B2S which in turns will benefit from libDOF.

Also with that PR, adding plugin(s) to support PixelCade, ZeDMD,... would be trivial.

I started a plugin for AltSound but stopped as for the time being LibAltSound manages its own audio output while ti could be contributed back to the host (with other sound channels) in order to have a single cleaner mixer setup. So I will wait for SLD_mixer to finish before going on that.

I thought this also included C++ 20 for DX9 build since it was moved to VS2022 (`Build DX versions using visual studio 2022, by installing deprecated v141_xp deprecated components`) but saw this morning I was wrong (sorry for that). If we can have C++ 20 across all targets, it would allow cleaner code, but I'm not sure if this is doable for the DX9 one.

PR #2181, opened 2025-02-25
docs: build instructions

francisdb:


toxieainc:
I personally would like to roughly keep the order as before, so at least starting with the windows quickstart variant, as this is currently still the most common i would say, especially given that we sometimes get contributions from people that have no real clue about larger/more-complex build environments and that want to have a smooth journey to just quickly contribute some fixes/improvements.

jsm174:
I know this was closed, but are we good if we move some of the compile section from standalone to the end of this doc?

francisdb:
You can do anything you want with the contents of this pr. It's just that I won't continue working on it.

PR #2188, opened 2025-02-27
ci: make sure libpipewire is installed

francisdb:
Otherwise SDL3 is compiled without pipewire support.

https://www.phoronix.com/news/SDL-3.0-Prefer-PipeWire

### DIFF
--- a/.github/workflows/vpinball.yml
+++ b/.github/workflows/vpinball.yml
@@ -90,7 +90,7 @@ jobs:
       - if: (matrix.platform == 'linux')
         run: |
           sudo apt-get update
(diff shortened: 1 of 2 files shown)


jsm174 (commented):
Are the releases by version number? We don’t have version numbers for any other dep.

francisdb:
That's not the version of the package, it's part of the name, I guess it's some kind of interface version

https://www.ubuntuupdates.org/package/core/noble/main/base/libpipewire-0.3-modules

jsm174:
ok cool, just double checking. Thanks!

francisdb:
This will require a cache clear or will be applied on the next SDL version update.

PR #2198, opened 2025-03-01
plugins: fix FlexDMD NewVideo and AnimatedActor type

jsm174:
The params for `NewVideo` were backwards and `AnimatedActor` would return the wrong type.

SpongeBob now gets much further:

```
label.Font = DMDDisplay(20,2)
```

`COMToScriptVariant`:

```
case TypeDef::TD_CLASS:
{
   assert(V_VT(&cv) == VT_DISPATCH);
 ```

### DIFF
--- a/plugins/flexdmd/FlexDMD.cpp
+++ b/plugins/flexdmd/FlexDMD.cpp
@@ -263,21 +263,21 @@ Font* FlexDMD::NewFont(const string& font, uint32_t tint, uint32_t borderTint, i
    return pFont;
 }
 
-AnimatedActor* FlexDMD::NewVideo(const string& szVideo, const string& szName)
+AnimatedActor* FlexDMD::NewVideo(const string& name, const string& video)
 {
-   if (szVideo.find('|') != string::npos)
-      return (AnimatedActor*)ImageSequence::Create(this, m_pAssetManager, szVideo, szName, 30, true);
+   if (video.find('|') != string::npos)
+      return (AnimatedActor*)ImageSequence::Create(this, m_pAssetManager, video, name, 30, true);
    else {
-      AssetSrc* pAssetSrc = m_pAssetManager->ResolveSrc(szVideo, nullptr);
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by jsm174** [comment]
@vbousquet - I looked some more into this:

Starting sb, the game calls:

```
DMDMessage1 "GAMESTART"                 ,""                             ,FontSponge16bb ,52,16  ,52,16  ,"solid"        ,2      ,"BG001"        ,"novideo"
```

```
Sub DMDMessage1 (message,message2,Font,posX,posY,endX,endY,state,time,image,video)
        Dim x
        For h = 0 to 9
                If DMDDisplay(h,0) = "" Then
                        DMDDisplay(h,0) = Message
                        DMDDisplay(h,1) = Message2
                        DMDDisplay(h,2) = Font
                        DMDDisplay(h,3) = posX
                        DMDDisplay(h,4) = posY
                        DMDDisplay(h,5) = endX
                        DMDDisplay(h,6) = endY
                        DMDDisplay(h,7) = state
                        DMDDisplay(h,8) = time
                        DMDDisplay(h,9) = image
                        DMDDisplay(h,10) = video
                        if message2 ="" then DMDDisplay(h,11)= 1 Else DMDDisplay(h,11) = 2

                        If h = 0 Then
                                FlexMsgTime = Frame + time * 63 + 5 ' little bit extra so dont blink
                                FlexMsg  = 1   ' START POS ! frame 1
                                Debug.print "I'm HERE"
                                For x = 0 to 12
                                        DMDDisplay(20,x) = DMDDisplay(0,x)
                                Next
                        End If
                        Exit For
                End If
        Next
  ```
  
  Followed by: 
  
  ```
  label.Font = DMDDisplay(20,2)
  ```

If I do an `IsObject()` on `DMDDisplay(20,2)` it returns false. If I call `TypeName()` it fails with a runtime error. 

I think references to the fonts are not being handled correctly when assigned to arrays like this.

Issue #2112, opened 2025-01-17
Prototyping SDL_mixer implementation


superhac:
Hi everyone,

I've been working on implementing the replacement of BASS with SDL_mixer with @jsm174.  I'm currently at the point where I have "rough" table sounds working.  And when I say "rough" I mean no panning, fading, and pitch adjustments.  There also some limitations with SDL_mixer as of today too:

- No 3D sound support for 7.1 (@francisdb has already filed this and its milestone on their roadmap)
- And mixer only supports one sound device to use it at time (I just filed an issue today with them ([issue](https://github.com/libsdl-org/SDL/issues/12005))

Given that there are some limitations that might affect certain uses cases, like SSF it still appears to be a viable path for unified sound across all platforms. I'm starting with stereo and then I will move on to what they currently support today for 3d sound in 5.1.

There would be some work "arounds" to the above mentioned issues.  Technically I could SDL Audio for the BG sounds and lock that to secondary audio device and run the table sounds via mixer at 5.1 on main sound device.  But that's further out out discussion.  

So anyways, I wanted to post this background so when I asked questions you'd have a frame of reference.

My first question is the volume parameter that is being sent when playing sounds.  From what I've so far determined the value range for this param is between 0-100.  But what I've notice for ball rolling sounds is that they mostly operate at the very low end.  Like .30f to 14f.  With SDL_mixer I scale those numbers to a range between 0-128.   When scaled these values are below the silence level(0).   Is there something special about the ball rolling sounds where the volume is treated differently that I'm missing?

Thanks!

francisdb:
This is great news! 

Related issues #1861 and #759
7.1 issue on sdl-mixer https://github.com/libsdl-org/SDL_mixer/issues/573

vbousquet:
This is indeed really great news and highly needed !

[...]

PR #2213, opened 2025-03-04
audio: replace BASS with SDL3 audio

superhac:
Hi folks!

We've made it the point where I think we can open this up to a wider testing audience. Its been quite the journey!  

### Platforms
This change impacts all platforms:  Windows, Linux, Mac, Android, iOS and SBCs.

### Implementation
This replaces the BASS audio system with both SDL Audio and SDL Mixer.  

- SDL Audio (Streaming): PUP (NA on Windows), VPinMame (NA on Windows) and WMP (NA on Windows). 
- SDL Mixer: PlaySound, PlayMusic

Both of these are broken down respectively as backglass and table sounds as it pertains to sound processing within the implementation.  

A single class now called PinSound now controls all the audio.   AudioPlayer, AudioMusicPlayer, etc have either been removed or consolidated.  Now there's a single class no matter what the audio source is that will support it through an interface.

SSF uses the same functions for processing fade and pan values from the table script.  So this should match BASS.  

Pitch was implemented via resampling the audio sample to the matching frequency given by the table.  Since SDL mixer doesn't have a builtin frequency adjusting method, I ended up trying both linear and cubic interpolation algorithms without success.  They both ended up introducing audio artifacts and clipping depending on the sample.   Your forced to interpolation because you can not adjust the length of the buffer when using MixEffects.  So in the end I just went with resampling at the new frequency anytime  the sample is triggered to play with a pitch adjustment.  This had minimal impact on performance in my testing. 

### Limitations / Notes
- Currently SDL_Mixer only supports one concurrent device at a time.  Thus  `SNDCFG_SND3DFRONTISREAR` and `SNDCFG_SND3DFRONTISFRONT` are not supported.  I'm not sure how popular the dual sound card modes are, but that configuration should be taken into consideration.  I did submit a issue for this to be added to their [3.0](https://github.com/libsdl-org/SDL_mixer/issues/661l) version. 

- We had to switch out the sound device id settings in VPinball.ini that used to be numerical to a string that represents the audio device profile.  This does not effect Windows users.  More information on why can be found [here](https://github.com/libsdl-org/SDL/issues/12278l). An example of this from the ini:
```;  Main sound device. Use "VPinballX -listsnd" to get the id's.
SoundDevice = CM106 Like Sound Device Analog Surround 7.1
;  Optional secondary sound device for backglass. Use "VPinballX -listsnd" to get the id's.
SoundDeviceBG = CM106 Like Sound Device Analog Surround 7.1
```
- On exit because the way vpinball shuts down half partical audio streams are put into the buffer that trigger the following errors in the log (harmless but noted):
```
2025-03-04 09:28:24.152 INFO  [2435025] [Player::~Player@1148] Player closed.
2025-03-04 09:28:24.166 ERROR [2435213] [PinSound::Pan2ChannelEffect@1169] unknown audio format..
2025-03-04 09:28:24.166 ERROR [2435213] [PinSound::Pan2ChannelEffect@1169] unknown audio format..
2025-03-04 09:28:24.187 ERROR [2435213] [PinSound::Pan2ChannelEffect@1169] unknown audio format..
2025-03-04 09:28:24.187 ERROR [2435213] [PinSound::Pan2ChannelEffect@1169] unknown audio format..
2025-03-04 09:28:24.209 ERROR [2435213] [PinSound::Pan2ChannelEffect@1169] unknown audio format..
2025-03-04 09:28:24.209 ERROR [2435213] [PinSound::Pan2ChannelEffect@1169] unknown audio format..
```

### Mode Implementation Status

| Mode | Dual Sound Device | Pan | Pitch & Random Pitch | Front to Rear Fade | Notes |
| ---- | ---- | ---- | ---- | ---- | ---- |
| SNDCFG_SND3D2CH | NA | X | X | NA | Supports 2 to 8 channels in this mode. |
| SNDCFG_SND3DALLREAR | NA | X | X | NA | Supports 4 to 8 channels in this mode. |
| SNDCFG_SND3DFRONTISREAR | X |  |  |  | We have to wait for SDL mixer 3.0 for dual sound card support.  https://github.com/libsdl-org/SDL_mixer/issues/661 |
| SNDCFG_SND3DFRONTISFRONT | X |  |  |  | We have to wait for SDL mixer 3.0 for dual sound card support.https://github.com/libsdl-org/SDL_mixer/issues/661 |
| SNDCFG_SND3D6CH | NA | X | X | X | Both SND3D6CH and SND3DSSF use the same SSF pan/fade range.  |
| SNDCFG_SND3DSSF | NA | X | X | X | Both SND3D6CH and SND3DSSF use the same SSF pan/fade range.  |

### Configuration Mapping
| Windows UI Naming | Code Naming Enum | Dual Sound Cards | Code Description |
| ---- | ---- | ---- | ---- |
| Standard 2 Channel | SNDCFG_SND3D2CH | No | Standard 2 channel stereo output |
| Surround (All effects to rear channels) | SNDCFG_SND3DALLREAR | No | All table effects shifted to rear channels.   This can replace the need to use two sound cards to move table audio inside the cab.  Default backglass audio and VPinMame audio plays from front speakers. |
| Surround (Front is front of cab) | SNDCFG_SND3DFRONTISREAR | Yes | Recommended mapping for a dedicated sound card attached to the playfield. Front channel maps to the front of the cab.   We "flip" the rear to the standard 2 channels, so older versions of VP still play sounds on the front most channels of the cab.    This mapping could also be used to place 6 channels on the playfield. |
| Surround (Front is rear of cab) | SNDCFG_SND3DFRONTISFRONT | Yes | Table effects are mapped such that the front of the cab is the rear surround channels.   If you were to play VPX in a home theater system with the TV in front of you, this would produce an appropriate result with the ball coming from the rear channels as it get closer to you. |
| 7.1 Surround (Front is rear, back is side, backbox is front) | SNDCFG_SND3D6CH | No | Rear of playfield shifted to the sides, and front of playfield shifted to the far rear.   Leaves front channels open for default backglass and VPinMame. |
| 7.1 Surround Feedback (enhanced) | SNDCFG_SND3DSSF | No | SSF: 6CH still doesn't map sounds for SSF as distinctly as it could.. In this mode horizontal panning and vertical fading are enhanced for a more realistic experience. |

### Mode to Speaker Map

| Mode | Speaker Channels | Dual Sound Cards | Backglass Speakers *** | Table Sounds Speakers *** |
| ---- | ---- | ---- | ---- | ---- |
| Standard 2 Channel | 2 | No | FL, FR | FL,FR |
| Surround (All effects to rear channels) | 4 | No | FL, FR  | BL, BR  |
| Surround (Front is front of cab) | 4 | Yes | FL, FR (Sound card 1)  | FL, FR (Front of cab), BL,BR (Rear of cab)   (Sound card 2) |
| Surround (Front is rear of cab) | 4 | Yes | FL, FR (Sound card 1)  | BL,BR (Front of cab), FL, FR (Rear of cab) (Sound card 2)    |
| 7.1 Surround (Front is rear, back is side, backbox is front) | 8 | No | FL, FR | BL,BR (Front of Table) - SL, SR (Rear of Table) |
| 7.1 Surround Feedback (enhanced)  | 8 | No | same as above | same as above but uses panning and fading effects to simulate locational sound on the table. |

Any questions or concerns please let me know!


passionforpins:
This sounds exciting and looking forward to test results. Hopefully SDL3 audio plays with WASAPI better than SDL2. Right now the SCUMMVM emulator has to have SDLAUDIO forced to DirectSound in windows environment string as it will not initialize WASAPI. 
Something to watch for.

src/audio/pinsound.cpp
```diff
@@ -2,706 +2,940 @@
...
+         if (audioDevice.name == soundDeviceBGName)
+         {
+            m_sdl_BG_idx = audioDevice.id;
+         }
+      }
+
+      if(m_sdl_STD_idx == SDL_AUDIO_DEVICE_DEFAULT_PLAYBACK) // we didn't find a matching name
+      {
+         PLOGE << "No sound device by that name found in VPinball.ini.  Using Default.";
+         m_sdl_STD_idx = SDL_AUDIO_DEVICE_DEFAULT_PLAYBACK;
```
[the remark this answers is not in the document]
superhac: > This `m_sdl_STD_idx` is already `SDL_AUDIO_DEVICE_DEFAULT_PLAYBACK` at this point. Also it's not because one is misconfigured that the other needs to be reset. I would change this to two if's and just do the logging.

This is just a check to see the values have changed.  The initial values are set to:
```
// SDL Sound Device Id for each output 
int PinSound::m_sdl_STD_idx = SDL_AUDIO_DEVICE_DEFAULT_PLAYBACK;  // the table sounds
int PinSound::m_sdl_BG_idx  = SDL_AUDIO_DEVICE_DEFAULT_PLAYBACK;  //the BG sounds/music
```

That `If` statement your referring to is inside of an `else` where it has determined that you set values in the `VPinball.ini`.  Inside that `else` is where it tries to find a match for that sound device profile name you specified and correlate it to an SDL sound device ID.  After that loop runs the `if` referring to basically sees if that value has changed and if its still set to the default then it means it couldn't find the name of the sound profile you specified.   

For where it stands today we can only use one sound device.  So even if one is wrong and the other is right there both set to the same ID.  

I think this what your talking about?  If I'm misreading what your saying let me know.  Or just paste in code that needs to be changed.  
 
> Further I would:
> 
>     * point the user to the specific setting:`Player:SoundDevice` or `Player:SoundDeviceBG`
> 
>     * show the value the user entered quoted (so he can see eg trailing spaces)

I added the following when the values are bad.

`2025-03-04 12:43:57.683 ERROR [357719] [PinSound::initSDLAudio@119] No sound device by that name found in VPinball.ini. SoundDevice:"NOTREALDEVICE" SoundDeviceBG:"NOT_REAL" Using default.
`

Good catch on this!  Thank you!

superhac:
> This sounds exciting and looking forward to test results. Hopefully SDL3 audio plays with WASAPI better than SDL2. Right now the SCUMMVM emulator has to have SDLAUDIO forced to DirectSound in windows environment string as it will not initialize WASAPI. Something to watch for.

I'm not familiar SCUMMVM or WASAPI.  I did look them up though.   I thought when SDL3 came out with their releaseI read something to the effect about revamped WASAPI code.  But don't quote me on.  You'll have to let us know about this one.

src/audio/pinsound.cpp
```diff
@@ -2,706 +2,941 @@
...
+ * @note Called to play the table sounds via pintable.cpp.  This is main call from the tables for playing sound.
+ */
+void PinSound::Play(const float volume, const float randompitch, const int pitch, 
+               const float pan, const float front_rear_fade, const int loopcount, const bool usesame, const bool restart)
 {
-   UnInitialize();
-
-   if(!IsWav())
+   // Clamp volume
+   float minVol = .08f;  // some table sounds like rolling are extreaming low.  Set a minimum or you cant hear it.
```
toxieainc, line 253: Harmless typo, 'extremely'

src/audio/pinsound.cpp
```diff
@@ -2,706 +2,941 @@
...
+ * @param rearLeft      Output parameter for the rear-left channel gain.
+ * @param rearRight     Output parameter for the rear-right channel gain.
+ */
+void PinSound::calcFade(float leftPanRatio, float rightPanRatio, float fadeRatio, float& frontLeft, float& frontRight, float& rearLeft, float& rearRight)
+{
+   // Clamp fadeRatio to the range [0.0, 2.5]
+   fadeRatio = std::max(0.0f, std::min(2.5f, fadeRatio));
+
+   // Calculate front and rear weights (linear fade)
+   float rearWeight = std::max(0.0f, 2.5f - fadeRatio) / 2.5f; // 1 at fadeRatio=0, 0 at fadeRatio=2.5
```
toxieainc, line 921: fadeRatio was already clamped for both cases here, so is something else intended maybe?

src/audio/pinsound.cpp
```diff
@@ -865,116 +1100,456 @@ float PinDirectSound::FadeSSF(float front_rear_fade)
 
 	if (fabsf(z) < 0.0001f)
 		z = -0.0001f;
-
-	return z;
+	
+	return fabsf(z); // I changed this to get a postive range from 0 to 2.5. not sure why before they returned negative
```
toxieainc, line 1104: If you look at the comment a few lines above this was added as values close to 0 were troublesome, so just a workaround for BASS or maybe the underlying windows API or even some specific drivers.

src/parts/pintable.cpp
```diff
@@ -3005,17 +3000,19 @@ HRESULT PinTable::SaveSoundToStream(const PinSound * const pps, IStream *pstm)
...
    HRESULT hr;
 
+   PinSound * const pps = new PinSound(m_settings);
+
+   // get length of filename
    if (FAILED(hr = pstm->Read(&len, sizeof(len), &read)))
       return hr;
 
-   PinSound * const pps = new PinSound();
+   // read in filenaame (only filename, no ext)
```
toxieainc, line 3015: minor typo: 'filename'

src/parts/pintable.cpp
```diff
@@ -7305,15 +7305,16 @@ STDMETHODIMP PinTable::PlaySound(BSTR bstr, int loopcount, float volume, float p
 
    PinSound * const pps = m_vsound[i];
 
+   // if these are set then its cumulatively added to the values the table script is gemerating.  A bias.
```
toxieainc, line 7308: minor typo: 'generating'

toxieainc:
..looks good to me. How it actually will work out in practice with all table/setup configs, we can let the users test/find then and then fix the remainders.

How does it interact with the altsound support (which is done via BASS there) in PinMAME? Does it also still work in principle?

jsm174:
I think for now VPinMAME / COM component still would use Bass since as AltSound is directly baked into it? I know if we download the artifact it has the Bass dlls. Maybe @droscoe could confirm.

For vpx standalone it uses the libaltsound library and that also still requires Bass as well. 

pinmame/libaltsound could be updated to use SDL audio/mixer as well, or possibly even just expose the audio as a stream and let a plugin take care of it (like its doing for pinmame). I'll defer to @superhac on the stream possibility..

src/audio/pinsound.cpp
superhac: Good catch... That was redundant.

src/audio/pinsound.cpp
superhac: fixed!

src/parts/pintable.cpp
superhac: fixed.

src/parts/pintable.cpp
superhac: fixed.

src/audio/pinsound.cpp
superhac: To be honest I had a hard time with this and was trying to move away from the old routines and use the table values as they are, but to the point in the comment there's a 4 decimal lull in the center 2/4 of the table.  I tried multiple tables and they all end up with this spot with very small (like .004xx) increments for sizeable moves like the ball.    Its definitely not linear.   its all broken out now so maybe @vbousquet can now work on getting this up to par.  I just don't know enough math probably and how those values are translated from the 3D space on the table.

superhac:
> How does it interact with the altsound support (which is done via BASS there) in PinMAME? Does it also still work in principle?

I tested on windows with both the altsound sounds for Baywatch and GNR.  They work fine.  If I understand correctly pinmame runs its own audio using BASS.  So in the end the sound device is mixing the vpinball and pinmame audio on windows.

superhac:
> What I had in mind was to avoid processing the audio data inside libaltsound but instead stream it back to be processed by the library host. This is what is done by libpinmame which sends the audio stream to its host which mixes it and send it to the audio device. The added benefit is that the host (VPX) can perform the overall mixing between sound coming from plugins (PinMame, AltSound) and its own sources (instead of having external setups for each controller).

Yes, were on the same page with that design.   That's one of the lingering questions that I have to get an answer on, can I use mixer capabilities without pumping audio to the sound device.  I have one way around it if it doesn't support that model,  using the POSTMIX callback to capture mix and route it back to vpinball and then clear the buffer so it doesn't play in libaltsound. Kinda hackish but it would work.

> What I did was just to build a plugin skeleton around libaltsound. So nothing really meaningful yet. If you are ready for the challenge, I would be very happy (don't worry for my toes, I have tons of other things to do :) ! ). It's here : https://github.com/vbousquet/vpinball/tree/altsound but since then we also updated the third party library build script (for the better) and things should be fairly easier now on this side too.

Ok sounds great!  I will continue to proceed on that and when I get far enough provide some updates.

jsm174:
everything is rebased with the latest if there is anything to further review.

francisdb:
Kind of hard to follow what changes with the continuous squashing of changes.

jsm174:
If there were changes, they'd be documented. Rebasing with the latest commits from a few hours ago. Anything else?

toxieainc:
I will just pull the trigger then.. ;) Thanks all!!!

toxieainc:
@superhac Could you maybe please track what is still missing/planned as a new issue? Thanks!

jsm174:
Congrat's @superhac!!

superhac:
> @superhac Could you maybe please track what is still missing/planned as a new issue? Thanks!

Yes I will put that in today! Also should I close the original [issue](https://github.com/vpinball/vpinball/issues/2112) discussing SDL audio implementation?

PR #2222, opened 2025-03-09
fix build error on Ubuntu 20.04 with C++20

baxelrod-bdai:
## Changes Made

 1. Fixes a bug when building on Ubuntu 20.04 with C++20.  I guess `constexpr` was only added to `std::string::begin()` in C++23.  This adds a preprocessor macro to only add `constexpr` when using C++23 or greater.

 2. Appease many `-Wunused-result` warnings.

## Testing Done

Built and ran `standalone` on Ubuntu 20.04

### DIFF
--- a/src/audio/pinsound.cpp
+++ b/src/audio/pinsound.cpp
@@ -664,8 +664,14 @@ PinSound *AudioMusicPlayer::LoadFile(const string& strFileName)
 	   pps->m_cdata = (int)ftell(f);
 	   fseek(f, 0, SEEK_SET);
 	   pps->m_pdata = new char[pps->m_cdata];
-	   fread_s(pps->m_pdata, pps->m_cdata, 1, pps->m_cdata, f);
+	   size_t n_read = fread_s(pps->m_pdata, pps->m_cdata, 1, pps->m_cdata, f);
 	   fclose(f);
+
+	   if (n_read == 0)
+	   {
+           ShowError("Error reading file.");
+		   return nullptr;
+	   }
 
 	   const SoundConfigTypes SoundMode3D = (pps->GetOutputTarget() == SNDOUT_BACKGLASS) ? SNDCFG_SND3D2CH : (SoundConfigTypes)g_pvp->m_settings.LoadValueWithDefault(Settings::Player, "Sound3D"s, (int)SNDCFG_SND3D2CH);
 
--- a/src/core/def.h
+++ b/src/core/def.h
@@ -24,10 +24,10 @@
 #endif
 #endif
 
-#if __cplusplus >= 202302L  // C++23 and later
-#define CONSTEXPR constexpr
-#else
+#if defined(__GNUC__) && (__GNUC__ < 12)
 #define CONSTEXPR
+#else
+#define CONSTEXPR constexpr
 #endif
 
 template <typename T>
--- a/src/core/def.h
+++ b/src/core/def.h
@@ -24,6 +24,12 @@
 #endif
 #endif
 
+#if __cplusplus >= 202302L  // C++23 and later
+#define CONSTEXPR constexpr
+#else
+#define CONSTEXPR
+#endif
+
 template <typename T>
 constexpr __forceinline T min(const T x, const T y)
 {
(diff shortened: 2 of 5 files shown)


toxieainc:
Hi!

The constexpr is actually c++20 compatible, so i would suggest to rather check for your version of gcc as a workaround.
As for the unused result warnings, @vbousquet whats your opinion on that? Should we go for workarounding these, or rather try to really handle them? (or leave the code as-is; plan D could even be something like `#define fscanf_s(f, format, ...) { [[maybe_unused]] const int unused = fscanf(f, format, ##__VA_ARGS__); }`)

vbousquet:
If we have the ressource, it would be nice to handle them. something like error |= (fscan(..) != 1), then discard tournament file if error (same for obj,...)

baxelrod-bdai:
Yeah, I'm not exactly sure why this one `constexpr` was causing an issue.  There are certainly lots of other `constexpr` in the code.  Maybe this version of gcc hasn't fully implemented this `constexpr` on `std::string::begin()`?

BTW, it looks like I am actually on 22.04. :facepalm: 

```
$ distro
Name: Ubuntu 22.04.5 LTS
Version: 22.04 (jammy)
Codename: jammy

$ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/11/lto-wrapper
OFFLOAD_TARGET_NAMES=nvptx-none:amdgcn-amdhsa
OFFLOAD_TARGET_DEFAULT=1
Target: x86_64-linux-gnu
Configured with: ../src/configure -v --with-pkgversion='Ubuntu 11.4.0-1ubuntu1~22.04' --with-bugurl=file:///usr/share/doc/gcc-11/README.Bugs --enable-languages=c,ada,c++,go,brig,d,fortran,objc,obj-c++,m2 --prefix=/usr --with-gcc-major-version-only --program-suffix=-11 --program-prefix=x86_64-linux-gnu- --enable-shared --enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --libdir=/usr/lib --enable-nls --enable-bootstrap --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes --with-default-libstdcxx-abi=new --enable-gnu-unique-object --disable-vtable-verify --enable-plugin --enable-default-pie --with-system-zlib --enable-libphobos-checking=release --with-target-system-zlib=auto --enable-objc-gc=auto --enable-multiarch --disable-werror --enable-cet --with-arch-32=i686 --with-abi=m64 --with-multilib-list=m32,m64,mx32 --enable-multilib --with-tune=generic --enable-offload-targets=nvptx-none=/build/gcc-11-XeT9lY/gcc-11-11.4.0/debian/tmp-nvptx/usr,amdgcn-amdhsa=/build/gcc-11-XeT9lY/gcc-11-11.4.0/debian/tmp-gcn/usr --without-cuda-driver --enable-checking=release --build=x86_64-linux-gnu --host=x86_64-linux-gnu --target=x86_64-linux-gnu --with-build-config=bootstrap-lto-lean --enable-link-serialization=2
Thread model: posix
Supported LTO compression algorithms: zlib zstd
gcc version 11.4.0 (Ubuntu 11.4.0-1ubuntu1~22.04) 
```

If you want to check for gcc version, should we check for gcc version > 12?

```
#if defined(__GNUC__) && (__GNUC__ < 12)
#define CONSTEXPR 
#else
#define CONSTEXPR constexpr
#endif
```

And is it worth protecting this for older versions of other compilers?  I'm not sure which version numbers to use, so this would be largely untested:
```
#if defined(__clang__)
    // Clang version check
    #if __clang_major__ >= 15
        #define CONSTEXPR constexpr
    #else
        #define CONSTEXPR
    #endif
#elif defined(__GNUC__)
    // GCC version check
    #if __GNUC__ >= 12
        #define CONSTEXPR constexpr
    #else
        #define CONSTEXPR
    #endif
#elif defined(_MSC_VER)
    // MSVC version check
    #if _MSC_VER >= 1930  // MSVC 2022+
        #define CONSTEXPR constexpr
    #else
        #define CONSTEXPR
    #endif
#else
    #define CONSTEXPR
#endif
```

Regarding the warnings.  I'd also recommend properly handling the return value.  I just don't know the file format and surrounding code well enough to do anything more intelligent.

baxelrod-bdai:
ok, i think i properly handle the return values now.  take another look.

PR #2225, opened 2025-03-11
libvpinball: add support for capturing screenshot via bgfx

jsm174:
The iOS and Android apps capture a snapshot of the screen whenever the table is quit or when the user touches the artwork button in the mobile LiveUI.

On iOS, we were getting a the Metal layer and then using an extension to convert it to a UIImage.

On Android, we were trying to get the Android surface from the SDL Activity.

Android is very sporadic on capture the image. Sometimes it would, but 99% of the time it doesn't. We tried all kinds of tricks (coroutine delays) with no luck.

As it turns out BGFX has the ability to capture a screenshot. So we decided to use that.

BGFX captures all the windows for it's metal renderer, see [here](https://github.com/bkaradzic/bgfx/issues/1833), [here](https://github.com/bkaradzic/bgfx/issues/3249).

After some trial an error we were able to get a patch working that seems to capture a window or the main screen: 
https://github.com/vbousquet/bgfx/commit/e4e6c9c0a2966f8b5a3901cffab5835cb4786d65

Capturing a screenshot results in a callback which we need to execute after a flip. Unfortunately I couldn't figure out how to implement this without using some static variables.

So to capture a screenshot, one would do:

```
g_pplayer->m_renderer->m_renderDevice->CaptureScreenshot(filename.c_str(),
      [](bool success) {
      });
```

We leverage this in iOS and Android by returning `CAPTURE_SCREENSHOT` events to the `VPinballManager` objects.

The screenshot will be saved using SDL3_image as a jpg or png, based upon the filename's extension. The callback is currently `#ifdef __STANDALONE__` as none of the Windows VPX isn't linking SDL3_image yet.

In addition to overhauling screen capture logic, the following other updates have been made:

- on iOS the artwork is no longer stored in SwiftData. Instead a <table>.jpg is placed next to the table. Now when exporting / importing a vpxz file, artwork will be included (to match Android)
- Android was updated to store the artwork as `<table>.jpg` as well.
- Removed creating a Metal Layer in swift and sending it to libvpinball. This now matches MacOS by using: `SDL_GetRenderMetalLayer`
- Fixed the Discord link to point to `#vpx-standalone-mobile` channel
- Added two new contributors to mobile Settings screens, @superhac + @Le-Syl21
- Set the default max texture dimensions on iOS to 1024
- Replace the default ZeDMD Wifi address in android from `192.168.4.1` to `zedmd-wifi.local`
- Fixed iOS simulator builds by properly creating the Bass Framework.
- Enabled logging for DMDUtil plugin


francisdb:
Is this also available for Linux?

jsm174:
Yes, it's available for all BGFX versions. It would need to be wired up. And for windows:

- SDL3_image would need to be linked in
- Remove `#ifdef` here: https://github.com/vpinball/vpinball/blob/aacb7228b151b8e9e6820d8b0b05cf77dad97a44/src/renderer/RenderDevice.cpp#L97-L137

vbousquet:
Actually, this could prove useful to implement a feature we discussed a while ago with mjrgh, that is to say running VPX with a commandline switch to perform a non realtime video capture

francisdb:
I would already be happy to have images stored for each screen when pressing a specific key.

vbousquet:
> I would already be happy to have images stored next to the table for each screen when pressing a specific key.

This is something most (if not all) OS implement seamlessly. I was more on features implying a dedicated behavior inside VPX (get the file for iOS/Android and integrate in the UI, or do not run realtime but at a fixed pace with continuous capture for video)

PR #2255, opened 2025-03-17
flexdmd: limit size to 256 * 64 max

francisdb:
https://vpuniverse.com/files/file/11755-hellraiser/ has the following code:

```vbscript
Const LCDScreen = True 
Dim UseFlexDMD
If Table1.ShowDT = True then
    UseFlexDMD = False
Else
    UseFlexDMD = True
End If
If LCDScreen = True Then
  FlexDMD.Width = 512
  FlexDMD.Height = 128
End If
```

However this segfaults in LibDMDUtil as the buffer is limited to `256 * 64`
https://github.com/vpinball/libdmdutil/blob/f54f084e995782948d30554391ea57a02010d3db/include/DMDUtil/DMD.h#L112

### DIFF
--- a/standalone/inc/flexdmd/FlexDMD.cpp
+++ b/standalone/inc/flexdmd/FlexDMD.cpp
@@ -187,6 +187,11 @@ STDMETHODIMP FlexDMD::get_Width(unsigned short *pRetVal)
 STDMETHODIMP FlexDMD::put_Width(unsigned short pRetVal)
 {
    m_width = pRetVal;
+   if (m_width > 256) {
+      // LibDMDUtil is limited 256 * 64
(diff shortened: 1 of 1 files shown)


vbousquet:
I think the bug is on the LibDMDUtil side. So, it would be better to fix it there (or discard DMD that are not supported) than spreading this limit to other components.

francisdb:
@jsm174 what's your thoughts on this? I can increase the buffer on the util side but I guess this needs to be validated/limited somewhere as the script can set it to whatever it wants?

jsm174:
@vbousquet beat me to it. This should be fixed on the libdmdutil side. And there is more work needed to libdmdutil then just increasing the buffer.  cc @mkalkbrenner

francisdb:
To be continued in https://github.com/vpinball/libdmdutil/issues/65

PR #2261, opened 2025-03-19
standalone: log missing sounds

francisdb:
I was debugging why Hellraiser is not playing background music and found out that missing sounds are only logged on windows.

What I want:
* Missing sounds are logged as `WARN` (marked yellow in log)
* This can be optional by setting some config value as some badly written tables try to play ball roll sounds that don't exist. For example this Hellraiser table defines 20 balls in the ball roll sound code but there are only 10 `fx_ballrolling#` sounds.

Went for the simplest solution but I'm expecting some feedback on this one.

On windows there is a debug screen.
There is an option to redirect log output to that screen.
The sound code is a a bit special in that it directly sends to that debug screen without going through the logs. 
On standalone there is no debug screen `m_hwndDebugOutput`. Therefore no missing sounds are logged.

There is also `DebuggerModule::Print` but this does not have any log levels.
This feature is controlled by a setting `Editor::EnableLog` and `Editor::LogScriptOutput` with different defaults depending on standalone.

Questions:

* How do we make this logging optional and where should this code that checks the setting before logging be placed? I guess there should not be multiple places where the same setting is checked?
* We might also want to log warnings if sounds are being stopped that don't exist?

### DIFF
--- a/src/parts/pintable.cpp
+++ b/src/parts/pintable.cpp
@@ -7206,12 +7206,23 @@ HRESULT PinTable::StopSound(BSTR Sound)
    WideCharToMultiByteNull(CP_ACP, 0, Sound, -1, szName, MAXSTRING, nullptr, nullptr);
    const string name(szName);
 
-   for (size_t i = 0; i < m_vsound.size(); i++)
+   size_t i;
+   for (i = 0; i < m_vsound.size(); i++)
       if (StrCompareNoCase(m_vsound[i]->m_szName, name))
-      {
-         m_vsound[i]->Stop();
          break;
+
+   if (i == m_vsound.size()) // did not find it
+   {
+      if (!name.empty() && !m_soundsMissing.contains(name))
+      {
+         PLOGW << "Request to stop \"" << name << "\", but sound not found.";
+         m_soundsMissing.insert(name);
       }
+      return S_OK;
+   }
+
+   PinSound * const pps = m_vsound[i];
+   pps->Stop();
 
    return S_OK;
 }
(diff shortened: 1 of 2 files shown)


vbousquet:
I did not dive into the code but I would guess this is legacy behavior. I added the logging library (PLOG) and its optional link to DebugWindow in 10.8 but there may be some old code doing this type of logging (directly wiring to an appender) which has not been refactored yet. I would just change all logging of this type to go through PLOG and let PLOG appenders to dispatch to the right output. For sure, I would avoid anything that is platform dependent on this issue

francisdb:
Would it not be better to store sounds in a `robin_hood::unordered_map` with lowercase keys instead of having to iterate all sounds and `StrCompareNoCase` each time?
I do have seen tables with duplicated sounds names so those would need to be skipped everywhere to avoid allocation issues
Anyway that would be for a different pr.

francisdb:
@jsm174 what is this waiting for?

jsm174 (changes requested):
We had something similar to this in PUPLabel:

```
void PUPLabel::UpdateLabelTexture(SDL_Renderer* pRenderer, SDL_Rect& rect)
{
   static robin_hood::unordered_map<string, robin_hood::unordered_set<string>> warnedLabels;

```

Any chance you could use `robin_hood` as that's what the rest of the code base uses. 

(Doesn't have to be static like PUPLabel - it's the wrong implementation now - when it was coded mobile app didn't allow for restarts, so static wasn't an issue)

The prefix `m_us`, took me a little bit to understand it what the prefix stood for. I think in these cases you don't really need the `us`.

jsm174 (approved):

toxieainc:
> Thanks. Didn't even know we switched to ankerl. I'll have to revisit some of the pup code.

I think i already catched all. `unordered_dense` is the follow up to `robin_hood` from the same author.

PR #2265, opened 2025-03-21
pup: reduce warnings

francisdb:
* load playlists only once
* move exact path not found warnings to info
* do not drop playlists with empty folders
* add custom function trigger

If somebody could check that the playlist copy I do for each screen is valid. I'm still wondering if that is actually required but currently each screen loads all playlists.


jsm174 (commented):
Can you provide some more details on your changes and I’ll review. Thanks!

standalone/inc/pup/PUPTrigger.cpp
```diff
@@ -13,12 +13,13 @@ const char* PUP_TRIGGER_PLAY_ACTION_STRINGS[] = {
    "PUP_TRIGGER_PLAY_ACTION_STOP_FILE",
    "PUP_TRIGGER_PLAY_ACTION_SET_BG",
    "PUP_TRIGGER_PLAY_ACTION_PLAY_SSF",
-   "PUP_TRIGGER_PLAY_ACTION_SKIP_SAME_PRTY"
+   "PUP_TRIGGER_PLAY_ACTION_SKIP_SAME_PRTY",
+   "PUP_TRIGGER_PLAY_ACTION_CUSTOM_FUNC"
```
francisdb, line 17: for `142,1,TransSet,D0,15,"{""mt"":301, ""SN"":7, ""FN"":19, ""TM"":1, ""TC"":16777215}",,,,,,,CustomFunc,0`

standalone/inc/pup/PUPScreen.cpp
```diff
@@ -169,35 +174,19 @@ PUPScreen* PUPScreen::CreateDefault(int screenNum)
...
-         pScreen = new PUPScreen(PUP_SCREEN_MODE_SHOW, PUP_SCREEN_PLAYFIELD, "Playfield", "", "", false, 100.0f, nullptr);
+         pScreen = new PUPScreen(PUP_SCREEN_MODE_SHOW, PUP_SCREEN_PLAYFIELD, "Playfield", "", "", false, 100.0f, nullptr, playlists);
       default:
-         pScreen = new PUPScreen(PUP_SCREEN_MODE_SHOW, screenNum, "Unknown", "", "", false, 100.0f, nullptr);
+         pScreen = new PUPScreen(PUP_SCREEN_MODE_SHOW, screenNum, "Unknown", "", "", false, 100.0f, nullptr, playlists);
    }
    return pScreen;
 }
 
-void PUPScreen::LoadPlaylists()
```
francisdb, line 185: Each screen used to load the playlists file. Now this is only done once in the manager. Passed to the screens when they are created and a defensive copy is performed there to make sure each screen has it's own playlists as before.

Later this can also be done for the triggers but there it's a bit more complicated as a trigger has access to a screen and vice versa.

standalone/inc/pup/PUPScreen.cpp
```diff
@@ -79,7 +79,12 @@ PUPScreen::PUPScreen(PUP_SCREEN_MODE mode, int screenNum, const string& szScreen
    m_pPageTimer->SetElapsedListener(std::bind(&PUPScreen::PageTimerElapsed, this, std::placeholders::_1));
    m_pParent = nullptr;
 
-   LoadPlaylists();
+   for (const PUPPlaylist* pPlaylist : playlists) {
+      // make a copy of the playlist
+      PUPPlaylist *pPlaylistCopy = new PUPPlaylist(*pPlaylist);
```
francisdb, line 84: Just keeping things as the old code, not sure if this copy is 100% required, depends on if playlists are mutated during gameplay.

standalone/inc/pup/PUPPlaylist.cpp
```diff
@@ -106,8 +106,8 @@ PUPPlaylist* PUPPlaylist::CreateFromCSV(const string& line)
    }
 
    if (!hasFiles) {
-      PLOGD.printf("Playlist folder is empty");
-      return nullptr;
```
francisdb, line 110: Empty playlists were skipped, this then later caused warnings if they are referenced from triggers. We could later add this to a pup pack audit.

standalone/inc/pup/PUPManager.cpp
```diff
@@ -104,11 +112,35 @@ void PUPManager::LoadConfig(const string& szRomName)
    return;
 }
 
-const string& PUPManager::GetRootPath()
+void PUPManager::LoadPlaylists()
 {
-   if (!m_init) {
-      PLOGW.printf("Getting root path before initialization");
+   // TODO remove duplicates
```
francisdb, line 117: Implementing this now

standalone/inc/pup/PUPManager.cpp
```diff
@@ -104,11 +112,35 @@ void PUPManager::LoadConfig(const string& szRomName)
    return;
 }
 
-const string& PUPManager::GetRootPath()
+void PUPManager::LoadPlaylists()
 {
-   if (!m_init) {
-      PLOGW.printf("Getting root path before initialization");
```
francisdb, line 110: In pup packs the root path is sometimes accessed before the player is initialized. This used to return nullpointers to the vbscript, now it's initialized in the constructor and this warning is no longer required.

standalone/inc/pup/PUPManager.cpp
```diff
@@ -70,7 +77,8 @@ void PUPManager::LoadConfig(const string& szRomName)
    for (auto& [key, pScreen] : m_screenMap) {
       PUPCustomPos* pCustomPos = pScreen->GetCustomPos();
       if (pCustomPos) {
-         PUPScreen* pParentScreen = GetScreen(pCustomPos->GetSourceScreen());
```
francisdb, line 73: Chicken and egg problem. GetScreen logs warnings if screens are referenced that don't exist (yet). However the code that populated the screens was already triggering these checks. So we directly access the m_screenMap instead.

standalone/inc/pup/PUPMediaManager.cpp
```diff
@@ -87,11 +87,11 @@ void PUPMediaManager::Stop(PUPPlaylist* pPlaylist, const string& szPlayFile)
 {
    string szPath = pPlaylist->GetPlayFilePath(szPlayFile);
    if (!szPath.empty() && szPath == m_pMainPlayer->szPath) {
-      PLOGD.printf("Main player path match: screen={%s}, path=%s", m_pScreen->ToString(false).c_str(), szPath.c_str());
+      PLOGD.printf("Main player stopping playback: screen={%s}, path=%s", m_pScreen->ToString(false).c_str(), szPath.c_str());
       m_pMainPlayer->player.Stop();
    }
    else {
-      PLOGW.printf("Main player no path match: screen={%s}, path=%s", m_pScreen->ToString(false).c_str(), szPath.c_str());
+      PLOGD.printf("Main player playback stop requested but currently not playing: screen={%s}, path=%s", m_pScreen->ToString(false).c_str(), szPath.c_str());
```
francisdb, line 94: This is normal, if something can happen at multiple places multiple things might be going on and any of those would for example need to be stopped.

jsm174:
Thanks for providing all the details. Helps me out big time. Let me review the copy logic and I'll merge. Thanks!

jsm174 (approved):

PR #2274, opened 2025-03-26
fix: bring back disabled vsync

### LINKED ISSUE #2267: No longer possible to disable vsync
Linux
10.8.1 master branch
BGFX

```
INFO  [Player::Player@351] Synchronization mode: 0 with maximum FPS: 10000, display FPS: 59.98
```

However the fps screen (F11) shows ~60FPS

```ini
; Sync the frame rate with the refresh rate of your monitor
; 0 - None: No synchronization.
; 1 - Vertical Sync: Synchronize on video sync which avoids video tearing, but has higher input latency.
; 2 - Adaptive Sync: Synchronize on video sync, except for late frames (below target FPS), also has higher input latency.
; 3 - Frame Pacing (default): Targets real time simulation with low input- and video-latency (also dynamically adjusts framerate).
SyncMode = 0
; Maximum frame rate. Will influence input latency unless frame pacing is configured.
; Defaults to the playfield display refresh rate.
; Use 0 for unbound frame rate.
; Will apply a minimum of 24fps.
MaxFramerate = 0
```

francisdb:
This fixes unlocked fps in the case of:
```ini
SyncMode = 0
MaxFramerate = 0
```

Regression  happened in 317bf42
Not sure if this is the best/correct way to fix this. Higher up in the code there is already `needsVSync = useVSync && !noSync` so that `!noSync &&` in the `if` is probably not needed.

However after some more testing there is still some strangeness for MaxFramerate > monitor refresh rate

https://github.com/vpinball/vpinball/blob/ded4119928d1979e3a7bd24b4f50a1470fe8229e/src/renderer/RenderDevice.cpp#L585

I wonder what this magic number `2000` is here and why this `clamp` is used. It causes frame rates of 68,2 fps for any `MaxFramerate` higher than the display refresh rate (60hz in my case). I would expect it to end up with a 90fps in below case.

```ini
SyncMode = 0
MaxFramerate = 90
```

_Further I also see that the `MaxFramerate = 0` is no longer required to have unlocked fps. Leaving the value empty is also working now._

fixes #2267

### DIFF
--- a/src/renderer/RenderDevice.cpp
+++ b/src/renderer/RenderDevice.cpp
@@ -572,7 +572,7 @@ void RenderDevice::RenderThread(RenderDevice* rd, const bgfx::Init& initReq)
          }
 
          // If the user asked to sync on another frame rate than the refresh rate, then perform manual synchronization
-         if (!noSync && (!needsVSync || (g_pplayer->GetTargetRefreshRate() != rd->m_outputWnd[0]->GetRefreshRate())))
+         if (!noSync && (needsVSync || (g_pplayer->GetTargetRefreshRate() != rd->m_outputWnd[0]->GetRefreshRate())))
          {
             g_pplayer->m_renderProfiler->EnterProfileSection(FrameProfiler::PROFILE_RENDER_SLEEP);
             #ifdef MSVC_CONCURRENCY_VIEWER


vbousquet:
The issue was more complex than the proposed fixed (which was just a revert of a fix of another problem), so I did a clean fix & PR here: https://github.com/vpinball/vpinball/pull/2329

PR #2277, opened 2025-03-29
fix: improve joystick/gamepad input handling and configuration

Le-Syl21:
This commit introduces significant improvements to input handling:

1. Fixed a bug where the physical button 0 would add credits instead of starting a game due to conflicting default mappings.

2. Separated gamepad and joystick handling by:
   - Adding a new InputLayout::Gamepad type
   - Using specific default mappings for standard gamepads
   - Setting all joystick default mappings to 0 to avoid conflicts

3. Added proper support for two previously unmapped actions:
   - Escape key (JoyEscapeKey)
   - Enable 3D mode (JoyEnable3DKey)

4. Improved SDL device initialization:
   - Removed redundant device refresh in SDLInputHandler constructor
   - Streamlined cleanup process in destructors
   - Fixed potential double-initialization of joysticks

5. Cleaned up configuration file management:
   - Removed unnecessary validation for joystick inputs in settings.cpp
   - Prevented creation of default entries with value 0 in the config file
   - Only user-configured joystick mappings will appear in the config file

6. Updated gamecontrollerdb.txt:
   - Removed pinball cabinet specific controllers that should be handled as joysticks
   - Integrated comprehensive gamepad mappings from https://github.com/gabomdq/SDL_GameControllerDB
   - Improved support for a wide range of standard game controllers

These changes provide a more predictable input experience with proper separation between gamepad and specialized joystick devices, while also improving performance and configuration management.


toxieainc:
Oh whoops @vbousquet, i missed your self assigned review. It looked good to me, sorry!!

vbousquet:
Ooppss, there are quite a lot of regression here. Could you revert the merge ? I will have a look when I have time but it's definitely not ready for merge

Le-Syl21:
> Ooppss, there are quite a lot of regression here. Could you revert the merge ? I will have a look when I have time but it's definitely not ready for merge

I think I create regression on windows. Like you've create on SDL. Can you talk together to think about replacement of Directinput by SDL to be multiplatform ???

vbousquet:
> "I think I'm creating a regression on Windows. Similar to what you've created with SDL. Can we discuss replacing DirectInput with SDL to make it multiplatform?"

Yes, that's exactly the reason I initially added myself as a reviewer to be able to discuss the different point and converge. The merge/revert was just a coordination mistake. The idea is to find a solution that fits in well.

Here are the point I did not have the time to write initially:
- Regarding the settings, you have removed the `Validate` things which is in fact going backward with where I try to head things. At the moment, we have a few issues with settings which I try to (slowly) solve:
  - Settings are usually defined multiple times (at least when used and in the UI dialog), not always with the same default value,
  - Most of the time settings are not validated against valid values, leading to occasional crashes when they are out of range,
  - Settings (and the default ini I introduced in 10.8) are a hell to maintain as they do not share a common defintion
  - Settings are slow leading to lots of state duplication and static behavior, this could be easily changed by using array access for predefined settings (the one not dynamically defined by plugins for example)

  The idea is to have all static settings to be defined in one place, the `Settings.Validate` method, forbid defining static settings elsewhere, forbid having defaults wandering a bit everywhere, be able to generate a default ini which is actually the list of declared settings, be able to move to an array based settings access.
  In this logic, all settings should be clearly defined with a default and validation range (and some use comments to be added at some point). Therefore this part may not be removed. Beside, it breaks all access to them in other part of the code (with this PR SDL joystick support would be broken)

  On this, the not satisfying part are
  - the fact that we `#ifdef` regarding SDL => we should just drop it (I kept it to keep the existing behavior)
  - the fact that it seems that the default mapping causes conflict with SDL (note that this did not change either)
  
  I would propose to use a default mapping that correspond to a `no assignment` state for all platforms

- Regarding removing `RefreshSDLDevice` at handler creation, it is there to avoid the handler to be in an undefined state, not matching the event state (as joystick/gamepad added/removed events can happen before and not be reemitted). I do not really get why you want to remove it. I would expect this removal to create issues.

- Regarding using a `GamePad` layout, this means that when using SDL with joystick that are recognized as GamePads, then the only way to define mapping will be through manual edition of the `Default_gamecontrollerdb.txt` (UI will not work anymore). Beside, there is no mapping for plunger. This seems very counter intuitive for users that are used to uses the UI to define mappings. Beside, there is a risk of overwriting custom mappings when updating VPX. I expect this to create a lot of user support needs, so I would really prefer something that integrates in the UI (this is true for all platforms, as standalone doesn't have UI yet but this is planned and slowly coming into life)
  
  This also makes me realize I created a regression by not fully reproducing `m_override_default_buttons` behavior that allows to override predefined layouts of existing controllers (VirtuaPin,...) when rewriting the input section.

- You have removed a comment section of the still pending work to be done on defining additional mappings for custom actions and controller (PinMame) actions. I don't think this can be removed as this is still to be done for proper input handling.

- You have added handling of escape and enable3D through key event (`FireGenericKeyEvent`) which is wrong as these are handled through action mapping (`FireActionEvent`)

- Regarding moving to SDL input only, there is still quite some road ahead. The idea of the input rewrite is to cleanup this part of the code, allowing to more easily move toward this goal, but also allowing support for other input system which I'm not sure can be simply based upon SDL input (especially custom remote & VR controllers).

vbousquet:
So, perhaps what I would propose:
- remove #ifdef SDL in Settings
- use a default mapping value that lead to an unexisting button in Settings
- implement esc/Enable3D as you did but using FireAction
- implement missing action mappings (need to be added to action) but no urgency, keep comments until it is done
- fix the missing test on button override for existing controller input layout
- exchange and define a way to move forward for GamePad which would include UI integration

What do you think ?

Le-Syl21:
Il y a beaucoup de sujet dans ce sujet.
Ping moi sur https://discord.com/channels/652274650524418078/1076655472846831667
Et on essai de se faire une réunion discord pour voir tout les points ensemble ?

Le-Syl21:
Si tu as pas 15 minutes pour qu'on se mette d'accord sur ce qu'il faut faire, nottament la gestion gamepad par template qui doit être séparé de la gestion joystick comme le voudrais la logique SDL, je vais juste bricoler ma version dans mon fork et je te laisse gérer la branch main

Issue #2266, opened 2025-03-21
standalone: pup playlist randomize is wrong


mrandromeda69:
**Describe the bug**
In the PUPPlaylist.cpp it is stated that AlphaSort from the playlists.pup file that AlphaSort is randomize. This is correct. But then is stated that if Alpasort =0 means randomize is checked. This is incorrect, randomize is on when the value is 1. Zero means do not randomize. You can check this with puppack editor. In the function CreateFromCSV this is also wrong translated.

You can test this with any puppack with for instance a music folder with multiple numbers.
For instance in gunship: In standalone the music order is always the same. In VPX windows the order is random.

jsm174:
I’m not sure about this from my research but I will double check. Here are my notes:

https://gist.github.com/jsm174/e5aa4ebe70052b5cf2ef49ab40c35dfb#playlistspup

mrandromeda69:
You can test it in the editor. When randomize is on, everytime you fire the event a random file is played. If you turn it off everytime you fire the event the files are played in sorted order . If you look in the pup file, then randomize is saved as value 1.

francisdb:
Not sure if this is related but this table uses triggers with custom playlist playback functions that are not implemented currently:

```
2025-03-22 09:58:17.746 WARN  [1468937] [PUPTrigger::CreateFromCSV@101] CustomFunc not implemented: 23,1,music,E802,12,"{""mt"":301, ""SN"":12, ""FN"":30, ""PM"":1}",,,,,,,CustomFunc,0
2025-03-22 09:58:17.746 WARN  [1468937] [PUPTrigger::CreateFromCSV@101] CustomFunc not implemented: 26,1,Next,E800,12,"{""mt"":301, ""SN"":12, ""FN"":30, ""PM"":2}",,,,,,,CustomFunc,0
2025-03-22 09:58:17.746 WARN  [1468937] [PUPTrigger::CreateFromCSV@101] CustomFunc not implemented: 26,1,Previous,E801,12,"{""mt"":301, ""SN"":12, ""FN"":30, ""PM"":2}",,,,,,,CustomFunc,0
```

see also https://github.com/francisdb/pup-research/blob/main/README.md

jsm174:
@mrandromeda69 - Thanks again for reporting this. 

I confirmed the correct behavior with @nailbuster, and the built in Help info was incorrect:

[...]

PR #2278, opened 2025-03-30
standalone: small updates from standalone branch

jsm174:


jsm174:
also fixes https://github.com/vpinball/vpinball/issues/2266

PR #2284, opened 2025-04-01
standalone: move pinmame, flex, b2s, and pup into one window

jsm174:
**This PR is not ready to be merged so it's set to draft.**

As we've discussed in the past, Standalone has had some long standing issues with rendering external windows:

- Since we are just using `SDL_Renderer`, PinMAME and FlexDMD windows are not using any shaders. The windows are pixelated and people have been wanting nice looking DMDExt style DMDs since the beginning.

- On mobile, there's no support for multiple SDL windows. People have to go through lots of work to copy DMD pixels and use Flashers and Textboxes. And there is no support for seeing PuP and B2S windows. 

With all the latest improvements to the BGFX builds, Standalone can now start to leverage built in multiple window rendering.

Since plugins are still a ways off, the easiest way to migrate FlexDMD and PinMAME, is to use the new `ScoreView`. It is linked to the DMD settings in `VPinballX.ini`:

```
[DMD]
DMDFullScreen = 0
# 0 = Disabled, 1 = Embedded, 2 = External
DMDOutput = 2
DMDWndX = 7
DMDWndY = 343
DMDWidth = 768
DMDHeight = 192
```

I've altered `DMDWindow.cpp` to leverage all the new DMD/Plugin code, so this pretty much requires no changes to `VPinMAMEController` or `FlexDMD`.

If we proceed with this change, then the following settings would go away:

```
PinMAMEWindow = 
PinMAMEWindowX = 
PinMAMEWindowY = 
PinMAMEWindowWidth = 
PinMAMEWindowHeight = 
PinMAMEWindowRotation = 
FlexDMDWindow = 
FlexDMDWindowX = 
FlexDMDWindowY = 
FlexDMDWindowWidth = 
FlexDMDWindowHeight = 
FlexDMDWindowRotation = 
```

One issue though is calls to `Show`/`Hide` in PinMAME or FlexDMD are not supported. Users would have to turn off `DMDOutput = 0` on games that didn't have a DMD, like EM's. 

As for migrating B2S and PuP, we can still use `SDL_Renderer`, but instead of drawing directly to an `SDL_Window`, we need to render to an `SDL_Texture` and then copy the pixels into a VPX `BaseTexture`. (We could probably port Standalone's `RendererGraphics.cpp` to render in BGFX but that might be a little overkill for now.)

One issue with this approach is on Linux and Android, if `SDL_Renderer` tries to use Vulkan, and BGFX is using Vulkan, (or both are using OpenGL) then creating an `SDL_Renderer` will fail. If it fails, it defaults to a slower software render. (MacOS/iOS does not have this issue and can use multiple Metal base renderers)

The last issue with all these changes are, we would no longer be able to support the GL/GLES builds of standalone. (I believe the multi-window support is currently only working for BGFX builds.)

I know this might be tough for some users, but I'm unsure of how to progress without some tough decisions.

### DIFF
--- a/src/core/main.cpp
+++ b/src/core/main.cpp
@@ -996,7 +996,7 @@ public:
          for (int display = 0; display < displays.size(); display++) {
             const SDL_DisplayMode* displayMode = SDL_GetDesktopDisplayMode(displays.at(display).adapter);
             PLOGI << "display " << displays.at(display).adapter << ": " << displayMode->w << 'x' << displayMode->h
-                  << " (refreshRate=" << displayMode->refresh_rate << ')';
+                  << " (refreshRate=" << displayMode->refresh_rate << ", pixelDensity=" << displayMode->pixel_density << ')';
          }
 
          PLOGI << "Available external window renderers:";


jsm174:
> I would like to see a way to embed the DMD in the B2SFullDMD window but I see no such option? Preferably this positioning could be relative

That seems like a good idea.

For right now, I don't want to expand this PR beyond getting what already exists today and works. We can definitely go back and add these capabilities.

toccata10:
Relative positioning would be great (as an improvement). This is what I tried to achieve in batocera's configen. It's really a pain today when you change your monitor (typical example: when I connect my laptop on my 4k tv...) to see that all your proportions/positions are screwed up.

vbousquet:
I'm not sure what is needed here, but just a few infos on what is being worked on:
- I'm slowly porting anciliary windows to VPX (B2S/PUP, ...)
- all anciliary windows can be either a full window or embedded in another
- window spreading over multiple displays won't be really supported as this is something that no OS and library really supports (due to mismatch in displays DPI and complexity of doing window composition with different DPI)

francisdb:
> window spreading over multiple displays won't be really supported as this is something that no OS and library really supports (due to mismatch in displays DPI and complexity of doing window composition with different DPI)

Maybe I read the docs wrong but cage only supports a single Window being visible and it's spread over all displays. Anyway, if you can render windows inside other windows then the only thing needed would be placing the playfield inside some other main window. For now we can continue with forcing XWayland instead of native Wayland where windows can't be placed.

vbousquet:
I don't know much about Wayland. The thing to be handled carrefully is DPI across multiple display. Usually DPI is handled by the operating system, on WIndows it is a simple scaling mechanism, on MacOs it is more cumbersome with a mix of (legacy ?) HiDPI of Retina displays and normal DPI. The problem arise when the window spread accross multiple displays that do not have the same DPI leading to some difficulties to get consistent results.

The second thing to worry for is how you perform synchronization as the different displays will likely have different VSync timing and even (slightly) different refresh rate.

jsm174:
This PR has been adapted to use the new ancillary window system in the most recent builds of 10.8.1.

The following standalone windows have been replaced with `ScoreView`:

```
FlexDMDWindow 
PinMAMEWindow
B2SDMD
PUPDMDWindow
PUPFullDMDWindow
```

The following standalone windows have been replaced with `Backglass`:

```
B2SBackglass
PUPBackglassWindow
```

The following standalone windows have been replaced with `Topper`:

```
PUPTopperWindow
```

If multiple windows access the same ancillary window, you should disable the window that is not needed. 

For example, a pup backglass usually is preferred over a b2s backglass, so you may need to do:

```
B2SBackglass=0
PUPBackglassWindow=1
```

Some other notes:

`*FullScreen` does not really work well on MacOS. I recommend:

```
[ScoreView]
ScoreViewFullScreen = 0
; Output mode
; 0 - Disabled
; 1 - Embedded in another output
; 2 - Native system window (maybe a window, exclusive fullscreen, VR headset,...)
ScoreViewOutput = 2
```

On mobile, `*Output` should be `1` for embedded.

```
[ScoreView]
ScoreViewFullScreen = 0
; Output mode
; 0 - Disabled
; 1 - Embedded in another output
; 2 - Native system window (maybe a window, exclusive fullscreen, VR headset,...)
ScoreViewOutput = 1
```

PR #2286, opened 2025-04-02
fix: initialize m_hitEvent for Primitive

francisdb:
While working on my [vpx read/write library](https://github.com/francisdb/vpin) I found strange boolean values for HTEV

This could also use a backport to 10.8.0


vbousquet:
I fixed it more globally so closing. Thanks for the report!

francisdb:
thanks for fixing in #2287

PR #2287, opened 2025-04-02
PartGroup fixes

vbousquet:


francisdb:
Thanks for reverting that change on the field order.  :1st_place_medal:

PR #2297, opened 2025-04-04
fix(standalone): vbscript GetRef for Functions

francisdb:
This is a fix for an upcoming table using https://github.com/mpcarr/vpx-glf
Meanwhile this table was released: https://vpuniverse.com/files/file/24875-dark-chaos-apophis-2025/

```vbscript
Function Glf_3()
    Glf_3 = 15000
End Function

Dim m_value: m_value = "Glf_3"
```

```
Script.Print 'GetRef(m_value): Object'
Script.Print 'GetRef(m_value)(): Object'
Script.Print 'GetRef(m_value)()(): Object'
```

With this patch:

```
Script.Print 'GetRef(m_value): Object'
Script.Print 'GetRef(m_value)(): 15000'
```

### DIFF
--- a/standalone/inc/wine/dlls/vbscript/global.c
+++ b/standalone/inc/wine/dlls/vbscript/global.c
@@ -141,17 +141,7 @@ static HRESULT WINAPI GetRef_Invoke(IDispatch *iface, DISPID id, REFIID riid, LC
        return E_FAIL;
 
     if (id == DISPID_VALUE && (flags & DISPATCH_METHOD)) {
-       if (dp->cArgs == 0 && (flags & DISPATCH_PROPERTYGET)) {
-          IDispatch *disp = &This->IDispatch_iface;
-          IDispatch_AddRef(disp);
-          V_VT(res) = VT_DISPATCH;
-          V_DISPATCH(res) = disp;
-
-          return S_OK;
-       }
-       else {
-          return exec_script(This->ctx, FALSE, This->func, NULL, dp, res);
-       }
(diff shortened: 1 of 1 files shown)


jsm174:
Wine's VBScript engine was completely missing support for `GetRef` so I made it using bits and pieces from other objects.

I searched for `if (dp->cArgs == 0 && (flags & DISPATCH_PROPERTYGET)) {` and couldn't find it anywhere else, so I definitely had it there from a reason.

The first two tables working in standalone were Grand Tour and T2, and I remember nothing worked without `GetRef`.

If I use this patch on T2, it won't start:

```
2025-04-07 08:12:50.226 ERROR [24409499] [CodeViewer::OnScriptError@1213] Script Error at line 252 : Description unavailable
2025-04-07 08:12:50.226 ERROR [24409499] [CodeViewer::AppendLastErrorTextW@4037] Runtime error: Line: 252, Character: 2, Description unavailable
```

The code is:

```
                MECHGUN.AddSw 33, 0, 5          'Gun Home
                MECHGUN.AddSw 32, 98, 105       'Gun Mark
                MECHGUN.Callback=GetRef("UpdateGun")
                MECHGUN.Start
```

francisdb:
That might be another case of https://bugs.winehq.org/show_bug.cgi?id=53844

francisdb:
@jsm174 do I understand it correctly that the current workaround is `GetRef(m_value)(null)` and that that in above case returns `15000`?

PR #2322, opened 2025-04-12
standalone: fix B2S PicBoxCollection null map check (@superhac)

jsm174:


francisdb:
Does this fix something visual?

superhac:
Confirmed the fix works!  Thanks!


> Does this fix something visual?
I'll have to let @jsm174 as I'm not sure.  But the table Goin' Nuts with [this](https://vpuniverse.com/files/file/23079-goin-nuts-gottlieb-1983-3-screen-backglass/) b2s would segfault.

PR #2331, opened 2025-04-15
Move table back/forth/left/right In VR

Panasony1008:
The table can be moved back and forth and left and right to make it easier to align with the pin controller when using VR. No default key bindings are set to avoid conflicts with other key bindings. Users set key bindings by adding the following to VPinballX.ini: 
TableForwardKey = <num>
TableBackKey = <num>
TableLeftKey = <num>
TableRightKey = <num>

### DIFF
--- a/pininput.cpp
+++ b/pininput.cpp
@@ -1074,6 +1074,14 @@ void PinInput::FireKeyEvent(const int dispid, int keycode)
       g_pplayer->m_pin3d.m_pd3dPrimaryDevice->tableUp();
    else if (keycode == g_pplayer->m_rgKeys[eTableDown] && dispid == DISPID_GameEvents_KeyUp)
       g_pplayer->m_pin3d.m_pd3dPrimaryDevice->tableDown();
+   else if (keycode == g_pplayer->m_rgKeys[eTableForward] && dispid == DISPID_GameEvents_KeyUp)
+      g_pplayer->m_pin3d.m_pd3dPrimaryDevice->tableForward();
+   else if (keycode == g_pplayer->m_rgKeys[eTableBack] && dispid == DISPID_GameEvents_KeyUp)
+      g_pplayer->m_pin3d.m_pd3dPrimaryDevice->tableBack();
+   else if (keycode == g_pplayer->m_rgKeys[eTableRight] && dispid == DISPID_GameEvents_KeyUp)
+      g_pplayer->m_pin3d.m_pd3dPrimaryDevice->tableRight();
+   else if (keycode == g_pplayer->m_rgKeys[eTableLeft] && dispid == DISPID_GameEvents_KeyUp)
+      g_pplayer->m_pin3d.m_pd3dPrimaryDevice->tableLeft();
    else
 #endif
 
--- a/pininput.cpp
+++ b/pininput.cpp
@@ -104,7 +104,11 @@ PinInput::PinInput()
    m_joytablerecenter = 0;
    m_joytableup = 0;
    m_joytabledown = 0;
-
+   m_joytableforward = 0;
+   m_joytableback = 0;
+   m_joytableleft = 0;
+   m_joytableright = 0;
+ 
    m_firedautostart = 0;
 
    m_pressed_start = false;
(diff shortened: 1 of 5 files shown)


vbousquet:
Hi ! I fully share the interest for being able to more easily adjust scene offset. This is something being worked on in the 10.8.1 branch. 

One thing that I would avoid is to add keyboard bindings for these as I think this comes with a few difficulties:
- this adds UX complexity as they need defining (in a ini file or somewhere deep in huge settings page), remembering, and using without clean feedback
- this require physcial button and pincab can not hold as many buttons as we would want
- and if we consider the keyboard binding as a limited ressource, this consume this ressource while other solutions could be used instead.

Following these, I moved all VR UI to the TweakUI (and reworked the TweakUI to work without the current hacks we have in 10.8). I think this need to follow the same path, but perhaps the easiest solution would be to test what is being developped in 10.8.1 and give some feedback (If you want, you can have a try by downloading latest builds of the BGFX master branch. It should work with any headset supporting OpenXR) ?

Panasony1008:
I tried the latest version of BGFX.
It seems to have a number of issues, such as always crashing when closing the table, but I was able to confirm that TweakUI works.
In my opinion, this menu is good, but I also felt that real-time operation using the current key bindings was necessary.
Since I position the table frequently, I felt it was cumbersome to do it every time with TweakUI.

vbousquet:
Did you encounter the crash with the latest from master ? Could you be more precise on the version and share a bug report for this (using the template to have more infos) to allow debugging as this crash does not occur on my end and other testers ?

The "5" is replaced by the reset button (usually Enter) and allows continuous recentering (just let him press) with live feedback. This makes things more easy when playing in XR where you need to precisely match the virtual table with the real world pincab.

Note that normal view recentering, the one provided by the headset (for meta long press on Meta key) is also supported.

gjupp:
Panasony's fork with the ability to move the table position more easily has been very useful so far, it would be great to have it incorporated into the official version. Secondly, it would be ideal if a key repeat for the movement were a possibility.

The1Dijk:
Good to have this, but please keep the 8, 5 and 2 keys functional. TweakUI works but you always need to go to a menu and in VR you just want to change quickly with a key press. 5 being the most important key.

rdestevao:
Also love the possibility to move the table with keybindings. I even dedicated keys for this in my self made controller! Would be a big loss to lose that.
Regarding the keybindings being a limited resource, would not be possible to use a combination of keys?
Example, keep magna save right pressed (or shift, or coin button, ...) while using the corresponding movement keys would move the table. Or, as an example, use the already existing up and down movement keys would move the table up and down, but if right magna save was being pressed it would move the table forward and backward, but if it was the left magnasave pressed would go left and right... Basically, having a second layer could also open other doors for limited keybindings.

Jglagojoh:
I say leave the key bindings to 2 (up), 8 (down) and 5 (reset to look position) for now. This would allow more testing for VR users to weigh in.

vbousquet:
> In my case, I don't like the table height being reset.
> You have to adjust the height every time you reset it.

Not sure which version you tested it, but in the latest, it does not recenter the height (it just keep it untouched).
It will reset it only if it is far away (high z offset).

OpenVR (previous) did not have a clean room scale space reference, so height would always need to be adjusted. I'm trying to change that as this is anyway needed for true mixed reality. Now, in OpenXR:
- it uses a room space reference (so the real space reference, which can be resetted by doing a long press on the Meta button for Meta Quest, leading to real ground position)
- uses the user defined cabinet real world size
- the table defined glass position to locate the lockbar

With this, the z offset is now exactly that: a slight offset (as so far I can't get it to perfectly match as is). Adjusting the height should be less needed using the OpenXR runtime. Ideally, it should be possible to have something that 'just' works, that is to say gives a stable reference but this is difficult to achieve (especially in PCVR, it seems that native VR has a more stable space reference).

Regarding adding input, I think what is really needed is to finish the input rewrite with a generic action mapping system. I started it (and broke XInput as it seems) but did not have time to finish. With a generic action mapping (in the code, UI, LiveUI), we could add whatever mapping without degrading overall code/app.

Panasony1008:
Consider use in VR as well as MR.
Some people will play while sitting on the floor.
Other times they might play at a computer desk.
Some people might want to play while looking down from above.

vbousquet:
> Consider use in VR as well as MR.

I do have that in mind. This needs to be known at runtime as the headset does not perform space localization the same depending on that (that's way most games ask if you are sitting or standing at the beginning). It's just not yet implemented and has to be considered with the other part of the VR input redesign (the OpenXR implementation can switch from room reference space to standing reference space).

KrakenByteKraken:
I'm very interested in this one, it's a good point.

One little question: Does somebody know if "HP Reverb G2" (WMR) is compatible with this VPX BGFX OpenXR version?

I only could make it works fine with the OpenVR (aka SteamVR) flavor, aka VPX OpenGL so: Am I losing something?

Thanks in advance!

Issue #2339, opened 2025-04-16
PlaySound not looping correctly


apophis79:
**Describe the bug**
When using PlaySound() with the loops argument set to -1, the sound does not loop at all, rather it just plays once.

**To Reproduce**
1. Open VPX 10.8.1 Beta Rev 3309
2. Open a new "Full Example Table"
3. Add the following near the top of the script:
       PlaySound "ballrelease", -1
4. Run the table 

**Expected behavior**
The ballrelease sound effect should endlessly repeat. But it only plays once. 

**Versions & Settings**
 - VPX: 10.8.1 Beta Rev 3309
 - Windows version 11

PR #2340, opened 2025-04-16
Fix for issue 2339 where PinSound is not respecting the loopcount.

superhac:
This is resoles issue [2339](https://github.com/vpinball/vpinball/issues/2339).  Tested with the following code in the example table:

```
Sub Plunger_Init()
        PlaySound SoundFX("ballrelease",DOFContactors), 0,1,AudioPan(BallRelease),0.25,0,0,1,AudioFade(BallRelease)
        'Plunger.CreateBall
        BallRelease.CreateBall
        BallRelease.Kick 90, 7
        BIP = BIP + 1
        PlaySound "ballrelease", -1
End Sub
```

This continues to loop until the ball drains.

### DIFF
--- a/src/audio/pinsound.cpp
+++ b/src/audio/pinsound.cpp
@@ -397,11 +397,11 @@ void PinSound::PlayBGSound(float volume, const int loopcount, const bool usesame
          Mix_HaltChannel(m_assignedChannel);
       Mix_Volume(m_assignedChannel, nVolume);
       if (restart || !usesame) // stop and reload
-         Mix_PlayChannel(m_assignedChannel, m_pMixChunkOrg, 0);
+         Mix_PlayChannel(m_assignedChannel, m_pMixChunkOrg, loopcount);
    }
    else { // not playing
       Mix_Volume(m_assignedChannel, nVolume);
-      Mix_PlayChannel(m_assignedChannel, m_pMixChunkOrg, 0);
+      Mix_PlayChannel(m_assignedChannel, m_pMixChunkOrg, loopcount);
    }
 }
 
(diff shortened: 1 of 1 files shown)


francisdb:
I have the impression on EM tables tat the flippers sound twice now.
On 4 roses flippers and coin drops repeat twice. (off by 1 error?)
And an other odd thing is that on Speakeasy (2 or 4) the flipper sounds repeat even more.
So there probably still is an issue.

superhac:
I don't have either of those tables but I tried Strikes and Spares.  I had the same issue as you pointed out above.  Here are the sound logs from a single fliper press(holding it):

```
2025-04-17 13:16:01.070 INFO  [263889] [PinSound::Play_SNDCFG_SND3D2CH@551] Playing Sound: fx_Flipperup SoundOut (0=table, 1=bg): 0 nVol: 1.0000000 pan: 0.0000000 Pitch: 0 Random pitch: 0.0000000 front_rear_fade: 0.0000000 loopcount: 1 usesame: 0 Restart? 1
```

Loop is set to `1`.  Which means it plays and loops once.  it should be `0` if they don't want to loop at all?

superhac:
I think I figured it out.  With the old BASS implementation, BASS used a LOOP flag which was either 0 (no loop) or 1 (keep looping until otherwise).  In Mixer its a hard value of the number of times to actually loop AFTER it initially plays with the exception of -1 which means play indefinitely.   I will put in the fix today.

francisdb:
As for the speakeasy tables:
https://vpuniverse.com/files/file/8675-speakeasy-2-player-bally-1982/

```vbscript
PlaySound SoundFX("fx_flipperdown",DOFContactors), VolFlip
```
Where the signature is:
```
PlaySound(string, int loopcount, float volume, float pan, float randompitch, int pitch, bool useexisting, bool restart, float front_rear_fade)
```


So these pass a volume of 5 as the loop count...

Table bug

Issue #1944, opened 2024-10-10
standalone: ultradmd assets not loaded case-insensitive


francisdb:
**Describe the bug**

Pup pack files are loaded case-insensitive (and logging a warning)

This does not seem to work for UltraDMD calls

```vbscript
         DMD "dmd baseBlank.png", " ", "",  3000
         DMDFLush
         DMD "dmd baseBlank.png", "SPECIAL KEY", "PRESSED 50.000" , 200
```

yields

```
2024-10-10 16:58:55.741 WARN  [383885] [AssetManager::Open@258] Asset not loaded: /home/me/vpinball/tables/Spooky Wednesday (Original 2024)/./SpookyWednesday.UltraDMD/dmd baseBlank.png
2024-10-10 16:58:55.741 WARN  [383885] [AssetManager::Open@258] Asset not loaded: /home/me/vpinball/tables/Spooky Wednesday (Original 2024)/./SpookyWednesday.UltraDMD/dmd baseBlank.png
```

And the dmd stop working

vpinball 10.8.1 branch
table: https://vpuniverse.com/files/file/18514-spooky-wednesday-vpx-original-2024-table-and-puppack/ (requires patching)

surtarso:
For the Diablo (JP) table I had to rename the Diablo.UltraDMD to diablo.UltraDMD or the assets would not load at all.

PR #2347, opened 2025-04-18
Standalone: Added case sensitive path hunting in the AssetManager Class targeting AssetSrcType_FlexResource and AssetSrcType_File types only.

superhac:
Fixes tables which use flexdmd and ultradmd where the asset names are case sensitive. Fixes issue [1944](https://github.com/vpinball/vpinball/issues/1944)

Mark as draft till testing and discussion is done.


francisdb:
Why was this closed?

superhac:
@jsm174 implemented the same functionally at a higher level.  I think he said it also covers pups too now.  This will be in his next push.

PR #2357, opened 2025-04-22
fix(pup): only join thread if running

francisdb:
When a pup pack is available but the plugin is disabled:

```ini
[Standalone]
PUPPlugin = 0
```

The `PupScreen` thread is never started. Joining it was causing a crash.

```
terminate called after throwing an instance of 'std::system_error'
  what():  Invalid argument
```


jsm174:
Good catch. `m_isRunning` not being defaulted could definitely cause issues on Android and Linux!

PR #2374, opened 2025-04-29
Improve window creation error handling

francisdb:
I had a typo in the SDL_RENDER_DRIVER env var

```
INFO  [812346] [VP::WindowManager::Start@105] Window manager start
ERROR [812346] [VP::Window::Init@72] Failed to initialize window: title=B2SBackglass
```

With this applied:

```
INFO  [822761] [VP::WindowManager::Start@105] Window manager start
ERROR [822761] [VP::Window::Init@53] Failed to create renderer for window: title=B2SBackglass vulkam not available
```

### DIFF
--- a/standalone/inc/common/Window.cpp
+++ b/standalone/inc/common/Window.cpp
@@ -39,39 +39,41 @@ bool Window::Init()
    SDL_SetNumberProperty(props, SDL_PROP_WINDOW_CREATE_FLAGS_NUMBER, flags);
    m_pWindow = SDL_CreateWindowWithProperties(props);
    SDL_DestroyProperties(props);
-   if (m_pWindow) {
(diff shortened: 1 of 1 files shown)


jsm174:
fwiw, most off this will go away when/if we finally merge https://github.com/vpinball/vpinball/pull/2284. (It will only use a software renderer. The other ones don't buy anything in b2s or pup)

> I do wonder if vpinball should not be shut down in this case?

I would rather keep it to not exit vpx if the renderer fails. 

As for the PR, I think the text should be a little more clear. ie instead of:

```
PLOGE.printf("Failed to create window: title=%s %s", m_szTitle.c_str(), SDL_GetError());
return false;
```

maybe

```
PLOGE.printf("Failed to create window: title=%s, error=%s", m_szTitle.c_str(), SDL_GetError());
return false;
 ```

Issue #2057, opened 2024-12-12
Menus not working correctly in BGFX on MacOS for some time


cwscws:
The menus (via ESC or F12) are not working correctly in BGFX on MacOS for some time. Mouse clicks or hovering above menu options to highlight them are not recognized. Key input also not recognized properly most of the time. So LiveUI not usable at all, access to table options also not working because of the not recognized input. 

The standalone 10.8.0 menus are working perfectly.

francisdb:
I can confirm this on Linux. I have the impression it has something to do with window focus. The Backglass window seems to take focus instead of the playfield.

![image](https://github.com/user-attachments/assets/c7a60157-b9d4-4019-a1bb-6e6fe35c3c61)

Seems that this only causing issues for the main `ESC` menu. Pressing `F12` or POV settings work correctly when the backglass is in focus.

It has been like this for months now.

francisdb:
I think the mouse clicks not being registered is macOS only.

jsm174:
I spent some time debugging this, and I think there are few things going on.

1) I think `ImGui_ImplSDL3_ProcessEvent` is really supposed to process events before the SDL loop.

```
// You can read the io.WantCaptureMouse, io.WantCaptureKeyboard flags to tell if dear imgui wants to use your inputs.
// - When io.WantCaptureMouse is true, do not dispatch mouse input data to your main application, or clear/overwrite your copy of the mouse data.
// - When io.WantCaptureKeyboard is true, do not dispatch keyboard input data to your main application, or clear/overwrite your copy of the keyboard data.
// Generally you may always pass all inputs to dear imgui, and hide them from your application based on those two flags.
// If you have multiple SDL events and some of them are not meant to be used by dear imgui, you may need to filter events based on their windowID field.
```

2) `ImGui_ImplSDL3_ProcessEvent` is called only if the window is the playfield window. If you click the mouse button down inside the playfield window and the release it outside the window, imgui never gets the mouse up message. (Seeing this behavior on the mac)

3) `imgui_impl_sdl3` isn't really accounting for retina scaling when it comes to mouse locations.

```
void ImGui_ImplSDL3_NewFrame()
{
    ImGui_ImplSDL3_Data* bd = ImGui_ImplSDL3_GetBackendData();
    IM_ASSERT(bd != nullptr && "Context or backend not initialized! Did you call ImGui_ImplSDL3_Init()?");
    ImGuiIO& io = ImGui::GetIO();

[...]

PR #2381, opened 2025-05-02
renderer: add m_drawableWidth/Height to better support high dpi in SDL3

jsm174:
SDL3 has much better support for HighDPI. 

The HighDPI docs are [here](https://wiki.libsdl.org/SDL3/README/highdpi)

One issue on Mac has always been the reported window size vs the drawable size. On MacOS, we are over-writing the window's internal `m_width` and `m_height` with the drawable size. This is confusing especially when you really need to know the real `m_width` and `m_height`. 

This PR adds a new `m_drawableHeight` and `m_drawableWidth` to the Window. Now we can always get the real `m_width` and `m_height` where-ever we need to.

This may introduce some regressions, but I hope it doesn't.

Also, the imgui mouse is still offset (see https://github.com/vpinball/vpinball/issues/2057#issuecomment-2649567181), so hopefully these changes can help get us closer to working.

### DIFF
--- a/src/core/vpinball.cpp
+++ b/src/core/vpinball.cpp
@@ -1134,8 +1134,8 @@ void VPinball::DoPlay(const int playMode)
                if (isPFWnd) {
                   // We scale motion data since SDL expects DPI scaled points coordinates on Apple device, while it uses pixel coordinates on other devices (see SDL_WINDOWS_DPI_SCALING)
                   // For the time being, VPX always uses pixel coordinates, using setup obtained at window creation time.
-                  e.motion.x *= g_pplayer->m_playfieldWnd->GetHiDPIScale();
-                  e.motion.y *= g_pplayer->m_playfieldWnd->GetHiDPIScale();
+                  e.motion.x *= SDL_GetWindowPixelDensity(g_pplayer->m_playfieldWnd->GetCore());
+                  e.motion.y *= SDL_GetWindowPixelDensity(g_pplayer->m_playfieldWnd->GetCore());
                   static float m_lastcursorx = FLT_MAX, m_lastcursory = FLT_MAX;
                   if (m_lastcursorx != e.motion.x || m_lastcursory != e.motion.y)
                   {
@@ -1148,16 +1148,19 @@ void VPinball::DoPlay(const int playMode)
                {
                   // Handle dragging of auxiliary windows
                   SDL_Window *sdlWnd = SDL_GetWindowFromID(e.motion.windowID);
-                  VPX::Window *windows[] = { g_pplayer->m_scoreviewOutput.GetWindow(), g_pplayer->m_backglassOutput.GetWindow() };
-                  for (size_t i = 0; i < sizeof(windows) / sizeof(VPX::Window *); i++)
+                  std::vector<VPX::Window*> windows = {
+                     g_pplayer->m_scoreviewOutput.GetWindow(),
+                     g_pplayer->m_backglassOutput.GetWindow()
+                  };
+                  for (VPX::Window* wnd : windows)
                   {
-                     if (windows[i] && sdlWnd == windows[i]->GetCore())
+                     if (wnd && sdlWnd == wnd->GetCore())
                      {
                         int x, y;
-                        windows[i]->GetPos(x, y);
+                        wnd->GetPos(x, y);
                         Vertex2D click(x + e.motion.x, y + e.motion.y);
                         if (dragging > 1)
-                           windows[i]->SetPos(static_cast<int>(x + click.x - dragStart.x), static_cast<int>(y + click.y - dragStart.y));
+                           wnd->SetPos(static_cast<int>(x + click.x - dragStart.x), static_cast<int>(y + click.y - dragStart.y));
                         dragStart = click;
                         dragging = 2;
                         break;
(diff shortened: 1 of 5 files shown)


vbousquet:
The change you propose can have large side effects as these breaks the contract that width/height are in pixel units not logical units. If possible, I think I would avoid that as this may be difficult and long to track all effects and stabilize. Beside, it will likely not solve issue with SDL input which has a hack for Retina display (see below)

Initially, widths and heights of outputs did not have a clear unit. I tried to solve the issues we had by defining width and height to always be in pixel units (and not logical units, that is to say DPI scaled). The Window class was created afterward to cleanup and (solve other issues). Apple devices were added afterward and we discovered the strange (hacky ?) retina + logical units of iOS/MacOS that was solved with a hack as SDL2 did not have clean things neither for it.

When we moved to SDL3, the core of this was not reviewed. This was a fast port and would need some more love. Here is what I propose:

First, I think we should still stick to clear definitions, with comments in `Window.h`:
- width/height should be the size in **pixel** units (to avoid breaking existing code that expect that, and also because DX9 path is done like that)
- we could add size in **logical** units, integrating Retina display support the way SDL3 did
For example, we would end up with m_width (or m_pixelWidth if we want to rename to be fully clear, which is the size for rendering) and m_logicalWidth (size for user interaction). With this all rendering would use pixel sizes, all positionning, sizing, UI would use logical sizes.

Regarding video mode selection, we should improve what we have here for MacOS: https://github.com/vpinball/vpinball/blob/670b001528ba0b1d651cbf27fbf1bc4745b24f00/src/renderer/Window.cpp#L571 as we need to present the user with an unscaled mode (what we do on Windows, but is it ok on MacOS ?) but use the scaled mode for pixel units (what we do on Windows as the pixel density added in SDL3 is always 1 for Windows, but will likely be 2 on retina display and we are not storing it nor presenting it to the user)

Regarding user interaction, what we do is very hacky (again due to history and SDL2 not having full DPI/Retina support) as we move back everything in pixel units instead of logical units (here: https://github.com/vpinball/vpinball/blob/670b001528ba0b1d651cbf27fbf1bc4745b24f00/src/ui/LiveUI.cpp#L1358 this should be removed). The right implementation (that we ovewrite) is defined in the SDL3 ImgUI implementation:
```
SDL_GetWindowSizeInPixels(bd->Window, &display_w, &display_h);
io.DisplaySize = ImVec2((float)w, (float)h);
if (w > 0 && h > 0)
    io.DisplayFramebufferScale = ImVec2((float)display_w / w, (float)display_h / h);
``` 
But this also means that all UI scaling need to be rechecked in logical units (everything is in LiveUI, shouldn't be complicated though, and easy to test).

jsm174:
I addressed the very first comment:

> For example, we would end up with m_width (or m_pixelWidth if we want to rename to be fully clear, which is the size for rendering) and m_logicalWidth (size for user interaction). With this all rendering would use pixel sizes, all positionning, sizing, UI would use logical sizes.

here: https://github.com/jsm174/vpinball/commit/9878b87331e80442fa10f6cb917906bc5f00349d

i does nothing except rename `m_width` and `m_height` to `m_pixelWidth` and `m_pixelHeight` for clarity.

If this is okay, I'll make a PR and merge it.

vbousquet:
I may be wrong but as I understand SDL3 doc, the name of the variable does not correspond to their content:
- m_drawableWidth/Height is initialized from SDL_GetWindowSizeInPixels, which is the real pixelWidth/Height (which units is indeed pixels, and must be used for all rendering)
- m_width/m_height (renamed in your branch to pixelWidth) is the window creation size, corresponding to SDL_GetWindowSize, which is the logical size of the window. Units of this are not pixels on MacOS, are platform dependent, only meaningful for presenting to the user and positioning on screen.

My idea was to rename the other way: pixelWidth for things in pixels, logicalWidth for things without unit (on Mac only, thanks to Apple's weird design). Then, when redering, everything is expressed in pixels, with UX content that should be scaled by SDL_GetWindowDisplayScale **multiplied** by pixelWidth/windowWidth.

jsm174:
> m_drawableWidth/Height is initialized from SDL_GetWindowSizeInPixels, which is the real pixelWidth/Height (which units is indeed pixels, and must be used for all rendering)

Sorry I keep get confused, but isn't this exactly what I suggested in the initial PR?  

I think windows has "just" worked because width/height is the same regardless of HD: 

Screenshot 2025-05-03 at 6 44 04 PM

I could rename `m_drawableWidth`, `m_drawableHeight` to `m_pixelWidth`, and `m_pixelHeight` if that makes it more clear?

vbousquet:
Looks good to me like this.

jsm174:
Thanks.

Okay, so I started looking into the UI code.

```
  #if defined(ENABLE_SDL_VIDEO)
   ImGui_ImplSDL3_NewFrame();
   #else
   ImGui_ImplWin32_NewFrame();
   #endif

   ImGuiIO &io = ImGui::GetIO();
   //io.DisplaySize = ImVec2(static_cast<float>(width), static_cast<float>(height)); // The render size may not match the window size used by ImGui_ImplWin32_NewFrame (for example for VR)
   //io.DisplayFramebufferScale = ImVec2(1.f, 1.f); // Retina display scaling is already applied since we override the value fom NewFrame with the rt size 
``` 

`ImGui_ImplSDL3_NewFrame` is doing all the work to set up `io`, so if I comment out changes to `DisplaySize` and `DisplayFramebufferScale` it does seem to render properly (minus rotation changes) if I change the `m_dpi` to 1.

vbousquet:
Yes, this part was a hack from the SDL2 days. Still it needs quite a bit of testing as the UI ends up in a wide variety of situations (including VR, where I can do the test when you have something set up). Also the UI DPI should be ensured to be the right one (I guess using SDL_GetWindowDisplayScale)

PR #2385, opened 2025-05-04
liveui: get window size and framebuffer scale from imgui new frame

jsm174:
This is a follow up to https://github.com/vpinball/vpinball/pull/2381#issuecomment-2849292093.

The LiveUI will get it's initial dimensions and scale from the imgui new frame.

I tested retina and non retina on MacOS and it seems to work fine.

> Yes, this part was a hack from the SDL2 days. Still it needs quite a bit of testing as the UI ends up in a wide variety of situations (including VR, where I can do the test when you have something set up). Also the UI DPI should be ensured to be the right one (I guess using SDL_GetWindowDisplayScale)

As you mentioned I'm not sure about VR, and an `m_dpi` of 1 seems to work fine for both cases.

### DIFF
--- a/src/ui/LiveUI.cpp
+++ b/src/ui/LiveUI.cpp
@@ -670,7 +670,10 @@ LiveUI::LiveUI(RenderDevice *const rd)
    else
    {
       // Use display DPI setting
-      m_dpi = SDL_GetWindowDisplayScale(m_player->m_playfieldWnd->GetCore());
+      // On macOS/iOS, keep m_dpi at 1.0f. ImGui_ImplSDL3_NewFrame applies a 2.0f DisplayFramebufferScale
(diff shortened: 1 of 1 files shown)


vbousquet:
m_dpi of 1 breaks content scaling (that is to say when user set a 150% DPI, the 150% is ignored). m_dpi must be equal to SDL_GetWindowDisplayScale to get correct content scaling (as it is in current builds).

I did test on Windows and can confirm that this PR would break LiveUI scaling. I did not test on VR though (but as it is untouched, it should be good).

jsm174:
This screenshot is when `m_dpi` is set to 2 on macos:

Screenshot 2025-05-04 at 7 21 28 PM

To get the `m_dpi` more around 1.0 we can use:

`m_dpi = SDL_GetWindowDisplayScale(m_player->m_playfieldWnd->GetCore()) / SDL_GetWindowPixelDensity(m_player->m_playfieldWnd->GetCore());`

https://wiki.libsdl.org/SDL3/README/highdpi

code has been updated

Issue #2157, opened 2025-02-13
standalone/linux: bgfx init segfaults on wayland


francisdb:
linux
wayland
bgfx
master branch

```
> SDL_VIDEODRIVER=wayland ./build/VPinballX_BGFX -play ~/Downloads/Volkan\ Steel\ and\ Metal\ \(Original\ 2023\)v1.3.vpx
...
2025-02-13 10:29:22.370 INFO  [7706] [ProgressDialog::SetProgress@220] Initializing Visuals... 10%
2025-02-13 10:29:22.371 INFO  [7706] [Player::Player@472] Initializing renderer (global states & resources)
2025-02-13 10:29:22.372 INFO  [7706] [RenderDevice::RenderDevice@653] Using graphics backend: Vulkan (available: Noop, Direct3D11, Direct3D12, OpenGL, Vulkan)
[1]    7706 segmentation fault (core dumped)  SDL_VIDEODRIVER=wayland ./build/VPinballX_BGFX -play 
```

```
XGetXCBConnection 0x0000ffffec0c06b8
RenderDevice::RenderThread RenderDevice.cpp:356
```

![Image](https://github.com/user-attachments/assets/59945e39-58f3-469a-8d0e-355f1f3c19b6)

francisdb:
Just tried Ubuntu 25.04 Beta and vpinball now defaults to wayland. Workaround is using `SDL_VIDEODRIVER=x11`.

francisdb:
Still need to check but this helped for graphics issues on some other project:

> A while ago I switched from a Nvidia to AMD graphics card. I had quite some old nvidia packages installed. Cleaning those up fixed all issues for me. 🎉

```
apt list --installed | grep nvidia
```

Le-Syl21:
sudo apt purge -y *nvidia*
sudo apt autoremove

francisdb:
Nope, that's not it. Still fails on Wayland.

jsm174:
So we use bgfx.cmake to build bgfx. Because of this, sometimes the defines aren't exactly what you would think.

```
if(BGFX_WITH_WAYLAND)
	target_compile_definitions(bgfx PRIVATE "WL_EGL_PLATFORM=1")
	target_link_libraries(bgfx PRIVATE wayland-egl)
endif()
```

https://github.com/bkaradzic/bgfx.cmake/blob/36a0d1d8d040113a954f99b1456121072dc77348/cmake/bgfx/bgfx.cmake#L88-L91

[...]

PR #2386, opened 2025-05-04
Standalone: bgfx init segfaults on wayland #2157

superhac:
This is a fix related to [2157](https://github.com/vpinball/vpinball/issues/2157).  Currently this stops the segfault but the pinmame window gets rendered behind the b2s.  In other words, it doesn't seem to be respecting the cords set for where the pinmame window should reside.   Not sure what the problem is.

### DIFF
--- a/platforms/linux-x64/external.sh
+++ b/platforms/linux-x64/external.sh
@@ -165,7 +165,6 @@ if [ "${BGFX_EXPECTED_SHA}" != "${BGFX_FOUND_SHA}" ]; then
       -DBGFX_BUILD_EXAMPLES=OFF \
       -DBGFX_CONFIG_MULTITHREADED=ON \
       -DBGFX_CONFIG_MAX_FRAME_BUFFERS=256 \
-      -DBGFX_WITH_WAYLAND=ON \
       -DCMAKE_BUILD_TYPE=${BUILD_TYPE} \
       -B build
    cmake --build build -- -j${NUM_PROCS}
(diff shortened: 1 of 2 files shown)


francisdb:
Wayland has no window positioning protocol so windows will appear wherever wayland decides to put them.
There's a limited protocol merge request that won't help us.
https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/264

For desktop use this is of course a great improvement no longer requiring X11 with env vars

platforms/linux-x64/external.sh
```diff
@@ -165,6 +165,7 @@ if [ "${BGFX_EXPECTED_SHA}" != "${BGFX_FOUND_SHA}" ]; then
       -DBGFX_BUILD_EXAMPLES=OFF \
       -DBGFX_CONFIG_MULTITHREADED=ON \
       -DBGFX_CONFIG_MAX_FRAME_BUFFERS=256 \
+      -DBGFX_WITH_WAYLAND=ON \
```
francisdb, line 168: I don't think this is required as this is set automatically on linux is my understanding of below linked line is correct.

see https://github.com/bkaradzic/bgfx.cmake/blob/36a0d1d8d040113a954f99b1456121072dc77348/CMakeLists.txt#L46

francisdb (approved):
I can confirm this fixes the crash on Wayland. There are other issues like window sizes not being respected but let's keep those for later.

jsm174 (commented):
I don’t think you need to create a whole new structure. Can’t you just set the type directly?

superhac:
> I don’t think you need to create a whole new structure. Can’t you just set the type directly?

Your right.  I can access it through bgfx::Init struct.  Fixed.   I also removed the compile option for wayland in the bgfx section of the `external.sh`.

superhac:
So at least on mine in cab mode, the table and BS2 screen are on the right monitor.  But the pinmame window is behind the B2S for some reason.

PR #2395, opened 2025-05-06
standalone: disable ball trails on mobile. bump libdmdutil

jsm174:


francisdb:
Why specifically on mobile?

jsm174:
"disable" was a bad choice of wording.

The mobile apps have their own native section of code to "default" values in the ini upon reset or app installation.

PR #2396, opened 2025-05-06
fix: PuP memory leak

### LINKED ISSUE #2388: PuP pack memory leak
vpinball BGFX
Linux
master Branch

I was testing 
https://vpuniverse.com/files/file/15565-transformers-pro-stern-2011/

with pup pack
https://vpuniverse.com/files/file/6835-transformers-stern-pup-pack/

And the further I got in the game the more sound stuttering I was getting. My system resources widget looked like this where vpinball was using 40GB of memory. The drop is where I ended vpinball.
 
Image

All you have to to is starting the table, not even starting a game and you will see it climb at a constant rate.

I don't see this when the pup pack is disabled (renamed pupvideos folder)

_After some more testing I think the sound stuttering might even not be fully related as starting a game with memory already at 40GB does not cause stuttering. Only after playing a while stuttering appears._

francisdb:
* a screen that is off should not handle events
* the media player should not have an unbound queue

fixes #2388

I'm keeping this to a minimum as this code is currently being moved to a plugin

### DIFF
--- a/standalone/inc/pup/PUPMediaPlayer.cpp
+++ b/standalone/inc/pup/PUPMediaPlayer.cpp
@@ -17,6 +17,8 @@
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 #endif
 
+constexpr size_t MAX_BUFFERED_FRAMES = 60;
+
 PUPMediaPlayer::PUPMediaPlayer()
 {
    m_loop = false;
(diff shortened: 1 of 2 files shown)


jsm174:
I had started this work in the embedded-window branch but took a different direction. I think this should work though until we migrate to the plugin. 

Maybe we should get confirmation if events are discarded for screens that are off. ie, could an event or one of those special messages change the mode to turn it back on? @nailbuster any thoughts on this?

francisdb:
After a quick search I could not find any mode changes that were indirect through messages.

nailbuster:
> Maybe we should get confirmation if events are discarded for screens that are off. ie, could an event or one of those special JSON messages change the mode to turn it back on? @nailbuster any thoughts on this?

you can ignore triggers for disabled/off pup-screens.

PR #2399, opened 2025-05-07
liveui: make keyboard input fps independent

### LINKED ISSUE #2375: Tweak menu ajust speed depends on framerate
Standalone
BGFX
master branch

**Describe the bug**

Tweak menu adjust speed depends on the frame rate

Example for `vertical offset`
Testing with a very light table that allows high framerates
https://www.vpforums.org/index.php?app=downloads&showfile=19021

- By default (60fps) a short tap gives me a 0.2cm adjustment
- With `SyncMode` set to `0` (900fps) a short tap gives me a 2-3cm adjustment

This is probably the base issue for previous issues #1583 and #1588

francisdb:
fixes #2375 

* makes the steps dependent on the time between key handler calls
* makes the volume control less touchy by using a float to keep track of when a step should be taken

If required some values can be tweaked to inc/dec faster, speed up/down faster, set max speed but now at least it's the same for everybody.

tested on 60 and 1000fps

### DIFF
--- a/src/ui/LiveUI.cpp
+++ b/src/ui/LiveUI.cpp
@@ -1721,6 +1721,11 @@ void LiveUI::UpdateTweakPage()
 
 void LiveUI::HandleTweakInput()
 {
+   static U32 lastHandle = msec();
+   const U32 now = msec();
+   const U32 sinceLastInputHandleMs = now - lastHandle;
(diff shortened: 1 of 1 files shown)


toxieainc:
Thanks!

Issue #814, opened 2023-09-25
Extractvbs is slow


francisdb:
**Describe the enhancement**

The `extractvbs` speed could be improved by not loading the whole table

vpinball: 6.612s total
[vpxtool](https://github.com/francisdb/vpxtool): 0.091s total

**To Reproduce**

vpinball (standalone)

```
> time ../vpinball/VPinballX_GL -extractvbs "Godzilla (Sega 1998) VPW v1.0.vpx"
2023-09-25 09:52:40.171 INFO  [8006818] [VPApp::InitInstance@602] Starting VPX...
2023-09-25 09:52:40.171 INFO  [8006818] [VPApp::InitInstance@605] m_logicalNumberOfProcessors=10
2023-09-25 09:52:40.171 INFO  [8006818] [VPApp::InitInstance@607] m_szMyPath=/Users/francisdb/vpinball/vpinball/
2023-09-25 09:52:40.171 INFO  [8006818] [VPApp::InitInstance@608] m_szMyPrefPath=/Users/francisdb/.vpinball/
2023-09-25 09:52:40.171 INFO  [8006818] [VPApp::InitVPX@742] Loading table from command line option: Godzilla (Sega 1998) VPW v1.0.vpx
2023-09-25 09:52:40.171 INFO  [8006818] [VPinball::ProfileLog@2688] OnInitialUpdate
2023-09-25 09:52:40.171 INFO  [8006818] [InitRegistryOverride@159] Using settings override file from:
2023-09-25 09:52:40.171 INFO  [8006818] [LoadIni@144] Settings file was not found at:
2023-09-25 09:52:40.171 INFO  [8006818] [VPinball::ProfileLog@2688] LoadGameFromFilename Godzilla (Sega 1998) VPW v1.0.vpx
2023-09-25 09:52:40.172 INFO  [8006818] [VPinball::ProfileLog@2688] LoadGameFromStorage
2023-09-25 09:52:40.182 INFO  [8006818] [VPinball::ProfileLog@2688] LoadData
2023-09-25 09:52:41.060 INFO  [8006818] [VPinball::ProfileLog@2688] GameItem
2023-09-25 09:52:41.254 INFO  [8006818] [VPinball::ProfileLog@2688] Sound
2023-09-25 09:52:46.602 INFO  [8006818] [VPinball::ProfileLog@2688] Image
2023-09-25 09:52:46.603 INFO  [8006818] [VPinball::ProfileLog@2688] Font
[...]

PR #2402, opened 2025-05-08
fix: init sdl after logs so driver and errors are logged

francisdb:
Mainly because if users send us logs we currently can't see what video driver is in use.

related to unfinished work at #2156

### DIFF
--- a/src/audio/pinsound.cpp
+++ b/src/audio/pinsound.cpp
@@ -1510,10 +1510,10 @@ int PinSound::getChannel()
 void PinSound::EnumerateAudioDevices(vector<AudioDevice>& audioDevices)
 {
    if (!SDL_WasInit(SDL_INIT_AUDIO))
-   if (!SDL_InitSubSystem(SDL_INIT_AUDIO)) {
-      PLOGE << "SDL Init Audio failed: " << SDL_GetError();
-      return;
-   }
+      if (!SDL_InitSubSystem(SDL_INIT_AUDIO)) {
+         PLOGE << "SDL Init Audio failed: " << SDL_GetError();
+         return;
+      }
 
    //output name of audio driver
    const char *pdriverName;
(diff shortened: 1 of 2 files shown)


src/audio/pinsound.cpp
```diff
@@ -1522,18 +1522,9 @@ void PinSound::EnumerateAudioDevices(vector<AudioDevice>& audioDevices)
    else
       PLOGE << "SDL Get Audio Driver failed: " << SDL_GetError();
 
-   // log default audio device
```
francisdb, line 1525: This part was removed because SDL_OpenAudioDevice returns a logical device id that can't be used on SDL_GetAudioDeviceName

I don't think there is a way to get log the name of the default device

francisdb: I got an ID of 27 on macOS and there was no such device.
The docs clearly mention this:

> This function returns the opened device ID on success. This is a new, unique [SDL_AudioDeviceID](https://wiki.libsdl.org/SDL3/SDL_AudioDeviceID) that represents a logical device.

https://discourse.libsdl.org/t/incorrect-audio-device-being-selected-using-sdl-openaudiodevicename/25414/2

francisdb: But let me put back the code and ignore any errors then

superhac: Maybe post an issue over at SDL too.  This should be consistent across all platforms.  It even works on Android now that I think about it.

francisdb: https://github.com/libsdl-org/sdl/issues/12977

jsm174:
I cleaned up some of the formatting for consistency. 

SDL_GetCurrentVideoDriver is only relevant for standalone and standalone external windows. 

Also, fwiw, this change now makes extracting vbs reliant on SDL initialization. 

I think SDL should only be initialized by the player, `listres`, `listsnd`, or `displayid`.

francisdb:
I tried to find the best place to put this with minimal change. Ideally the config loading would be the first thing done before anything else. Also #814 already mentions that way too much is done for extracting vbs.

jsm174:
> Also #814 already mentions that way too much is done for extracting vbs.

True, but since this is how the code is now and isn't changing anytime soon, this should be updated.

src/audio/pinsound.cpp
superhac: Now with the latest pull as of today I'm getting a segfault

I added to the SDL issue [12077](https://github.com/libsdl-org/SDL/issues/12977) to see if its related.
segfault

The `devid` is coming back with as the Default is not a valid ID.  In my case its coming back with 33.

My sound device ID's:
```
2025-05-10 13:26:15.165 INFO  [530133] [VPApp::InitInstance@1003] id 15: name=Starship/Matisse HD Audio Controller Digital Stereo (IEC958), channels=2
2025-05-10 13:26:15.165 INFO  [530133] [VPApp::InitInstance@1003] id 23: name=CM106 Like Sound Device Analog Surround 7.1, channels=8
```

Issue #2366, opened 2025-04-25
PUP: difference in SetBG behaviour with windows 11 version


toccata10:
I'm developping the SetBG problem already reported in https://github.com/vpinball/vpinball/issues/2291

I'm testing on the jumanji puppack with
(2 Screens) Original Overlay, 1 Topper, DMD on backglass bottom, 4th Screen Topper off by Default

Without any modification, it displays:
![Image](https://github.com/user-attachments/assets/4e17cd42-7bf4-4892-829e-8f3c0703a7c3)

Remove this line from `screens.pup`, otherwise it hides almost everything (zorder problem):
```
3,"DefaultTopperLocation (optional)",Backglass,"TOPPER by Jagandy82 TY.mp4",0,off,,"0,0,0,100,100"
```
![Image](https://github.com/user-attachments/assets/ff1bd02f-4f0e-4550-8ff8-e04347dac069)
I's an improvement, but the png overlay is still under the video. 

Unless, I remove the SetBG:
```
7,1,Overlay,D0,5,PupAlphas,"Overlay 1 by Balutito.png",0,1,,,,SetBG,0
```
has to be replaced by
```
7,1,Overlay,D0,5,PupAlphas,"Overlay 1 by Balutito.png",0,1,,,,,0
```
then it's ok (but the 6, Backglass should go deeper in order to display attract mode video)

![Image](https://github.com/user-attachments/assets/4720f5b3-1151-401a-a4b7-6f3eb3a378e8)

Testing on windows 11: `7,1,Overlay,D0,5,PupAlphas,"Overlay 1 by Balutito.png",0,1,,,,SetBG,0` or `7,1,Overlay,D0,5,PupAlphas,"Overlay 1 by Balutito.png",0,1,,,,,0` gives the same (correct) result (with `3,"DefaultTopperLocation (optional)",Backglass,"TOPPER by Jagandy82 TY.mp4",0,off,,"0,0,0,100,100"` still removed from `screens.pup`).

Final touch: if you want a still better result: change (`screens.pup`)
```
6,Backglass,,,0,ForceBack,,"5,15.4,17.8,69.5,64.2"
```
to
```
3,Backglass,,,0,ForceBack,,"5,15.4,17.8,69.5,64.2"
```
And (`triggers.pup`)
```
9,1,Backglass,D0,6,Backglass,"Jumanji Video Backglass by Nesta78.mp4",0,1,,,,SetBG,0
```
to
```
[...]

Issue #2344, opened 2025-04-17
standalone puppack: lower priority video still interrupts current video


toccata10:
I took the monster bash puppack as a base to create a super simple configuration. I first tested everything on windows (with a fresh "baller" install) to ensure I was getting the expected results. Then I switched to standalone Linux.
Switches are described at p116 of the manual (https://www.ipdb.org/files/4441/Williams_1998_Monster_Bash_Operations_Manual_with_schematics_OCR_searchable.pdf ):
W64 is the switch in the upper launch lane.
W54 is the right left bumper
I'll use them to trigger 2 different videos: 1st one with the upper lane switch, with an interrupt as soon as the ball hits the bumpers. I used "Attract_Music" to have a long playing video in some tests. The playlist.pup is unchanged. 

1st configuration (working as expected on standalone): 
screens.pup:
```
ScreenNum,ScreenDes,PlayList,PlayFile,Loopit,Active,Priority,CustomPos
2,Backglass,,,0,ForceBack,,
```

triggers.pup
```
ID,Active,Descript,Trigger,ScreenNum,PlayList,PlayFile,Volume,Priority,Length,Counter,RestSeconds,Loop,Defaults
71,1,Creature1,D0,2,Creature,"Creature Surfacing.mp4",,,,,,SetBG,0
36,1,"Ball Drain",W64,2,BallDrain,,0,10,,,5,,0
```

2nd test on priority: when on same priority it works fine: The BallDrain video shows when triggered, then the "Creature Surfacing" restarts in a loop: that's perfect.
screens.pup is the same

triggers.pup
```
ID,Active,Descript,Trigger,ScreenNum,PlayList,PlayFile,Volume,Priority,Length,Counter,RestSeconds,Loop,Defaults
71,1,Creature1,D0,2,Creature,"Creature Surfacing.mp4",,,,,,SetBG,0
36,1,AttractMode,W64,2,AttractMode,"Attract_Music.mp4",0,10,,,5,,0
37,1,"Ball Saved",W54,2,"Ball Saved",,150,10,,,5,,0
```

3rd test: with higher priority: the ball saved interrupts the Attract_Music: that's perfect
```
ID,Active,Descript,Trigger,ScreenNum,PlayList,PlayFile,Volume,Priority,Length,Counter,RestSeconds,Loop,Defaults
71,1,Creature1,D0,2,Creature,"Creature Surfacing.mp4",,,,,,SetBG,0
[...]

PR #2405, opened 2025-05-09
pup: support for advanced triggers

### LINKED ISSUE #1786: triggers.pup not supporting the equal sign nor multiple conditions (standalone)
**Describe the bug**
While trying to understand why some pup videos don't launch, I made an interesting discovery:
Let's take "back to the future" https://vpuniverse.com/files/file/11060-back-to-the-future-mod-by-nitronimbus-cw-pup-pack/page/5/?tab=comments (for a more convenient download go to https://mega.nz/file/zmpFDJoC#yrdQhrt5Td38Y7VRF3ZWsVBP61KBp7gj1gyGaesy6oE as indicated in the comments).
The ball drain didn't trigger anything:
28,1,Drain,W10=1,2,Drain_5,,100,9,,,8,,0
But if I replace by:
28,1,Drain,W10,2,Drain_5,,100,9,,,8,,0
then it works fine.
So, I'm pretty sure that when there's a equal, it's not triggered.

**Versions & Settings**
 - VPinballX_GL-10.8.1-2362-ad4cd07-Release-linux-x64

francisdb:
eg: `L23,W38=1,W39=1,W40=0`

this fixes #1786

* removed the trigger filters where only 1 values were passed
* keeps track of latest values
* parses all pup triggers
* makes sure all values match

The only thing I am not 100% sure about is if for example in above case `W40` goes to `0` and the rest already matched if we should also cause an event? Currently we see the first one as the main trigger so only a change to that one while the rest already matches can initiate an event.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -3,11 +3,7 @@
 BasedOnStyle: WebKit
 AccessModifierOffset: -3
 AllowShortCaseLabelsOnASingleLine: true
-BreakBeforeBraces: Custom
-BraceWrapping:
-  AfterControlStatement: false
-  BeforeCatch: true
-  AfterFunction: true
+BreakBeforeBraces: Allman
 ColumnLimit: 190
 ConstructorInitializerIndentWidth : 3
 ContinuationIndentWidth: 3
--- a/.clang-format
+++ b/.clang-format
@@ -3,7 +3,11 @@
 BasedOnStyle: WebKit
 AccessModifierOffset: -3
 AllowShortCaseLabelsOnASingleLine: true
-BreakBeforeBraces: Allman
+BreakBeforeBraces: Custom
+BraceWrapping:
+  AfterControlStatement: false
+  BeforeCatch: true
+  AfterFunction: true
 ColumnLimit: 190
 ConstructorInitializerIndentWidth : 3
 ContinuationIndentWidth: 3
--- a/standalone/inc/pup/PUPManager.cpp
+++ b/standalone/inc/pup/PUPManager.cpp
@@ -120,7 +120,7 @@ void PUPManager::LoadPlaylists()
    std::ifstream playlistsFile;
    playlistsFile.open(szPlaylistsPath, std::ifstream::in);
    if (playlistsFile.is_open()) {
-      ankerl::unordered_dense::set<std::string> lowerPlaylistNames;
+      ankerl::unordered_dense::set<string> lowerPlaylistNames;
       string line;
       int i = 0;
       while (std::getline(playlistsFile, line)) {
--- a/standalone/inc/pup/PUPManager.cpp
+++ b/standalone/inc/pup/PUPManager.cpp
@@ -230,14 +230,10 @@ TTF_Font* PUPManager::GetFont(const string& szFont)
 
 void PUPManager::QueueTriggerData(PUPTriggerData data)
 {
-   if (data.value == 0)
-      return;
-
    {
       std::lock_guard<std::mutex> lock(m_queueMutex);
       m_triggerDataQueue.push({ data.type, data.number, data.value });
    }
-
    m_queueCondVar.notify_one();
 }
 
--- a/standalone/inc/pup/PUPScreen.cpp
+++ b/standalone/inc/pup/PUPScreen.cpp
@@ -476,14 +476,8 @@ void PUPScreen::QueueTrigger(char type, int number, int value)
       for (const auto& [expectedTypeNumber, expectedValue] : pTrigger->GetTriggers())
       {
          auto currentValue = m_triggersState.find(expectedTypeNumber);
-         if (currentValue == m_triggersState.end())
-         {
+         if (currentValue == m_triggersState.end() || currentValue->second != expectedValue)
             return;
-         }
-         if (currentValue->second != expectedValue)
-         {
-            return;
-         }
       }
 
       auto* pRequest = new PUPTriggerRequest();
--- a/standalone/inc/pup/PUPScreen.cpp
+++ b/standalone/inc/pup/PUPScreen.cpp
@@ -468,13 +468,11 @@ void PUPScreen::QueueTrigger(char type, int number, int value)
    if (!pTriggers)
       return;
 
-   for (PUPTrigger* pTrigger : *pTriggers)
-   {
+   for (PUPTrigger* pTrigger : *pTriggers) {
       if (!pTrigger->IsActive())
          continue;
 
-      for (const auto& [expectedTypeNumber, expectedValue] : pTrigger->GetTriggers())
-      {
+      for (const auto& [expectedTypeNumber, expectedValue] : pTrigger->GetTriggers()) {
          auto currentValue = m_triggersState.find(expectedTypeNumber);
          if (currentValue == m_triggersState.end() || currentValue->second != expectedValue)
             return;
--- a/standalone/inc/pup/PUPScreen.cpp
+++ b/standalone/inc/pup/PUPScreen.cpp
@@ -257,7 +257,7 @@ void PUPScreen::AddTrigger(PUPTrigger* pTrigger)
    if (!pTrigger)
       return;
 
-   m_triggerMap[pTrigger->GetTrigger()].push_back(pTrigger);
+   m_triggerMap[pTrigger->GetMainTriggerName()].push_back(pTrigger);
 }
 
 vector<PUPTrigger*>* PUPScreen::GetTriggers(const string& szTrigger)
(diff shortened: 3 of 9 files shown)


standalone/inc/pup/PUPManager.cpp
```diff
@@ -230,14 +230,10 @@ TTF_Font* PUPManager::GetFont(const string& szFont)
 
 void PUPManager::QueueTriggerData(PUPTriggerData data)
 {
-   if (data.value == 0)
-      return;
-
    {
```
jsm174, line 233: I remember specifically adding this. I feel like it was related to dof or something, but this would get hammered.

standalone/inc/pup/PUPScreen.cpp
```diff
@@ -459,15 +459,30 @@ void PUPScreen::QueueBG(int mode)
 
 void PUPScreen::QueueTrigger(char type, int number, int value)
 {
-   if (value == 0)
-      return;
+   // we store the whole state to be able to mach later
+   const std::string name = std::string(1, type) + std::to_string(number);
+   m_triggersState[name] = value;
```
jsm174, line 465: core/stdafx.h namespaces std::string and std::vector, so you don't need std:: anywhere you use these.

standalone/inc/pup/PUPScreen.cpp
```diff
@@ -459,15 +459,30 @@ void PUPScreen::QueueBG(int mode)
...
    vector<PUPTrigger*>* pTriggers = GetTriggers(type + std::to_string(number));
    if (!pTriggers)
       return;
 
    for (PUPTrigger* pTrigger : *pTriggers) {
-      PUPTriggerRequest* pRequest = new PUPTriggerRequest();
+      if (!pTrigger->IsActive())
+         continue;
+
+      for (const auto& [m_sName, value] : pTrigger->GetTriggers()) {
```
jsm174, line 475: `m_` represents a member class variable. use name name or something else.

standalone/inc/pup/PUPScreen.cpp
```diff
@@ -459,15 +459,30 @@ void PUPScreen::QueueBG(int mode)
...
+         auto currentValue = m_triggersState.find(m_sName);
+         if (currentValue == m_triggersState.end()) {
```
jsm174, line 477: for formatting consistency, single line if's you don't need braces unless you have another large if, for, or are using PLOG statements. also, it looks like you could have just OR'd it

```
if (currentValue == m_triggersState.end() || currentValue->second != value)
   return;
```

standalone/inc/pup/PUPScreen.h
```diff
@@ -173,4 +173,5 @@ class PUPScreen
    bool m_isRunning;
    std::thread m_thread;
    std::mutex m_renderMutex;
+   std::unordered_map<std::string, int> m_triggersState;
```
jsm174, line 176: should we be using these other maps?

standalone/inc/pup/PUPTrigger.cpp
```diff
@@ -56,11 +56,11 @@ const char* PUP_TRIGGER_PLAY_ACTION_TO_STRING(PUP_TRIGGER_PLAY_ACTION value)
      D = PupCap DMD Match
 */
 
-PUPTrigger::PUPTrigger(bool active, const string& szDescript, const string& szTrigger, PUPScreen* pScreen, PUPPlaylist* pPlaylist, const string& szPlayFile, float volume, int priority, int length, int counter, int restSeconds, PUP_TRIGGER_PLAY_ACTION playAction)
+PUPTrigger::PUPTrigger(bool active, const string& szDescript, const std::vector<StateTrigger>& vTriggers, PUPScreen* pScreen, PUPPlaylist* pPlaylist, const string& szPlayFile, float volume, int priority, int length, int counter, int restSeconds, PUP_TRIGGER_PLAY_ACTION playAction)
 {
    m_active = active;
    m_szDescript = szDescript;
-   m_szTrigger = szTrigger;
+   m_vTriggers = vTriggers;
```
jsm174, line 63: we don't usually use v to represent vectors. just the name.

standalone/inc/pup/PUPTrigger.cpp
```diff
@@ -159,6 +159,46 @@ PUPTrigger* PUPTrigger::CreateFromCSV(PUPScreen* pScreen, string line)
    );
 }
 
+std::vector<StateTrigger> PUPTrigger::ParseTriggers(const std::string& triggerString) {
```
jsm174, line 162: same std::vector -> vector comment from earlier.

standalone/inc/pup/PUPTrigger.cpp
```diff
@@ -159,6 +159,46 @@ PUPTrigger* PUPTrigger::CreateFromCSV(PUPScreen* pScreen, string line)
...
+            }
+            trigger.value = std::stoi(token.substr(equalPos + 1));
+         } else {
+            // Parse triggers without state (e.g., "S10")
+            trigger.m_sName = token;
+            trigger.value = 1;
+         }
+
+         vTriggers.push_back(std::move(trigger));
+      } catch (const std::invalid_argument& e) {
```
jsm174, line 192: catches on next line for consistency

standalone/inc/pup/PUPTrigger.h
```diff
@@ -19,17 +19,39 @@ typedef enum
    PUP_TRIGGER_PLAY_ACTION_CUSTOM_FUNC     // Call a custom function
 } PUP_TRIGGER_PLAY_ACTION;
 
+struct StateTrigger{
```
jsm174, line 22: we tried to stick with prefixing structs with PUP for example, also space after Trigger {

standalone/inc/pup/PUPTrigger.h
```diff
@@ -19,17 +19,39 @@ typedef enum
...
 {
 public:
-   ~PUPTrigger() {}
-
+   ~PUPTrigger() { }
    static PUPTrigger* CreateFromCSV(PUPScreen* pScreen, string line);
    bool IsActive() const { return m_active; }
    const string& GetDescription() const { return m_szDescript; }
-   const string& GetTrigger() const { return m_szTrigger; }
+   /**
```
jsm174, line 38: Any class methods that have lots of code, I'd rather have in the cpp instead of the h file. Especially since you have lots of comments about the purpose of the function.

if you wanted to stay in, you could do something like:

```
const string& GetMainTriggerName() const { return !m_triggers.empty() ? m_triggers.front().m_sName : NO_TRIGGER; }
```

jsm174 (changes requested):

standalone/inc/pup/PUPManager.cpp
francisdb: Pup triggers can trigger when a value goes to 0. Not sure why dof is related as this is in the pup code?

jsm174: As I said, it was a long time since I wrote that, but I do remember specifically putting that in for a reason.  We'll see what happens.

standalone/inc/pup/PUPScreen.cpp
francisdb: Can such things be put in the clang-format file?

francisdb:
I will squash once this gets through the review

standalone/inc/pup/PUPTrigger.cpp
```diff
@@ -76,7 +76,8 @@ PUPTrigger::PUPTrigger(bool active, const string& szDescript, const string& szTr
 PUPTrigger* PUPTrigger::CreateFromCSV(PUPScreen* pScreen, string line)
 {
    vector<string> parts = parse_csv_line(line);
-   if (parts.size() != 14) {
+   if (parts.size() != 14)
```
jsm174, line 79: I think the brace formatting got changed on this file. { after if's etc.

francisdb: Is there some kind of command I can run to validate formatting or even better reformat?

jsm174: unfortunately no, it's more just stick with the rest of the code base for consistency. sorry

.clang-format
```diff
@@ -3,7 +3,11 @@
 BasedOnStyle: WebKit
 AccessModifierOffset: -3
 AllowShortCaseLabelsOnASingleLine: true
-BreakBeforeBraces: Allman
+BreakBeforeBraces: Custom
+BraceWrapping:
+  AfterControlStatement: false
```
vbousquet, line 8: Not sure why you want to change this. The clang-format is not that old and the codebase is being normalized slowly as the code is refactored/updated (we decided against a global format update). Whanging the clang-format now would mean that a large part of the codebase would have to be updated and that it will take longer to get consistent formatting in the project.

francisdb: Because @jsm174 told me the rules were different.

jsm174: There are no official rules. I was just asking you to keep the code similar to what we had in every other PUP file. 
I don't have my editor try to automatically format my code.. That stuff never works 100%.

vbousquet: No problem, just please, don't modify the clang-format file as it will makes things a bit more difficult in the src folder where I have autoformatting enabled in MSVC

francisdb: reverted

francisdb:
> I have been facing the same difficulties on the pup plugin I'm implementing at the moment. In the end, I had to entirely remove triggering through a string (`QueueTrigger`) since it would fail to evaluate correctly if faced to more complex triggers that would not match the string lookup. If you want, you can have a look at the branch where things are being prepared: https://github.com/vbousquet/vpinball/tree/pup_plugin

So I guess any further work for #1786, #2366 or #2344 better be postponed until that plugin is ready?

jsm174:
@vbousquet - I merged this, but I talked to @nailbuster to better understand how these work, and was going to start working on a revision. I had an idea on what I wanted to do, but I'll take a look at your work.

vbousquet:
> > I have been facing the same difficulties on the pup plugin I'm implementing at the moment. In the end, I had to entirely remove triggering through a string (`QueueTrigger`) since it would fail to evaluate correctly if faced to more complex triggers that would not match the string lookup. If you want, you can have a look at the branch where things are being prepared: https://github.com/vbousquet/vpinball/tree/pup_plugin
> 
> So I guess any further work for #1786, #2366 or #2344 better be postponed until that plugin is ready?

I based my work on the standalone PUP renderer. So whatever you do will help. The plugin system is a long shot and will need time to be stable, so pushing updates to regular standalone allows to improve things, have more regular mobile release, and is also usefull for the plugin branch.

PR #2409, opened 2025-05-10
standalone: more multiple trigger enhancements

jsm174:
This PR makes the following changes:

- `PUPTriggerCondition` is now a class that contains an `Evaluate` method and is responsible for parsing the triggers field
- `PUPManager` now contains a cache for all past trigger states
- `PUPTrigger` is now responsible for evaluating if it has triggered
- Fixes an issue with `PUPLabel` being marked dirty [thanks @vbousquet]

Tested on TNA, B66, and MB.

- Does introduce a regression where the background is not showing up in TF.  Will follow up with a fix

### DIFF
--- a/standalone/inc/pup/PUPTriggerCondition.cpp
+++ b/standalone/inc/pup/PUPTriggerCondition.cpp
@@ -1,0 +1,40 @@
+#include "core/stdafx.h"
+
+#include "PUPTriggerCondition.h"
+
+#include "PUPManager.h"
+
+PUPTriggerCondition::PUPTriggerCondition(const string& szId, int value)
+{
+   m_szId = szId;
+   m_value = value;
+}
+
+vector<PUPTriggerCondition*> PUPTriggerCondition::CreateFromCSV(const string& line)
+{
+   vector<PUPTriggerCondition*> conditions;
+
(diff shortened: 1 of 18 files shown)


standalone/inc/pup/PUPTriggerCondition.cpp
```diff
@@ -0,0 +1,40 @@
+#include "core/stdafx.h"
+
+#include "PUPTriggerCondition.h"
+
+#include "PUPManager.h"
+
+PUPTriggerCondition::PUPTriggerCondition(const string& szId, int value)
```
francisdb, line 7: Where did the error handling go? If I write a pup pack with invalid triggers I want to know.

jsm174: Sorry if I missed something. Can you add some
more details and I’ll put it back.

jsm174: I'd say this would never happen, but we have seen some crazy things in pup files.

When I refactored, instead of `stoi`, I used all the helper functions we made along the way and are all over in pup and flex. In this case specifically `string_to_int`, (which leverages more custom functions `try_parse_int` which leverages `trim_string`)

PR #2434, opened 2025-05-17
windows: allow non-playfield windows to use per table settings

jsm174:
As we work to switch FlexDMD and VPinmameController for Standalone to use the new ScoreView, all windows are being loaded from the master ini. The ScoreView (especially on mobile), needs to be configurable per game. On desktop, for example CheckPoint's ScoreView needs to be placed in a small section over the backglass. I believe @francisdb had use cases for this on Linux as well.


vbousquet (changes requested):
This is a nice and needed change.

One thing that is missing is that `SetPos` will still use the main settings even if the window is created from a table settings. I think the Window should hold a reference on the settings and use it in `SetPos`.

Also, to actually uses per table settings, player should use m_ptable->m_settings for creating Wdinows

jsm174:
Cool.

Scoreviews window is already created with live_table->m_settings so we should be good there?

I’ll look into SetPos.

vbousquet:
Indeed, you are right that anciliary windows are created using table's settings. So this is already good

francisdb:
For me the goal is putting the score view in (not over) the b2s "fulldmd" window and configured per table using relative coordinates so that switching to a different pc / cab / screenres keeps the DMD in the correct place.

Are there plans to be able to embed the score window in other windows than the playfield?

Ultimate goal is no longer requiring b2s files and the backglass/fullDMD windows just being cameras on the vr table? But I have no idea how a backglass is rendered in vr. Is that something like the desktop mode or actually using a b2s?

jsm174:
I had some difficulty saving the aux windows positions. I tried to keep a reference to the settings when the window gets created, but the pointer to `live_table->m_settings` seems to change. 

I ended up with:

```
Settings* settings = (m_settingsSection == Settings::Player) ? &(g_pvp->m_settings) : &(g_pvp->m_ptableActive->m_settings);
```

Hopefully this is okay. If not we can fix in another PR.

Issue #2418, opened 2025-05-12
pup: another memory leak


francisdb:
linux
bgfx
master branch

https://vpuniverse.com/files/file/16461-simpsons-treehouse-of-horror-mod-simpsons-pinball-party-the-stern-2003-vpw-mod/

- Only with the full pup pack enabled
- Leak starts when you press the start button
- Sound begins to stutter after a while like we saw with #2388 

![Image](https://github.com/user-attachments/assets/492f5814-e874-4f32-9e2c-01a60e8e6443)

francisdb:
It's elsewhere, found where it happens but still looking at the cause

francisdb:
It's in the mp3 background music playback on that specific pup pack. Was already the case before the media player updates.

francisdb:
I don't get why this leaks

To reproduce with the above linked table
 
* in `triggers.pup` set active to `0` for all `W54` triggers except `137 MusicStart`
* launch the table
* press the start button

You will see it start leaking memory **very fast**

Workaround

comment out

```cpp
m_pPinSound->StreamUpdate(pBuffer, resampledDataSize);
```

in

https://github.com/vpinball/vpinball/blob/bb91cf583e49428e0d805876aea312000f34b2b7/standalone/inc/pup/PUPMediaPlayer.cpp#L591-L594

No more background music, no more leak. But why does `SDL_PutAudioStreamData(m_pstream, buffer, length)` leak memory?

francisdb:
The only thing I find fishy is that play is called twice directly after pressing the start button, maybe there's a problem with two threads playing sound at the same time?

francisdb:
Found another pup pack with this issue: `Stranger Things - Stranger Edition`

PR #2438, opened 2025-05-19
standalone: temp throttle audio only streams in pup media player

jsm174:


jsm174:
workaround for: https://github.com/vpinball/vpinball/issues/2418

PR #2441, opened 2025-05-20
standalone: add Collection::Item proxy

francisdb:
For this case

```vbscript
	Set kicRouletteStart = colBonusKickers.Item(KickerWithBall)
```

which is the same as

```vbscript
	Set kicRouletteStart = colBonusKickers(KickerWithBall)
```

### DIFF
--- a/standalone/vpinball_standalone_i_proxy.cpp
+++ b/standalone/vpinball_standalone_i_proxy.cpp
@@ -7,7 +7,8 @@ STDMETHODIMP Collection::GetIDsOfNames(REFIID /*riid*/, LPOLESTR* rgszNames, UIN
 		DISPID dispId;
 	} namesIdsList[] = {
 			{ NULL },
-			{ L"Count", 8000 }
+			{ L"Count", 8000 },
+			{ L"Item", 8001 }
 	};
 
 	size_t min = 1, max = ARRAY_SIZE(namesIdsList) - 1, i;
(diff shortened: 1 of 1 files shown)


francisdb:
@jsm174 not sure where these magic numbers come from so just iused previous + 1

jsm174:
Thanks for this. I've had a long standing issue with my custom IDL parser when it came to collections where I did some manual tweaking. 

Anyway, `Items` and `_newenum` are special ones. `Items` is actually `DISPID_VALUE` (`0`). We've never had a table call this externally, so it never went through `GetIDsOfNames` first. It always went directly to `Invoke`.

I will fix the parser as this PR is not correct. I believe the code was already there and I think the fix would be:

```
namesIdsList[] = {
			{ NULL },
			{ L"Count", 8000 },
                        { L"Item", DISPID_VALUE }
	};
```

As far as magic numbers, nothing magic about it. 

- In the idl file, some people/tools define it, for example

```
		[id(1), helpstring("method Run")] HRESULT Run([in,defaultvalue(0)] LONG_PTR hParentWnd, [in,defaultvalue(99)] int nMinVersion);
```

so the parser would take `id(1)`.

- since `DISPID_VALUE` is 0, it would inherit `id(0)`

- when `id` isn't included we need to satisfy the `switch/case`, so I just generate a number starting at 8000 incrementing by 1. (This was to try to avoid any collisions, but technically could happen)

https://github.com/vpinball/vpinball/blob/2b032881c27ddecc9b171e8741e86037cbc4561e/standalone/idl/parser/src/main/java/org/vpinball/Method.java#L35-L40

francisdb:
> As far as magic numbers, nothing magic about it.

In my book that 8000 is a magic number and deserves a constant.

> I will fix the parser as this PR is not correct.

Are you sure it's not correct? You mean you could reduce the copy paste? 

> Something else that is probably incorrect is that the `if` case is wider than the `else if` making the else obsolete?

https://github.com/vpinball/vpinball/blob/2b032881c27ddecc9b171e8741e86037cbc4561e/standalone/vpinball_standalone_i_proxy.cpp#L39-L52

jsm174:
> Not sure if this helps but I also had to handle default methods for the plugin script support. So DISPID_VALUE = 0 means default object member which is object dependent and I believe can depend on the call context (call arguments)... This is stated in a few places in the MS docs like here: https://learn.microsoft.com/en-us/previous-versions/windows/desktop/automat/dispid-constants. For collection, the default member is likely accessing the collection content, hence the Items array property.

Thanks! I've updated the parser to catch the other id's, but luckily nothing else is in use right now. I also added `// Default method` comments to the proxies.

jsm174:
> In my book that 8000 is a magic number and deserves a constant.

I'm reading from a different book on this specific issue.

> Are you sure it's not correct? You mean you could reduce the copy paste?

As I initially alluded to, your fix worked, but really required a much larger fix and cleanup. 

I also added a `README.md` since this was one of the last things I never documented.

Thanks again for finding this!

PR #2470, opened 2025-05-27
plugins: add support for building static plugins

jsm174:
This is needed for mobile builds. Currently it only enables ScoreView.

plugin assets will be copied in the applications resources for iOS, and assets for Android (not to be confused with src/assets).

iOS:
```
rm -rf "${BUNDLE_RESOURCES}/plugins"
mkdir -p "${BUNDLE_RESOURCES}/plugins/scoreview"
cp -r "$PROJECT_DIR/../../plugins/scoreview/layouts" "${BUNDLE_RESOURCES}/plugins/scoreview"
```

Android:
```
        from("${layout.buildDirectory}/../../../../plugins/scoreview/layouts") {
           into("plugins/scoreview/layouts")
        }
```

Also, all compile definitions have been moved `target_compile_definitions` in all cmakes. This way, VPinball's cmake doesn't pollute the plugin's cmakes.


vbousquet (approved):
Looks good to me. Nothing to add on my side

PR #2501, opened 2025-06-04
libvpinball: rework settings to use section name strings

jsm174:
This updates mobile builds to use section name strings rather than ints for loading and saving settings.
This will allow setting and loading values for plugins, such as `[Plugin.ScoreView]` and `[Plugin.DMDUtil]` 

Also defaults for mobile were moved from the native apps, back to libvpinball using `Settings::Validate` (which I didn't realize existed, and is super awesome)


jsm174:
This is now ready to go. 

All changes to Settings have been wrapped with either, `__LIBVPINBALL__`, `__STANDALONE__`, or `__ANDROID__` to not disrupt Windows users.

After this commit, mobile users should reset their settings. 

PinMAME Plugin, ScoreView Plugin, AlphaDMD Plugin, DMDUtil Plugin have been tested successfully on mobile.
FlexDMD Plugin does not seem to load the script overrides.

VPinMAMEController and B2S from legacy standalone are working on mobile.

PR #2513, opened 2025-06-09
Use placeholder when b2s image fail to be interpreted.

Nicals:
When loading the "Die Hard Trilogy" table ([vpuniverse.com/files/file/17645](https://vpuniverse.com/files/file/17645-die-hard-trilogy-vpw-2023/)), VPinball crashes with a segmentation fault during DirectB2S file processing.

The crash occurs when FormBackglass::Base64ToImage() fails to decode base64 image data and returns nullptr.
This method currently returns NULL when something is wrong but there aren't any checks on NULL on the caller side.

The decode failure was traced to empty or placeholder "Image" attributes in the DirectB2S file (
[Die Hard Trilogy (VPW 2023) v0.98.directb2s.gz](https://github.com/user-attachments/files/20652031/Die.Hard.Trilogy.VPW.2023.v0.98.directb2s.gz)
).

This PR adds some validation in the Base64ToImage method, logs any error and returns a 1px placeholder if anything fails.
This way, we should have a degraded backglass in case of failure to decode an image but no more crash.

I'm not sure this is the best solution. I'm not familiar enough with B2S to properly interpret what represents an empty Image attribute.

### DIFF
--- a/standalone/inc/b2s/forms/FormBackglass.cpp
+++ b/standalone/inc/b2s/forms/FormBackglass.cpp
@@ -1625,11 +1625,24 @@ SDL_Surface* FormBackglass::CropImageToTransparency(SDL_Surface* pImage, SDL_Sur
 
 SDL_Surface* FormBackglass::Base64ToImage(const string& image)
 {
+   if (image.empty()) {
+      PLOGW.printf("Found empty image.");
+      return CreatePlaceholder();
+   }
+
    vector<unsigned char> imageData = base64_decode(image);
+
+   if (imageData.empty()) {
+      PLOGE.printf("Base 64 image decode failed or resulted in empty data.");
+      return CreatePlaceholder();
+   }
+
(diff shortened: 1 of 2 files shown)


jsm174:
@Nicals - Thanks for this! I need a little time to review. 95% of B2S has been ported line for line which makes it easy to compare the two code bases. So whenever we deviate, I just like to be extra careful.

This is really interesting as it's been the first time that I've seen this case come up.

Nicals:
> @Nicals - Thanks for this! I need a little time to review. 95% of B2S has been ported line for line which makes it easy to compare the two code bases. So whenever we deviate, I just like to be extra careful.
> 
> This is really interesting as it's been the first time that I've seen this case come up.

First time too. I'm sure it was working fine some time ago with this exact same directb2s file.

jsm174:
@Nicals - closing this as the standalone B2S has been migrated to the `B2SLegacy` plugin. Also, it no longer uses `SDL_image` for image loading but VPX's internal `Texture` classes.

Issue #2516, opened 2025-06-10
Backglass file not found when there is a space in the tablename


mkl66:
In the latest windows version (V10.8.1 Beta (Rev.3545)) there is an issue with the backglass when the table name contains a space character. The  *.direct2b file could not be found.
Example: 
When I load the table "Zephy (LTD 1982).vpx" into Visual Pinball and then try to play the table I got a warning that the file "Zephy.directb2s" could not be found. Therefore the backglass is not displayed.
When I rename the backglas to the name without any blanks (zephy.direct2bs) then everything works OK.

mkl66:
Today I checked other tables and found out that my initial assumption was wrong. It has nothing to do with a space in the name of the backglass file. VPX uses the name of the romfile as the name of the backglass file.
ex.: 24 (Stern 2009) has "twenty5_150" as cGameName and tries to load "twenty5_150.directb2s"

mkl66:
correction : it' "twenty4_150"

garybrowndev:
Discovered this too the other day and I believe I tracked it down to B2S not being able to find the Visual Pinball window (due to it being hidden early on in the startup process) and grab the name from the title.

See PR https://github.com/vpinball/vpinball/pull/2529 with more details

sverrewl:
Could we put some eyes on this again ? Personally, I find it annoying that b2s isn't starting with newer builds. It is an essential part to the experience for a cab. So, new versions are just left alone and not tested by me very often. Having only rom name as an alternative, is a problem. As you all know, not all machines uses roms.

Pigjes:
Yes, indeed not all machines use roms and won't show a backglass.

vbousquet:
This should be solved using latest B2S builds

PR #2529, opened 2025-06-19
Delayed hiding of window so B2S.Server can find the VPX window name

garybrowndev:
This should fix [https://github.com/vpinball/vpinball/issues/2516](url)
Introduced by [https://github.com/vpinball/vpinball/commit/1a6b1120392f08fa89b17b26c7dcaea1bc66fe1f](url)
Although, Im not 100% sure as there still may be a timing issue where B2S Backglass Server make take a little more time to load and then still not find the window. I've tested this build in Debug and Release environments.

The hiding of the window changed behavior that b2sbackglassserver was relying on.
In this file [https://github.com/vpinball/b2s-backglass/blob/master/b2sbackglassserver/b2sbackglassserver/Classes/Processes.vb](url), the follow code below determines the name of the directb2s file based on the window name of the running Visual Pinball editor window. See snippet of code below:

This code enumerates all "IsWindowVisible()" windows. Due to hiding the window, this function no longer finds the VPX window
`    Private Function EnumWinProc(ByVal hwnd As IntPtr, ByVal lParam As Int32) As Boolean
        If IsWindowVisible(hwnd) AndAlso GetParent(hwnd) = IntPtr.Zero AndAlso GetWindowLong(hwnd, GWL_HWNDPARENT) = 0 Then
            Dim str As String = String.Empty.PadLeft(GetWindowTextLength(hwnd) + 1)
            GetWindowText(hwnd, str, str.Length)
            If Not String.IsNullOrEmpty(str.Substring(0, str.Length - 1)) Then windowlist.Add(New ProcInfo(str.Substring(0, str.Length - 1), hwnd))
        End If
        EnumWinProc = True
    End Function
    Private Sub RefreshWindowList()
        windowlist = New List(Of ProcInfo)
        EnumWindows(AddressOf EnumWinProc, CInt(True))
    End Sub
`

Then during startup of the B2S Backglass Server, the following code is run which looks for "Visual Pinball - <TABLE_NAME>".
`	    Public Sub New()
        RefreshWindowList()
        For Each proc As ProcInfo In windowlist
            If Not String.IsNullOrEmpty(proc.Name) Then
                If proc.Name.StartsWith("Visual Pinball - ", StringComparison.CurrentCultureIgnoreCase) Then
                    'Visual Pinball - [Tom and Jerry (Original 2019) v 1.33]
                    If String.IsNullOrEmpty(_tablename) Then
                        _tablename = proc.Name.Substring(17)
                        '[Tom and Jerry (Original 2019) v 1.33]___
                        If _tablename.StartsWith("[") AndAlso Not _tablename.EndsWith("]") Then
	`

### DIFF
--- a/src/core/vpinball.cpp
+++ b/src/core/vpinball.cpp
@@ -1048,7 +1048,6 @@ void VPinball::DoPlay(const int playMode)
       return;
 
    PLOGI << "Starting Play mode [table: " << table->m_tableName << ", play mode: " << playMode << ']';
-   ShowWindow(SW_HIDE);
    bool initError = false;
    if (false)
    {
@@ -1212,6 +1211,8 @@ void VPinball::DoPlay(const int playMode)
       #else
       auto processWindowMessages = []() {};
       #endif
+      ShowWindow(SW_HIDE);
+
       g_pplayer->GameLoop(processWindowMessages);
 
       #if (defined(__APPLE__) && (defined(TARGET_OS_IOS) && TARGET_OS_IOS))
@@ -1221,9 +1222,9 @@ void VPinball::DoPlay(const int playMode)
 
       delete g_pplayer;
       g_pplayer = nullptr;
-   }
 
-   ShowWindow(SW_SHOW);
+     ShowWindow(SW_SHOW);
+   }
 
    if (initError)
    {


toxieainc:
Hi! Couldn't there be some other mechanism that can also check for minimized windows, etc?

vbousquet:
Hi all and sorry for the delay! I had a look and I too think that there should be a better way to fix this. So I dived in the B2S code but it's not the easiest one. So, for the time being, we have a communication between VPX and B2S using:
- the wIndow title (which is not really garanteed)
- the window behavior (being visible at a precise time point)
- the fact that the script is started at a specific point in time (which happens to have changed in 10.8: it used to be after player start, now it is before, luckily it looks like it did not break).

These way of communicating are obviously hacky and prone to break. Ideally we should have a cleaner way to send the required information, for example through a dedicated call. At least, for the time being, I would prefer this to be fixed on B2S side (which implements the hacky behavior) perhaps by searching all windows, including invisible ones ?

garybrowndev:
Agreed that the current VPX<-->B2S implementation is "hacky". This PR was purposely meant to be simple/quick and not to redesign the communication mechanism between VPX and B2S (I should have mentioned this in the original comment).

I think removing the check for "IsVisible" would work, however, it would require everyone in the world to update their B2S software, not sure if we would want to require that just yet.

I believe the "best" solution would be to research and utilize a well known (hopefully simple), cross-platform inter-process communication technique. This can be accomplished in various ways using registries, files, global pipes, shared memory...unfortunately, none of which are great solutions. This is probably why the original implementor went with the Window Title trick, which is essentially another type of "shared memory".

If all ideas are up for debate, I would suggest modifying the Controller code that creates and launches the B2S process from the VBScript in VPX. Im not sure the rammifications of this (for example, if all VB script files for all tables would need to be changed, this is a bad idea). Im not sure exactly how to do it, but if there was a way to pass in the name of the VPX file during the B2S startup (like the sample code below), I think this would be the most stable and less prone to error.

`B2SController = CreateObject("B2S.Server")`
`B2SController.Run(<pass_VPX_Filename_Here_Somehow??>)`

sverrewl:
@garybrowndev Thanks for this pr. A bit fed up with auto builds that has totally broken db2s functionality. So for now, compiling my own VPX with your suggested change.

vbousquet:
I have proposed the simple fix to B2S. This is really a B2S issue, not VPX, so I think this has to be solved there.

PR #2533, opened 2025-06-21
plugins: add new dof plugin

jsm174:
This PR lays the plumbing to be able to start test libdof on Windows.

Currently right now Pinscape, Pinscape Pico, and TeensyStripControllers work.

I'll leave this in draft until these two issues can be figured out:

```
static void onGameStart(const unsigned int msgId, void* userData, void* msgData)
{
   const CtlOnGameStartMsg* msg = static_cast<const CtlOnGameStartMsg*>(msgData);
   assert(msg != nullptr && msg->gameId != nullptr);

   if (pDOF) {
      LOGI("DOFPlugin: onGameStart");

      VPXTableInfo tableInfo;
      vpxApi->GetTableInfo(&tableInfo);
      pDOF->Init(tableInfo.path, msg->gameId);
   }
}
```

1) msg always resolves at null. This is needed to send the romid to libdof.

2) @vbousquet, how can we get all the events to forward to libdof?


jsm174:
I believe this as a first draft is ready to go.  I'm not sure how to generate the Visual Studio project files.

jsm174:
Wemos D1 Mini Pro
Teensy
Pinscape Pico
Pinscape

all tested as working

AutoConfig in `CabinetConfig.xml` should be disabled as PinOne devices are search for and this can take a long time

```
  <AutoConfigEnabled>false</AutoConfigEnabled>
```

Issue #1222, opened 2023-12-01
VPX editor forcing screen to rotate upside down on Lenovo Legion Go


gregorybeal:
**Describe the bug**
The Lenovo Legion Go is set to Landscape(flipped) screen orientation by default. When opening the VPX editor it forces the screen into Landscape, which on this device is upside down. I have turned off auto-rotate and it is still persisting 

**To Reproduce**
Launch VPX editor or table on Legion Go.

If the bug appears when playing a specific table, give a link to a place where the exact version of the table file can be downloaded.

**Expected behavior**
The VPX window adhere to current windows rotation setting.

**Versions & Settings**
 - VPX:  10.7 and 10.8 beta
 - Windows 11
 - If this bug report is for VPX 10.8+, please join your setting file (locate 'VPinballX.ini' in '%AppData%\VPinballX', either zip it or change its extension to '.log', then drag'n drop it here)
 - If this bug report is for VPX 10.8+, please enable log in your settingd and join your log file

[VPinballX.log](https://github.com/vpinball/vpinball/files/13531417/VPinballX.log)
![IMG_2508](https://github.com/vpinball/vpinball/assets/56313762/589eea26-8736-4042-86df-9ca8312edbad)

mxl6:
I have the same device and issue.

USNCPO:
same device and same issue, something contained in the vpx exe causing this.  whats funny is Future Pinball games render correct orientation.  Strictly a vpx.exe issue.
good to see I wasn't the only one having this issue.

PeterTucker:
Bump. Same on my Legion Go. Please fix.

PeterTucker:
Also, it does this for laptops 2 in 1s. So if you want to turn your Laptop sideways, and play pinball, you simply can't.

USNCPO:
Bump. Need a standalone .vpx exe by the developers of the actual vpx code to fix this.  They are so busy though with everything else it could be a while or never.  :-(

toxieainc:
Note to self: This means we should revisit https://github.com/vpinball/vpinball/blob/c9b276972aed9db2dc9345882102b99a8b46b4b3/src/core/main.cpp#L429

But back then there was also a reason to add this, so best would be to have a developer that owns such an affected device (i do not) to fool around with that (make it a configurable setting or something that could be detected).

USNCPO:
> Note to self: This means we should revisit
> 
[...]

PR #2543, opened 2025-06-27
Update orientation settings to handle Landscape (flipped) for tablet and gaming devices

Dunkalunk:
This should fix Issue #1222 

I started using VPX only a few weeks ago and ran into the same issue mentioned on my Legion Go.  I've been a developer for several years but have done hardly anything in C, so full disclosure, I used AI tools to find the solution.  But it builds and plays properly on my Legion Go after making these changes.  Tested with Space Cadet, Terminator 2, Addams Family, and a Super Mario table, so it should be good to go.  The fix is also in the spot where a comment was added saying "disable auto-rotate on tablets" that was mentioned in the 1222 issue, so it should be the right place.

Feel free to make any adjustments or give me feedback and I will happily build and test on my Legion Go again.

Thanks!

### DIFF
--- a/src/core/main.cpp
+++ b/src/core/main.cpp
@@ -506,12 +506,19 @@ public:
 #ifdef _MSC_VER
       // disable auto-rotate on tablets
 #if (_WIN32_WINNT <= 0x0601)
-      SetDisplayAutoRotationPreferences = (pSDARP)GetProcAddress(GetModuleHandle(TEXT("user32.dll")),
-         "SetDisplayAutoRotationPreferences");
+      pSDARP SetDisplayAutoRotationPreferences = (pSDARP)GetProcAddress(GetModuleHandle(TEXT("user32.dll")), "SetDisplayAutoRotationPreferences");
       if (SetDisplayAutoRotationPreferences)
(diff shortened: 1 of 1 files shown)


vbousquet:
Thanks a lot for this!

I did merge with a little change to keep the use of the diretc function on post Win7 builds : https://github.com/vpinball/vpinball/commit/f7a17c961cf4d90cb8026890e2b43cb3ffe2753b

Dunkalunk:
@vbousquet perfect, I assumed there would be some adjustments needed.  I'll test it out as soon as I can!  Thanks!

USNCPO:
sorry to repost. so how do i fix the issue? do I copy and paste something somewhere or is there a new file to download. sorry I don't understand what to do to fix my legion go screen flipping. 
thanks. 
r/
mike

vbousquet:
Nothing to do, it should be fixed in next version of VPX (10.8.1)

USNCPO:
> Nothing to do, it should be fixed in next version of VPX (10.8.1)

Thank you for replying and for the guidance. Much appreciated.

PR #2560, opened 2025-07-02
Update README.md

teamsuperpanda:
Add a link to an unofficial community wiki in the docs. The goal is to gather useful info in one place. It’s currently focused on Batocera and the standalone beta, but other areas are being added gradually. It’s a work in progress, but a good starting point.


toxieainc:
Sounds good, thanks!

PR #2578, opened 2025-07-12
standalone: fix MakeChar segfault. Prevent LiveUI assertions

jsm174:
@toxieainc, @vbousquet - not sure if I made the correct fix, but `SortAgainstValue` calls `MakeChar` with L"", which then returns `null` which segfaults `lstrcmp`. All the other `Make*` seem to return an empty string, so I made it do the same.
 
```
int CodeViewDispatch::SortAgainstValue(const wstring& pv) const
{
   const char* szName1 = MakeChar(pv.c_str());
   const char* szName2 = MakeChar(m_wName.c_str());
   int ret = lstrcmp(szName1, szName2);
   delete[] szName1;
   delete[] szName2;
   return ret;
}
```

`standalone/inc/wine/wine.c:`

```
__forceinline INT WINAPI lstrcmpA(LPCSTR str1, LPCSTR str2)
{
   return strcmp(str1, str2);
}
```

### DIFF
--- a/src/core/def.cpp
+++ b/src/core/def.cpp
@@ -287,7 +287,7 @@ wstring MakeWString(const char * const sz)
 char *MakeChar(const WCHAR* const wz)
 {
    const int len = WideCharToMultiByte(CP_ACP, 0, wz, -1, nullptr, 0, nullptr, nullptr); //(int)wcslen(wz) + 1; // include null termination
-   if (len <= 1)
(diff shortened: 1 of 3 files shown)


toxieainc:
@jsm174 Thanks! I actually kinda expected that some place would not check for nullptr, but didn't encounter one myself so far so left it like is.

PR #2598, opened 2025-07-21
feat: Add comprehensive VPX 10.8.1 configuration guide (FR/EN)

Le-Syl21:
🎯 Complete Visual Pinball X configuration documentation with real default values

✨ Features:
- Exhaustive .ini file documentation in French and English
- Real default values extracted from Settings.cpp source code
- 13 detailed sections covering all VPX settings
- Comprehensive explanations for each parameter
- Cabinet, VR, DOF, and plugin configurations
- Keyboard/gamepad mappings with DirectInput codes
- DMD/Alpha display profiles for different hardware
- Editor and script syntax highlighting setup

📚 Sections included:
- Standalone & Plugin configurations
- Audio, controls, and physics settings
- Video rendering and 3D stereo options
- VR headset and cabinet hardware setup
- Complete keyboard and gamepad mappings
- DMD and segmented display profiles
- Editor workflow and script highlighting
- Default object properties for table creation

🌍 Bilingual support for international VPX community 🔧 Ready-to-use configuration templates
📖 Technical reference for developers and cabinet builders

This documentation bridges the gap between VPX's powerful features and user-friendly configuration, making advanced pinball simulation accessible to everyone! 🚀⚡

### DIFF
--- a/src/assets/Default_VPinballX.ini
+++ b/src/assets/Default_VPinballX.ini
@@ -1,92 +1,367 @@
 ; #######################################################
-; #  Visual Pinball X settings file
+; #  Visual Pinball X 10.8.1 - Complete Configuration File
+; #  with exact default values extracted from source code
 ; #
-; # This file holds all visual pinball settings.
-; # If you need to reset all settings to their default,
-; # just delete this file and it will be recreated on
-; # next application start, eventually using your
-; # existing settings from Windows registry.
+; # This file contains all Visual Pinball settings with
+; # real default values and detailed explanations.
 ; #
-; # When a property is not defined (nothing after the
-; # equal '=' sign), VPX will use a default value for it.
+; # When a property is not defined (nothing after the '=' sign),
+; # VPX will use the default value indicated in the comments.
 ; #
 ; #######################################################
-; # Standalone
-[Standalone]
-VPRegPath = 
-# Path to the PinMAME folder. (Default is ~/.pinmame)
-PinMAMEPath = 
-# 0 turns DMD window off, 1 turns DMD window on (Default is 1)
-AltSound = 
-AltColor = 
-# 0 to show grill (if it exists), 1 to hide grill
-B2SHideGrill = 
-# 0 to show extra DMD frame (if it exists), 1 to hide extra DMD frame
-B2SHideB2SDMD = 
-# 0 to show backglass, 1 to hide backglass
-B2SHideB2SBackglass = 
-B2SHideDMD = 
-B2SDualMode = 
-# 0 to hide all the B2S windows (grill, DMD frame, backglass), 1 to show them according to the other settings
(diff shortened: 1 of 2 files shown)


jsm174:
This looks pretty neat, but I've noticed several errors, and I've only look at the first few:

```
RenderingModeOverride = 
; Force specific rendering mode
; Default: 2 (for mobile/embedded version)
; Values: 0 = automatic, 1 = force software, 2 = force hardware
; Info: Overrides automatically detected rendering mode
```

```
TableListSort = 
; Table list sorting criteria
; Default: 0 (sort by name)
; Values: 0 = name, 1 = modification date, 2 = author
; Info: Determines table display order
```

```
LiveUIOverride = 
; Force mobile/touch interface
; Default: 0 (automatic detection)
; Values: 0 = auto, 1 = force mobile interface
; Info: Enables touch interface even on desktop
```

If these were generated with AI assistance, we'd really have to go through carefully.

Le-Syl21:
Yes, done  with an AI. Got assets/Default_VPinballX.ini and core/Settings.cpp
What are false in your exemple ? The description ?

Le-Syl21:
I've found "TableListSort":

This setting stores how the list of tables are sorted when displayed to the user. A value of 1 sorts the tables in alphabetical order while 0 sorts them in reverse alphabetical. I will correct it.

I know it's not perfect but today a large part of config is not understandable by anyone. It has to be correct on time, lot of values are not use anymore on 10.8.1

It's just the first step, a good start to make a great documented configuration

Le-Syl21:
I found it here: https://github.com/dekay/vpinball-wiki/wiki/VPinball-Ini but a lot of options are unknow

This option is only for mobile.
I can't test it or found any information on the sources

vpinball/src$ grep -ri "TableListSort" *
assets/Default_VPinballX.ini:TableListSort = 
assets/Default_VPinballX_Fra.ini:TableListSort = 

So can we push it and ask maintainer to take a look on the options they have implement or using

Le-Syl21:
I will rework it

vbousquet:
Hi! We definitely would benefit from a more user friendly config file. There is an ongoing effort to do so through the `Validate` method of the `Settings` class: https://github.com/vpinball/vpinball/blob/6c30902c8d52da46ab7e18e61c2ec288ea383d61/src/core/Settings.cpp#L105

As described in the comment, this is still incomplete (the idea is to move settings little by little) but the idea is to have an integrated way to guarantee settings validit/documentation. In the end, the idea is that all the setting docs should be added there, with setting value validation, and then the default ini file I introduced in 10.8 will be removed and generated from code instead. This will guarantee that things are always up to date, and also that wrong setting file won't crash the app (this si what happens atm for lots of settings if fed with invalid values).

The base implementation is english only though. But we could eventually extend this.

PR #2625, opened 2025-08-15
plugins: update B2SLegacy to poll PinMAME for some tables (BTTF)

jsm174:
This fixes the issue reported by @superhac where segments were not displaying in the new B2SLegacy plugin for some games like bttf. 

Somewhat related to https://github.com/vpinball/vpinball/pull/1887


superhac:
I tested against BTTF and Mata Hari which had this specific issue.  All works great now!  Thank you very much!

PR #2639, opened 2025-08-17
plugins: simplify DMDOverlay to be added to base form. Fix autopos.

jsm174:
This PR fixes autopos for the DMDOverlay for the `B2SLegacyPlugin`

I didn't have much choice but to added the window's width and height to the `VPXRenderContext2D`. 

On MacOS, if a person defines the backglass window as 300x300, since it's retina, the buffer is 600x600. If they are manually trying position it, and say, leave a 20 pixel gap on each side, they might do:

```
[Backglass]
BackglassDisplay =
BackglassFullScreen = 0
; Output mode
; 0 - Disabled
; 1 - Embedded in playfield output
; 2 - Native system window (maybe a window, exclusive fullscreen, VR headset,...)
BackglassOutput = 2
BackglassWndX = 0
BackglassWndY = 0
BackglassWidth = 300
BackglassHeight = 300

[Plugin.B2SLegacy]
.
.
BackglassDMDOverlay = 1
BackglassDMDAutoPos = 0
BackglassDMDX = 20
BackglassDMDY = 20
BackglassDMDWidth = 260
BackglassDMDHeight = 260
```

They aren't not going to realize it got upscaled to 600x600 so it would be offset.

I've tested on mobile as well, and everything is working well.

Screenshot 2025-08-17 at 2 36 52 PM

### DIFF
--- a/plugins/b2slegacy/Server.cpp
+++ b/plugins/b2slegacy/Server.cpp
@@ -26,10 +26,7 @@ Server::Server(MsgPluginAPI* msgApi, uint32_t endpointId, VPXPluginAPI* vpxApi)
    : m_msgApi(msgApi),
      m_vpxApi(vpxApi),
      m_endpointId(endpointId),
-     m_pinmameApi(nullptr),
-     m_resURIResolver(*msgApi, endpointId, true, false, false, false),
-     m_scoreviewDmdOverlay(m_resURIResolver, m_dmdTex, nullptr, vpxApi),
-     m_backglassDmdOverlay(m_resURIResolver, m_dmdTex, nullptr, vpxApi)
+     m_pinmameApi(nullptr)
 {
    m_pB2SSettings = new B2SSettings(m_msgApi);
    m_pB2SData = new B2SData(this, m_pB2SSettings, m_vpxApi);
@@ -61,9 +58,6 @@ Server::Server(MsgPluginAPI* msgApi, uint32_t endpointId, VPXPluginAPI* vpxApi)
    m_nLamps = 0;
    m_mechIndex = -1;
    m_nMechs = 0;
-
-   m_scoreviewDmdOverlay.LoadSettings(msgApi, "B2SLegacy"s, "Scoreview"s);
-   m_backglassDmdOverlay.LoadSettings(msgApi, "B2SLegacy"s, "Backglass"s);
 }
 
 Server::~Server()
@@ -1614,7 +1608,7 @@ void Server::Startup()
 void Server::ShowBackglassForm()
 {
    if (!m_pFormBackglass)
-      m_pFormBackglass = new FormBackglass(m_msgApi, m_vpxApi, m_pB2SData);
+      m_pFormBackglass = new FormBackglass(m_vpxApi, m_msgApi, m_endpointId, m_pB2SData);
 
    m_pFormBackglass->Show();
    m_pFormBackglass->SetTopMost(true);
--- a/plugins/b2slegacy/Server.h
+++ b/plugins/b2slegacy/Server.h
@@ -5,7 +5,6 @@
 #include "forms/FormBackglass.h"
 #include "classes/B2SCollectData.h"
 #include "core/ResURIResolver.h"
-#include "utils/DMDOverlay.h"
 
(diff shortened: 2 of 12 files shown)


vbousquet:
I will have a look but I think adding these additional coordinates is very likely wrong. All ancillary rendering is done in a virtual surface which virtual size is defined by srcWidth/srcHeight. This defines the virtual coordinate system used for rendering. So this coordinate system is unrelated to anything physical and it must stay this way as ancillary window will also be rendered in 3D for VR mode.

As srcWidth/srcHeight are purely a custom virtual coordinate system for a given render call, they may be adjusted by the plugin to the virtual size it expects. This is what all other plugins do (PUP, new B2S, Scoreview) and I guess that's what legacy B2S plugin should do also.

jsm174:
Okay. If it's not there, then there needs to be a way for a plugin to be able to query it's requested size so that the end user can logically assign values. They shouldn't have to know anything about MacOS (or wayland) converting sizes to something else.

Take this example:

```
[ScoreView]
ScoreViewDisplay =
ScoreViewFullScreen = 0
; Output mode
; 0 - Disabled
; 1 - Embedded in another output
; 2 - Native system window (maybe a window, exclusive fullscreen, VR headset,...)
ScoreViewOutput = 2
ScoreViewWndX = 0
ScoreViewWndY = 0
ScoreViewWidth = 500
ScoreViewHeight = 300
Priority.B2SLegacyDMD = 2
Priority.PUP = 1
Priority.ScoreView = 0
```

Screenshot 2025-08-18 at 7 19 00 AM

Since autodetection doesn't work for this at the moment, they would have to put these values in:

```
ScoreviewDMDOverlay = 1
ScoreviewDMDAutoPos = 0
ScoreviewDMDX = 12
ScoreviewDMDY = 12
ScoreviewDMDWidth = 476
ScoreviewDMDHeight = 118 
```

And to me this makes sense, width and height are smaller than 500x300
  
Screenshot 2025-08-18 at 7 21 25 AM

I don't think we can expect the end user to know that internally their window is 1000x600 and then have to use values like:

```
ScoreviewDMDOverlay = 1
ScoreviewDMDAutoPos = 0
ScoreviewDMDX = 24
ScoreviewDMDY = 24
ScoreviewDMDWidth = 952
ScoreviewDMDHeight = 236 
```

vbousquet:
Sorry, I was not clear. I entirely agree that the user has nothing to do with the rendering output. What I mean is that the plugin does not either. This is managed by VPX, not the plugin. For this to work, the plugin has to declare its expected rendering size, for example like this (from new B2S, but you will find the same for PUP or ScoreView):
```
bool B2SRenderer::RenderBackglass(VPXRenderContext2D* ctx)
{
   ctx->srcWidth = m_b2sWidth;
   ctx->srcHeight = m_b2sHeight;
```

srcWidth/srcHeight automatically defaults to the render surface output size, and this seems to be wrong on MacOS but I think this is another problem. In this situation where the plugin knows the render size it expects, it should define it and act accordingly.

jsm174:
ah okay. I think I might understand. I've never touched srcWidth and srcHeight before. They always defaulted to whatever outWidth and outHeight. I guess I have more refactor work then. 😄

vbousquet:
Edit: `srcWidth/srcHeight automatically defaults to the render surface output size, and this seems to be wrong on MacOS but I think this is another problem.` => in fact, no these values are right but I understand that these values are not user friendly as user expects to see coordinates that do not correspond to real screen coordinates on MacOs device (thanks Apple for this incredibly loosy design...)...


So you are right that if you want to express the coordinates not relative to the B2S but to the user screen, things will be complicated. But I would avoid doing this as this is not stable across users (can't share), desktop (breaks when user change his settings) and use (VR vs desktop for example).

vbousquet:
Thinking about it again, and knowing the way legacy B2S works, that is to say resize all image to expected output size, then flipbbok of them, I would say that using the target render width/height as srcWidth/srcHeight is the right choice (as images will be resized to this target). So the way it is done right now seems good to me. What should be changed is the way things are exposed to user: the user value should be in the background image cooridnate space, then converted to this render space by the simple formula: `render pos = render size * user pos / background image size`

PR #2655, opened 2025-09-01
Changed SDLK_PLUS to SDLK_EQUALS for volume up for liveui

### LINKED ISSUE #2642: Standalone: Volume Up ("=") not working
**Describe the bug**
While the volume down key("-") works the volume up key ("=") no longer does.

**To Reproduce**
1. Open any table and press "=". Doesn't not show the volume UI. 
2.Try volume down ("-") and you'll see the UI pop up and the volume goes down.

**Versions & Settings**
 - VPX: Lastest Master Linux BGFX

KutsuyaYuki:
This fixes the handling of the "=" sign on a keyboard that is used for the volume up for the liveui.
Tested on Arch Linux, x64, BGFX.

Fixes #2642


superhac:
@KutsuyaYuki This indeed fixes the problem!  Good catch and thanks!

vbousquet:
Thanks !

PR #2669, opened 2025-09-09
FlexDMD misc fixes

jsm174:
This PR has the following FlexDMD fixes/updates:

- Fixes crashes when `NewImage` and `NewVideo` return `nullptr`. VBScript needs to see this as `Nothing`.  This fixes the new GOT 1.2 from script erroring since the logo gif is not present:

```vbs
Function NewSceneWithVideo(name,videofile)
        'Debug.Print "NewSceneWithVideo|Group:"&name&",Video:"&videofile
    Dim actor
    Set NewSceneWithVideo = FlexDMD.NewGroup(name)
    Set actor = FlexDMD.NewVideo(name&"vid",videofile & ".gif")
    If actor is Nothing Then
                Debug.Print "NewSceneWithVideo - Falling back to image: "&name&",File:"&videofile
                FlexDMD.LockRenderThread
        Set actor = FlexDMD.NewImage(name&"img",videofile&".png")
                FlexDMD.UnLockRenderThread
        if actor is Nothing Then
            debug.print "Warning: "&videofile&" image not found"
            Exit Function
        End if
    End If
    NewSceneWithVideo.AddActor actor
End Function
```
 
- Fixes UltraDMD from not displaying animated gifs, as the parameters were reversed somewhere along the line when ported to the plugin. This can be tested with Kiss, (fwiw, I think the colors need to be orange)

image

- Adds support for decoding videos using ffmpeg. Some tables such as JP's Diablo, the videos are all `.wmv`. The past fix was to convert them to `.gif` and then hack the script.  @vbousquet / @toxieainc  - I didn't update any vc project files to include the ffmpeg libs. Also, I wasn't quite sure how to use the new libav.h approach.

Also before we get reports of not seeing the DMD with Diablo and `B2SLegacy` plugin, this one can not do auto dmd positioning because there is no black rectangle:

Screenshot 2025-09-08 at 8 02 54 PM
  
Screenshot 2025-09-08 at 8 03 30 PM

- Miscellaneous extra whitespace removal.

### DIFF
--- a/make/CMakeLists_plugin_FlexDMD.txt
+++ b/make/CMakeLists_plugin_FlexDMD.txt
@@ -129,19 +129,34 @@ if(BUILD_SHARED)
          $<$<CONFIG:RELEASE>:/OPT:REF>
          $<$<CONFIG:RELEASE>:/OPT:ICF>
          $<$<CONFIG:RELEASE>:/LTCG>
+         /SAFESEH:NO
       )
       if (PluginArch STREQUAL "x64")
          set_target_properties(FlexDMDPlugin PROPERTIES
             OUTPUT_NAME "plugin-flexdmd64"
          )
          target_link_directories(FlexDMDPlugin PUBLIC ${CMAKE_SOURCE_DIR}/third-party/build-libs/windows-x64)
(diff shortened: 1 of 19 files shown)


superhac:
This works great!  Much appreciated!  

dmd

PR #2674, opened 2025-09-11
ci: fix pipeline to build with supported windows runners

jsm174:
10.8.0 pipelines are failing. This fixes that.

Unfortunately this prevents us from using `v141_xp` flag. Not sure if this is an issue @toxieainc @JockeJarre ...

### DIFF
--- a/.github/workflows/prerelease.yml
+++ b/.github/workflows/prerelease.yml
@@ -12,7 +12,7 @@ env:
 
 jobs:
   version:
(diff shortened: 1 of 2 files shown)


JockeJarre:
I'd say that Windows XP had its LAST extended support 2014 according to Microsofts webpage.
If someone isn't interested in upgrading the OS, why be interested in upgrading VPX?
Sometime it is time to let go...

Thanks Jason

PR #2694, opened 2025-09-15
BGFX: fix incorrect mipmap leading to failed DX11 vs Vulkan test

vbousquet:


Augusto7743:
Thanks for doing it.
Hey ... VPX BGFX require what Vulkan minimum version ? 1.2 ?

PR #2745, opened 2025-10-02
fix: plugin ordering

### LINKED ISSUE #2744: B2SLegacyPlugin `B2S.Server` registration order issue
**Describe the bug**
Linux (Ubuntu 25.10)
Master branch

I have a clean install with default config file
When launching a table with B2S file I get the following error:

```
2025-10-02 11:42:21.365 WARN  [92157] [external_create_object@915] Creating an object of type "B2S.Server" is not supported
2025-10-02 11:42:21.365 ERROR [92157] [CodeViewer::OnScriptError@1187] Script Error at line 107 : VBSE_CANT_CREATE_OBJECT
2025-10-02 11:42:21.365 ERROR [92157] [CodeViewer::AppendLastErrorTextW@3839] Runtime error: Line: 107, Character: 2, VBSE_CANT_CREATE_OBJECT
```

I went through the debugger and the issue seems to be the delayed registration of the plugin because the `Controller` class is not yet available

https://github.com/vpinball/vpinball/blob/3d7c7a000760c28553196385ea0582788d52b3d9/plugins/b2slegacy/B2SLegacyPlugin.cpp#L174-L175

https://github.com/vpinball/vpinball/blob/3d7c7a000760c28553196385ea0582788d52b3d9/plugins/b2slegacy/B2SLegacyPlugin.cpp#L188-L189

This is the order of things:

* plugin is registered but the com object override is delayed because `Controller` is not defined
* script tries to create a `B2S.Server` object
* script error or table not showing b2s
* delayed registration kicks in and `Controller` is defined now (too late) 

What registers `Controller` and why after the plugin is registered?

francisdb:
Fixes plugin loading on linux by making sure the pinmame plugin is loaded first

closes #2744 (dir rename workaround does not work)


francisdb:
I suggest hiding whitespace on github diff view to evaluate this one.

francisdb:
The mobile app has a different solution (but less flexible)

https://github.com/vpinball/vpinball/blob/f3618fdbffbede0a4f9dfb55ab5deeb522a9e0ba/standalone/VPinballLib.cpp#L108-L127

evilwraith:
Can confirm this works on Linux Standalone.

jsm174:
I'll defer to @vbousquet on this. 

This solution does work, however I know the plugin system is still early days and needs a few more things: dependency system, proper shutdown, etc.

francisdb:
So we are stuck without a solution on Linux? Renaming files did not fix the issue for me, does not change iteration order. Why can't this be a temporary commit that is removed later on when things are set up properly?

vbousquet:
As described in the review, the proper clean up is being worked on at the moment. I don't see the point of merging a hack on a development branch just to remove it a few days after.

francisdb:
@jsm174 do you know if has been fixed meanwhile?

jsm174:
> @jsm174 do you know if has been fixed meanwhile?

Hello. It is not fixed yet.

Issue #2305, opened 2025-04-06
Rework input to be full SDL


Le-Syl21:
I want a meeting with @vbousquet to defined a SDL input rework.

Can you please reply and I can give you my mail or phone or discord...

vbousquet:
It's difficult for me to plan a meeting in the coming days. Perhaps could we exchange on the problem the current implementation has and how they could be solved ?

I tried to write a review of the PR you proposed here: https://github.com/vpinball/vpinball/pull/2277#issuecomment-2772764012

Leading to the proposed changes:
- [ ] remove #ifdef SDL in Settings
- [ ] use a default mapping value that lead to an unexisting button in Settings
- [ ] implement esc/Enable3D as you did but using FireAction
- [ ] implement missing action mappings (need to be added to action) but no urgency, keep comments until it is done
- [x] fix the missing test on button override for existing controller input layout
- [ ] exchange and define a way to move forward for GamePad which would include UI integration

Since then, I fixed point 5, and I had a look to the code and the default mapping 0 seems quite good to me as it indeed corresponds to an unmapped state (valid mapping starts at 1), so I would say point 2 is also solved.

It is not totally clear for me which problem you are trying to solve. In fact, I do not entirely understand why we would want to use SDL Gamepad instead of SDL joystick since SDL gamepad is just a mapping layer over SDL joystick which does not support all pinball interfaces and is somewhat duplicated to the mapping VPX applies.

Could you elaborate on what are the problems you see with the current implementation and the solution you propose ?

vbousquet:
To move forward, before coming with a solution, perhaps we could try to write down the issues we have them to be sure the solution we build encompass all of them.

So far, this is what I perceive:
1. we have multiple input layers with likely some duplicates that could be cleaned up (for example, SDL includes DirectInput and XInput joystick, and could be likely also used for mouse/keyboard, but we are still providing them separately)
2. we need to interface with a wide range of hardware: gamepads, generic controllers, keyboard encoders, wireless, specific OpenPinDev devices, ,... with digital and analog inputs (nudge acceleration and/or speed, same for plunger)
3. we display raw input naming which are not user friendly nor cross platform (platform dependent button and axis number instead of button/axis name corresponding to real controller naming)
4. on Windows, people expect VPX to also drive external applications like VPinMame. This is somewhat supported but not cleanly, with partial input / action mapping (before 10.8.1, since arch will be different with plugins, more in line with standalone implementation).
[...]

Issue #2676, opened 2025-09-11
OpenPinDev not working anymore in 10.8.1


jueank:
I'm using the 10.8.1 build for Windows 64-bit, with BGFX from the daily GitHub action runs.
I also have a PinscapePico board connected for buttons, the plunger, and nudging/accelerometer input.

Since 10.8.1, the OpenPinDev devices no longer work.
Previously, I had the two flipper buttons configured as OpenPinDev inputs instead of sending keystrokes. This no longer works, so I had to switch back to sending keys.
The plunger and nudging also stopped working.

To be fair, I didn’t dig too deeply into the plunger and nudging issue, since the VPX settings dialog for those features is quite confusing. The options aren’t very self-explanatory, and I usually forget after a few months which combination of checkboxes is supposed to work together.

This issue might be a candidate for @mjrgh

jueank:
Btw. why are there 4 PinscapePico listed in the sub-dialog of the settings dialog in VPX?

<img width="337" height="296" alt="Image" src="https://github.com/user-attachments/assets/a0138d83-8fe3-4d66-8aa4-27006d8d97bd" />

mjrgh:
> This issue might be a candidate for @mjrgh

It's a regression, so it'll be more efficient for whoever broke it (presumably whoever's been working on the input or USB code) to investigate and fix it, while the breaking changes are still fresh in their mind.  ("But I don't have a way to test it":  Easily solved.  Grab a Pico (they're $4 apiece; everyone really ought to be keeping a drawer full of them on hand), load [Pinscape Pico](https://github.com/mjrgh/PinscapePico/) onto it, and install [this configuration](https://github.com/mjrgh/PinscapePico/blob/main/GUIConfigTool/ConfigTemplates/OpenPinDevTester.json).  That will generate continuous alternating button presses on the flipper buttons on a 2-second cycle, and continuous full-range sine waves on the nudge and plunger axes, so it will provide continuous, predictable input without having to wire anything to the Pico and without the need for any user action.  The OPD HID is the only one it will expose, so it won't create any conflicts with existing joysticks or other devices.)

vbousquet:
@jueank Could you test with the latest nightly build ? I would expect this issue to be solved. Note that this version will need you to setup your inputs (through the in game UI) to work well

jueank:
@vbousquet Absolutely!
[...]

PR #2748, opened 2025-10-05
New input

vbousquet:
@toxieainc @jsm174 @Le-Syl21 This is a (more or less) full rewrite of the input API with aquite a few fixes and new features which match what was discussed here: https://github.com/vpinball/vpinball/issues/2305#issuecomment-2925230768

Taking from the updated changelog:
- selecting between an input API is no more needed, SDL is now used on all platforms (on Windows it supports simultaneously all XInput & DirectInput devices),
- the user interface to define inputs has been moved to the in-game user interface
- it is now possible to use a combination of multiple inputs for an action (for example pressing 2 buttons simultaneously)
- it is now possible to define multiple bindings per action, for example a keyboard key as well as 2 different joystick buttons
- staged flippers are now always implemented with the default binding mapping both lower and upper flippers to the same button
- using joystick stick/triggers for buttons is supported, for example allowing to map staged flipper at different trigger thresholds
- joystick 'hats' can be used as buttons, each direction being considered as an independent button
- nudge now also support position sensors, allowing to use a joystick to nudge
- more keyboard layouts are now supported (japanese keyboards, ...)
- new actions have been added, for example for adjusting the VR scene position
- the option to disable the 'Escape' key has been removed and replaced by a mapping like all other actions (it is used to open interactive exit)
- all key events are now broadcasted to the table's script, separately from the named action event (left flipper, right magna save, ...)
- on Windows platform, the presence of xinput.dll is no more needed, allowing to run VPX on Windows installs without the XBox Gaming Subsystem
- on Windows platform, the accessibility shortcuts are now fully disabled when playing
- on Windows platform, the Windows keys can now be used as a normal key input
- a few hardware devices and generic layouts are automatically detected, when this happens, the user is proposed to apply the corresponding default mapping. Support for some very old hardware has been removed (Microsoft's Sidewinder, UltraCade's UltraPin,...)
- the hardware acceleration filter may no more be enabled for keyboard nudge (as it is not supposed to be used for keyboard nudge)

  The new input setup is stored along the old one to allow keeping older versions of VPX together with newers.
  This change requires users to redefine their input settings in the new UI.

I'm pretty happy with how the new UI helps setup & understanding what is wrong. This also brings playing on a computer with an xbox like controller a level higher thanks to stick nudging and nice staged flippers.

The change being very large without real testing capabilities, regressions are to be expected. They will be solved as they are tested/reported.

As much as possible, I always try to perform the refactoring/cleaning step by step with small enough commits. Sadly, this was not possible here and this is likely the biggest PR I have ever made on this project, making it somewhat difficult to review. Sorry for that.

Also, the Remote Control plugin is broken with this update but the rework is in progress.


vbousquet:
@mjrgh I have included quite a few changes here, including an update of OpenPinDev support.
This contains also a follow up to this: https://github.com/vpinball/vpinball/commit/3d7c7a000760c28553196385ea0582788d52b3d9 with new explicit derivative functions.

I you have some time and will to spend on this, here are a few things to look for:
- I have moved the first/second derivative for position/velocity to acceleration out of the physics engine into clean classes that allow more easily testing/debugging. I have also added a complete UI that allows to better understand the results (inside the In Game UI, in nudge and plunger). This lead to a slightly different implementation for the first derivative: you were using a simple delta between 2 last value, causing acceleration 'spikes' (after each USB communication making physics to depend on the USB data flow), I moved to a continuous derivative. I tested with my Pinscape and it seems nice to me, but this is very subjective. What do you think ?
- I have updated the OpenPinDev to be fully integrated in the new device scheme. That is to say devices are identified and propose a default mapping that the user may or may not choose to apply. Since I don't have any OpenPinDev, this was not tested with any hardware.
- I have updated the way the OpenPinDev reports are processed as they used to only apply the last report state which in certain situations could lead to missing a fast press (still very unlikely, but seems more clean to me)
- I have included support for Pinscape device as I have one, it would be nice and easy to add support for your new PinscapePico if it is not already included through the OpenPinDev support

I also expect this PR to have solved: https://github.com/vpinball/vpinball/issues/2676

PR #2885, opened 2025-11-15
quest: add support for meta quest builds

jsm174:
This adds two build flavors to the android app: `mobile`, `quest`.

Before this can be merged, we need to get a patch into our bgfx fork.


vbousquet:
This is really awesome!

I have merged the changes to our BGFX branch and updated VPX to point to the updated commit.

PR #2950, opened 2025-12-09
Update build scripts to Visual Studio 2026

rockfordroeNG:
The build instructions state Visual Studio 2026 is needed yet the dependency build scripts still use Visual Studio 2022.


jsm174:
Thanks for this!

None of the dependent projects are building with your PRs, probably because of the Windows runners at still at an older cmake. We will have to wait until they get fixed before merging this.

toxieainc:
There is no need to switch yet. The building instructions just mention VS2026, because its the easiest to find/download.
The first windows build line mentions VS2019 and above.

toxieainc:
I also just clarified that in the line of the readme that could have been misinterpreted.

rockfordroeNG:
I also filed this PR because CMake 4.2.0 would not detect installations of VS 2026 unless the generator was set to that string.

vbousquet:
The external build scripts have been moved to VS2026 as part of https://github.com/vpinball/vpinball/commit/6f6ce03b3a9d551631481a0a23d1245dc6a551e3

The MSVC project have also been updated to fully support MSVC2026 as part of https://github.com/vpinball/vpinball/commit/d980f15bbeb80bc4ca7a7f998fd8992e94a3b0f2

Therefore I will close this PR as it doesn't seem to be needed anymore.

PR #2973, opened 2025-12-28
Added option to enable visual shake with hardware nudge

AmigaAbattoir:
This PR adds optional settings to apply a camera shake when using hardware sensors like Pinscape. It defaults to off so it shouldn't affect anyone until they turn it on.

I put a Pinscape in an arcade controller that I used connected to my computer, this way when I'm playing I can get some sort of sense that the table is moving on screen.


vbousquet:
This is a very nice proposal! I have just added it to VPX but implemented differently. The problem with your proposal is that it would apply an additional physics nudge, so I wrapped up a quick version that implements it by computing a visual nudge without touching the actual physics. Thanks a lot for the idea and poroposal!

PR #2975, opened 2025-12-29
Fix the DOF plugin

jermatic1:
I'm definitely no expert with DOF but the current plugin wasn't working. It was comparing the current state with a resized vector value but I couldn't find anywhere the values were actually set.

There is some duplication and this could be simplified but it's at least working for me on AFM now. I confirmed flipper, start, and launch lamps and flipper, slingshot, and shaker solenoids.

My strobe light isn't working so made this a draft

### DIFF
--- a/plugins/dof/DOFPlugin.cpp
+++ b/plugins/dof/DOFPlugin.cpp
@@ -126,53 +126,73 @@ static void PollThread(const string& tablePath, const string& gameId)
    assert(pDOF != nullptr);
    SetThreadName("DOF.PollThread"s);
    pDOF->Init(tablePath.c_str(), gameId.c_str());
-   bool isInitialState = true;
-   vector<bool> wireStates, solStates, lampStates, giStates;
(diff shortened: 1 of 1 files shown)


jsm174:
FWIW, the current libdof plugin for 10.8.1, only handles pinmame events right now. 

It is not receiving events from the B2S/B2S Legacy plugins:

'E' - MyB2SSetData
'B' - MyB2SSetScore
'C' - MyB2SSetScorePlayer 

It don't think it receives events from the PUP plugin either.

jermatic1:
I figured out the strobe issue https://github.com/vbousquet/pinmame/pull/1 but it still didn't work after the fix. The value of the strobe sol is less than 0.5f when it's on.  Not sure if that should be accounted for in pinmame or here yet.

jsm174:
I think this can be closed as DOF was recently fixed in 10.8.1. 

https://github.com/vpinball/vpinball/pull/3242

PR #3008, opened 2026-01-11
Standalone: Fix -listres/-listsnd and add -listctrl/-listall options

Le-Syl21:
## Summary
- Fix `-listres` and `-listsnd` commands that were not working (SDL not initialized)
- Add `-listctrl` to list controllers with detailed info (axes, buttons, vendor/product, serial)
- Add `-listall` to list all devices at once
- Display platform and SDL version in list output

## Test plan
- [ ] Run `./VPinballX_BGFX -listres` and verify displays are listed
- [ ] Run `./VPinballX_BGFX -listsnd` and verify sound devices are listed
- [ ] Run `./VPinballX_BGFX -listctrl` and verify controllers are listed with details
- [ ] Run `./VPinballX_BGFX -listall` and verify all devices are listed

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/core/VPApp.cpp
+++ b/src/core/VPApp.cpp
@@ -243,6 +243,8 @@ static const string options[] = { // keep in sync with option_names & option_des
    "Audit"s,
    "listres"s,
    "listsnd"s,
+   "listctrl"s,
+   "listall"s,
    "CaptureAttract"s,
    "c1"s,
    "c2"s,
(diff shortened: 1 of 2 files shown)


vbousquet:
This PR fixes and extends the -list... commands, but these commands were created to provide a quick hack to solve the missing user interface on other platform than Windows. We have worked against this core issue and now all platforms are meant to be setup through the common user interface. So, what I have planned is to remove these commands (and other legacy commandline dating from before the improvment of the setting system) before the next release (as they were not part of any release so far) avoiding all the drawback they would add to VPX (overall worse user experience, harder to understand commandline, more docs, moving away from a clear unified way of using the app across platform, manual modification of settings files,...).

Le-Syl21:
Got it, thanks for the explanation! 👍
I'll close this PR, and following the same logic, I'll also close #3009 (Standalone: Improve -displayid command for multi-display setups) since it falls into the same category of legacy CLI commands.

PR #3055, opened 2026-01-22
Add coin door key to Sega/Stern WhiteStar

Herschel:
This adds support for the coin door key for Sega/Stern WhiteStar games, pending https://github.com/vpinball/pinmame/pull/546.


toxieainc:
Also thanks for that one here!

PR #3064, opened 2026-01-25
Readme update to compile on RPi

colas-sebastien:


colas-sebastien:
I added libraries and their dependencies to avoid to have a question during the libraries installation

PR #3116, opened 2026-02-08
Add tables directory configuration from CLI

Nicals:
On Linux systems where no Documents directory is configured (e.g. removed from xdg-user-dirs), SDL_GetUserFolder(SDL_FOLDER_DOCUMENTS) returns  nullptr.
This gets passed straight into std::filesystem::path(), causing a segfault at startup.

This PR fixes that null pointer crash with a proper error message, and adds a -TablesPath CLI parameter so users can explicitly set the tables  directory — which also serves as a workaround for systems without a Documents folder.

I ran the application with these changes but could not fully validate the feature since I hit a second, unrelated crash in StandaloneMsgLoop::MainMsgLoop where m_vpxEditor is null.

### DIFF
--- a/src/core/AppCommands.cpp
+++ b/src/core/AppCommands.cpp
@@ -253,6 +253,7 @@ enum option_names
    OPTION_CUSTOM8,
    OPTION_CUSTOM9,
    OPTION_PREFPATH,
+   OPTION_TABLESPATH,
    OPTION_INVALID,
 };
 struct CommandLineOption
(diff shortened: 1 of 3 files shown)


vbousquet:
We already had and managed null doc folders on Android. I have extended the behavior to Linux. I would prefer not to add a new commandline option as this would be mostly a hack around a bug. With the change, it will start in the app folder (which is always defined) on first run, then will use last used folder. Therefore, I will close the PR. Thanks for your proposal!

PR #3121, opened 2026-02-09
FlexDMD: Implement score formatting with locale and commas

evilwraith:
Added functions to format scores with commas and locale support.

### DIFF
--- a/plugins/flexdmd/actors/ScoreBoard.cpp
+++ b/plugins/flexdmd/actors/ScoreBoard.cpp
@@ -1,4 +1,42 @@
 #include "ScoreBoard.h"
+#include <locale>
+#include <sstream>
+
+namespace {
+const std::locale& ScoreLocale()
+{
+   static const std::locale loc = [] {
+      try { return std::locale(""); } // Use user/system locale when available
+      catch (...) { return std::locale::classic(); }
+   }();
+   return loc;
+}
+
+std::string FormatWithCommas(int value)
+{
+   const bool negative = value < 0;
+   uint64_t n = negative ? static_cast<uint64_t>(-(int64_t)value) : static_cast<uint64_t>(value);
+   std::string s = std::to_string(n);
+   for (int i = static_cast<int>(s.size()) - 3; i > 0; i -= 3)
+      s.insert(static_cast<size_t>(i), 1, ',');
+   return negative ? "-" + s : s;
+}
+
+std::string FormatScore(int value)
+{
+   std::ostringstream oss;
+   oss.imbue(ScoreLocale());
+   oss << value;
+   std::string s = oss.str();
+   for (char& c : s)
+      if (static_cast<unsigned char>(c) == 0xA0) // NBSP grouping separator
+         c = ' ';
+   if ((value <= -1000 || value >= 1000) && s == std::to_string(value))
+      return FormatWithCommas(value);
+   return s;
+}
+}
+
 
 namespace Flex {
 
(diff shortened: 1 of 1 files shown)


vbousquet:
Hi and thanks for this proposal! The `ScoreBoard` class is a port from the old UltraDMD which did not support score formatting. If we merge this, it will break the layout of some of the existing tables that were designed with a text size without comma/dots. Still, I think this would be cool to have, but perhaps with a setting to allow disabling.

evilwraith:
> Hi and thanks for this proposal! The `ScoreBoard` class is a port from the old UltraDMD which did not support score formatting. If we merge this, it will break the layout of some of the existing tables that were designed with a text size without comma/dots. Still, I think this would be cool to have, but perhaps with a setting to allow disabling.

I saw that it looks different with the Freezy extensions on Windows and wanted to figure out why. On Windows, it actually has the comma separators. 

DMDEXT:
image

FlexDMD Plugin:
image

FlexDMD Plugin with patch:
Screenshot from 2026-02-10 09-50-31

vbousquet:
You are entirely right. The commas were there. They disappeared while porting to the plugin arch. Thanks!

jsm174:
FWIW, some background..

When I ported FlexDMD from C# to C++, I had:

https://github.com/vpinball/vpinball/blob/afc7c38ffb1078076ec523c37b30055061c93ff1/standalone/inc/flexdmd/actors/ScoreBoard.cpp#L91C1-L106C2
```
void ScoreBoard::SetScore(LONG score1, LONG score2, LONG score3, LONG score4)
{
   char score[20];
    
   snprintf(score, sizeof(score), "%'d", score1);
   m_pScores[0]->SetText(score);

   snprintf(score, sizeof(score), "%'d", score2);
   m_pScores[1]->SetText(score);

   snprintf(score, sizeof(score), "%'d", score3);
   m_pScores[2]->SetText(score);

   snprintf(score, sizeof(score), "%'d", score4);
   m_pScores[3]->SetText(score);
}
```

Looking at the C#, I did that because of:

https://github.com/vbousquet/flexdmd/blob/6357c1874e896777a53348094eafa86f386dd8fe/FlexDMD/Actors/ScoreBoard.cs#L96C1-L102C10

```
        public void SetScore(Int64 score1, Int64 score2, Int64 score3, Int64 score4)
        {
            _scores[0].Text = score1.ToString("#,##0");
            _scores[1].Text = score2.ToString("#,##0");
            _scores[2].Text = score3.ToString("#,##0");
            _scores[3].Text = score4.ToString("#,##0");
        }
```

`snprintf` seems to be much simpler then the proposed PR, however, doing some research the `"%'d"` isn't supported everywhere, and format in C++ 20, doesn't have support.

toxieainc:
Hi! Sorry, for being 'late' on this.. I just took the liberty to tweak this a bit, so this should be more similar to the original behavior.
It will not respect locale grouping anymore (e.g. Indian formatting), which maybe also makes more sense, as otherwise the planned layout for the score may be too much off.

evilwraith:
> Hi! Sorry, for being 'late' on this.. I just took the liberty to tweak this a bit, so this should be more similar to the original behavior. It will not respect locale grouping anymore (e.g. Indian formatting), which maybe also makes more sense, as otherwise the planned layout for the score may be too much off.

My thinking with adding locale is it seemed to be how it was intended to work, looking at old forum postings about the comma issue and the comma only appearing in locales that supported it. I will gladly bow to the wisdom of "don't break shit".

toxieainc:
The thousands separator will still be respected, so this stays exactly the same as in your PR. Its just the grouping, so it will always separate after every 3 digits (so no Indian 10,00,000 style for example).

Issue #3191, opened 2026-03-01
WMP Plugin:  Pathing Wrong


superhac:
**Describe the bug**
The pathing appears to be off in American Graffiti for the music folder.  Its point to the root of the music files:

```
2026-03-01 06:33:01.004 INFO  [3931732] [VPXPluginAPIImpl::PluginLog@239] WMP Core URL set to: /home/superhac/tables/American Graffiti (Original 2024)/music/AGSoundsXERBbumper1.mp3
2026-03-01 06:33:01.004 INFO  [3931732] [VPXPluginAPIImpl::PluginLog@239] Volume updated: 50 -> 0.50
2026-03-01 06:33:01.104 INFO  [3931732] [VPXPluginAPIImpl::PluginLog@239] WMP Core URL set to: /home/superhac/tables/American Graffiti (Original 2024)/music/AGAG4.mp3
2026-03-01 06:33:01.104 INFO  [3931732] [VPXPluginAPIImpl::PluginLog@239] Volume updated: 50 -> 0.50
2026-03-01 06:33:01.204 INFO  [3931732] [VPXPluginAPIImpl::PluginLog@239] WMP Core URL set to: /home/superhac/tables/American Graffiti (Original 2024)/music/AGSoundsXERBbumper1.mp3
2026-03-01 06:33:01.204 INFO  [3931732] [VPXPluginAPIImpl::PluginLog@239] Volume updated: 50 -> 0.50
2026-03-01 06:33:01.304 INFO  [3931732] [VPXPluginAPIImpl::PluginLog@239] WMP Core URL set to: /home/superhac/tables/American Graffiti (Original 2024)/music/AGAG23.mp3
2026-03-01 06:33:01.304 INFO  [3931732] [VPXPluginAPIImpl::PluginLog@239] Volume updated: 50 -> 0.50
2026-03-01 06:33:01.404 INFO  [3931732] [VPXPluginAPIImpl::PluginLog@239] WMP Core URL set to: /home/superhac/tables/American Graffiti (Original 2024)/music/AGSoundsXERBbumper1.mp3
2026-03-01 06:33:01.404 INFO  [3931732] [VPXPluginAPIImpl::PluginLog@239] Volume updated: 50 -> 0.50
``` 

Where as the music files are setup like this:
```
/home/superhac/tables/American Graffiti (Original 2024)/music/
├── AG
│   ├── AG10.mp3
│   ├── AG11.mp3
│   ├── AG12.mp3
[...]

Issue #3159, opened 2026-02-18
10.8.1-4685 BGFX-x64 Windows won't play music if music folder has spaces in the names


passionforpins:
**Describe the bug**
It seems to not like spaces in the folder names like Blood Machines OST, Charlie Brown or Tom and Jerry. I notice music folder without spaces all seem to work correctly and all work as expected in 10.8.0

vbousquet:
Not sure for which version this issue is reported. Just tested for 10.8.1 and all work fine (not that Blood Machine does not uses VPX for its music at all). I will close as the informations here are not sufficient to go any further

passionforpins:
10.8.1-4685 BGFX-x64 Windows
maybe it's a plugin related thing then.  At any rate it's a reliable finding for me that folders with spaces in the name will not play. I was asked by another user which is when I decided to investigate.

jsm174:
Could you maybe provide your log file?

Should be in 

`C:\Users\xxx\AppData\Roaming\VPinballX\10.8`

see:

https://github.com/vpinball/vpinball/blob/master/docs/FileLayout.md

passionforpins:
Sure thing

[vpinball.log](https://github.com/user-attachments/files/25396620/vpinball.log)

jsm174:
I see a bunch of things in the log. 

1) You could try disabling WMP plugin to see if your built in Windows Media Player has no issues. 

```
2026-02-13 10:16:19.976 INFO  [Main] [VPXPluginAPIImpl::PluginLog@239] WMP Core URL set to: D:\Visual Pinball\Music\Blood Machines OST\Carpenter Brut - BLOOD MACHINES OST - 05 Mima.mp3
2026-02-13 10:16:19.982 INFO  [Main] [VPXPluginAPIImpl::PluginLog@239] Volume updated: 50 -> 0.50
2026-02-13 10:16:41.123 INFO  [Main] [Player::~Player@752] Closing player...
```

If it couldn't find the file, we'd see an error here. It looks like it played for 20+ seconds though. (I never personally tried WMP plugin on Windows, but I know it works on linux and mac)

2) As for Tom And Jerry + Charlie Brown:

```
2026-02-17 18:13:56.473 INFO  [Main] [Player::Player@718] Startup done
2026-02-17 18:14:15.673 ERROR [Main] [ScriptGlobalTable::PlayMusic@134] Failed to stream music: Tom and Jerry\bgout_T&JbackgroundMusic0.mp3
[...]

PR #3192, opened 2026-03-01
misc: fix MusicDirectory returning path without separator

jsm174:
`MusicDirectory` wasn't returning paths with path separators after https://github.com/vpinball/vpinball/commit/a447f7c9460c53283d4ab11d70196c62b0c09b94. 

This would cause:

`/Users/jmillard/tables/batocera/userdata/roms/vpinball/ag/music/AGsfx_RadioScan.mp3`

instead of 

`/Users/jmillard/tables/batocera/userdata/roms/vpinball/ag/music/AG/sfx_RadioScan.mp3`

Also replaced `SysAllocStringLen` to `MakeWideBSTR` to match other functions.

Fixes https://github.com/vpinball/vpinball/issues/3191 and could be related to https://github.com/vpinball/vpinball/issues/3159


vbousquet:
Thanks!

PR #3276, opened 2026-03-31
Fix: X11 window placement on multi-monitor with GNOME/Mutter

Le-Syl21:
On Linux/X11, GNOME/Mutter repositions borderless windows that match the screen size, ignoring the requested position. This causes the playfield window to appear on the wrong display in multi-monitor pincab setups.

Fix: set SDL_HINT_X11_FORCE_OVERRIDE_REDIRECT before window creation on non-Windows platforms. This uses X11 override_redirect to bypass the window manager, ensuring all windows are placed at their requested coordinates.

Tested on Ubuntu 24.04 / GNOME / X11 with 3 monitors (4K + 1440p + 1080p). Verified: playfield, backglass and DMD all correctly placed on their assigned displays.

### DIFF
--- a/src/renderer/Window.cpp
+++ b/src/renderer/Window.cpp
@@ -213,8 +213,11 @@ Window::Window(const string& title, const Settings& settings, VPXWindowId window
       if (m_fullscreen)
          wnd_flags |= SDL_WINDOW_FULLSCREEN;
 
-      #if !defined(_MSC_VER) // Win32 (we use _MSC_VER since standalone also defines WIN32 for non Win32 builds)
-      // On Windows, always on top is not always respected and if using SDL_WINDOW_UTILITY windows may end up being hidden with no way to select and move them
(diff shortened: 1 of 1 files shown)


Le-Syl21:
Thx, I will push more for infobox size ans clicable ;-)

jsm174:
If your following up with another PR, I think there might be better ways to detect X11 than blindly setting it like was done here.

You should look at `RenderDevice.cpp`

```
   if (SDL_GetCurrentVideoDriver() == "x11"sv) {
   }
```

jsm174:
@Le-Syl21  - This was reverted as it broke other users running X11 including myself on Batocera. d81760fb47975a9433d9aadc93814345a0b24397

Le-Syl21:
OK does Batocera use X11 or WL ?

jsm174:
> OK does Batocera use X11 or WL ?

X11

PR #3324, opened 2026-04-13
macOS: fix ~2 FPS with Frame Pacing (SyncMode=3) due to broken Metal waitable swapchain

Raketemensch:
## Problem

On macOS with `SyncMode=3` (Frame Pacing), the game runs at ~2 FPS with severe input lag. The root cause is the recently-added BGFX Metal waitable swapchain support (commit cf910309c): `bgfx::waitForSwapchain()` calls `[CAMetalLayer nextDrawable]`, which blocks for ~1 second per call because presented drawables are never retired by the Core Animation compositor.

Log evidence:
```
waitForSwapchain=995775us
waitForSwapchain=996316us
FPS=2.90016 logic_frame=344808us sleep=315328us
```

## Fixes

**1. Disable waitable swapchain on macOS** (`RenderDevice.cpp`)

The Metal waitable swapchain is not yet functional. Disabling it on macOS causes Frame Pacing to fall back to VSync-based synchronization, which works correctly at 60fps.

**2. Replace `SDL_CreateRenderer` with direct `CAMetalLayer` creation** (`RenderDevice.cpp`, new `standalone/macos/MetalLayerHelper.mm`)

The original code used `SDL_GetRenderMetalLayer(SDL_CreateRenderer(..., "Metal"))` to obtain a `CAMetalLayer`. This creates an SDL Metal renderer that internally manages the drawable lifecycle, competing with BGFX for drawables on the same layer.

Replaced with a `CAMetalLayer` created directly on the NSWindow's content view. Additionally, AppKit requires view/layer manipulation on the **main thread** — the layer is now created in the `RenderDevice` constructor (main thread) and passed to the render thread. Creating it on a background thread causes it to not be registered with the display compositor, which is likely the underlying reason drawables were never being retired.

`MetalLayerHelper.mm` is an isolated ObjC++ file to avoid `BOOL` typedef conflicts between `<objc/objc.h>` and Wine headers.

**3. Throttle SDL event pump on macOS** (`player.cpp`)

`SDL_PollEvent` on macOS triggers the Cocoa run loop on every call. Throttle `SDL_PumpEvents()` to ~500Hz while continuing to drain the SDL event queue every iteration to avoid event accumulation.

**4. Skip `SDL_WINDOW_HIGH_PIXEL_DENSITY` on macOS** (`Window.cpp`)

Retina display doubled the Metal backbuffer resolution with no mechanism to scale it back, causing 4x the pixel fill rate.

## Testing

Tested on Apple M-series MacBook with `SyncMode=3`, `MaxPrerenderedFrames=1`. Before: ~2 FPS. After: stable 60 FPS.

### DIFF
--- a/make/CMakeLists_bgfx-macos-arm64.txt
+++ b/make/CMakeLists_bgfx-macos-arm64.txt
@@ -42,6 +42,7 @@ add_executable(vpinball
    src/input/OpenPinDevHandler.cpp
 
    standalone/macos/main.mm
+   standalone/macos/MetalLayerHelper.mm
 
    third-party/include/imgui/imgui_impl_sdl3.cpp
    third-party/include/imgui/imgui_impl_sdl3.h
--- a/src/renderer/RenderDevice.cpp
+++ b/src/renderer/RenderDevice.cpp
@@ -9,11 +9,12 @@
 
 #include <thread>
 
-#ifdef __LIBVPINBALL__
 #ifdef __APPLE__
 #include <pthread.h>
 #include <sys/qos.h>
+extern "C" void* VPX_CreateMetalLayer(void* nsWindow);
 #endif
+#ifdef __LIBVPINBALL__
 #endif
 
 #if !defined(DISABLE_FORCE_NVIDIA_OPTIMUS) && defined(ENABLE_DX9)
@@ -552,7 +553,14 @@ void RenderDevice::RenderThread(RenderDevice* rd, bgfx::Init init)
       bool gpuVSync = false;
       int framePacingFlushing = 0;
 
+      // Waitable swapchain on macOS Metal is broken: nextDrawable() blocks ~1s because
+      // presented drawables are never retired (recently added BGFX feature, not yet working).
+      // Disable it on macOS and fall back to VSync-based frame pacing.
+#if BX_PLATFORM_OSX
+      const bool waitableSwapchain = false;
+#else
       const bool waitableSwapchain = (bgfx::getCaps()->supported & BGFX_CAPS_WAITABLE_SWAPCHAIN) != 0;
+#endif
       if (waitableSwapchain)
          bgfx::waitForSwapchain();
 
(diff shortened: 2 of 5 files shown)


Raketemensch:
These changes make the game completely playable on an M1 Air.

Raketemensch:
I separated this into 2 PRs to hopefully make it easier for you, with the Metal changes in this one and the SDL changes in [the other](https://github.com/vpinball/vpinball/pull/3325).

jsm174:
Thanks for this PR.

As mentioned in #3325, I would like to get https://github.com/vpinball/vpinball/issues/3327 solved first as I think that is more related to current MacOS performance issues.

We may have to look into removing the waitable swap chain only on MacOS if that is causing problems for others, but I will need @vbousquet help on that. (It's better to fix in BGFX than hack vpx).

>2. Replace SDL_CreateRenderer with direct CAMetalLayer creation (RenderDevice.cpp, new standalone/macos/MetalLayerHelper.mm)

We pass the layer into BGFX, and the app does not use SDL for any rendering, so this "competing with BGFX for drawables on the same layer." doesn't seem right.

> Additionally, AppKit requires view/layer manipulation on the main thread 

I believe this was the case until #3327, which is why I'd like to see if this could get fixed first
 
> Skip SDL_WINDOW_HIGH_PIXEL_DENSITY on macOS (Window.cpp)

We definitely want HIGH_PIXEL_DENSITY.

HIGH_PIXEL_DENSITY

Screenshot 2026-04-14 at 6 35 23 PM

NO HIGH_PIXEL_DENSITY

Screenshot 2026-04-14 at 6 49 10 PM

You'll notice how jaggy everything is.  This change was probably your biggest performance boost because the buffer size much smaller.

Raketemensch:
No worries, it if helps it helps, if it doesn't, it doesn't.

Thanks for doing what you do.

vbousquet:
I have removed the buggy waitable swapchain impleùmentation from our BGFX flavor. This should solve the issue at the price of latency (like for GL backend). Proper frame pacing will need a Metal maintainer to actually implement it but this should wait for the frame pacing game loop to be fully stabilized (it still needs some adjsutments to be considered viable)

jsm174:
Tested the removal and everything is back to normal on mac builds.

PR #3325, opened 2026-04-13
macOS: throttle SDL event pump to reduce Cocoa run loop overhead

Raketemensch:
## Problem

On macOS, `SDL_PollEvent` triggers the Cocoa run loop on every call via expensive kernel traps. In the VPX game loop this is called hundreds of times per second (once per physics step at ~1000Hz), adding measurable CPU overhead.

## Fix

Separate `SDL_PumpEvents()` (which drives the Cocoa run loop) from the SDL event queue drain, and throttle the pump to ~500Hz. The SDL event queue is still drained on every iteration via `SDL_PollEvent` so `SDL_EVENT_POLL_SENTINEL` is handled correctly and no events are starved or accumulated.

Note: an earlier attempt used `SDL_PeepEvents` instead of `SDL_PollEvent` for the drain, which caused `SDL_EVENT_POLL_SENTINEL` events to accumulate in the queue (~2.6GB memory growth over ~1 hour). The current approach avoids this by keeping `SDL_PollEvent` for the drain.

### DIFF
--- a/src/core/player.cpp
+++ b/src/core/player.cpp
@@ -1462,7 +1462,22 @@ void Player::ProcessOSMessages(const bool isInitialized)
    static Vertex2D dragStart;
    static int dragging = 0;
 #ifndef __LIBVPINBALL__
+   // On macOS, SDL_PollEvent triggers the Cocoa run loop (expensive kernel traps) on every call.
+   // Throttle the OS-level pump to ~500Hz and drain the SDL queue cheaply in between.
+   // We must still drain via SDL_PollEvent (not SDL_PeepEvents) so SDL_EVENT_POLL_SENTINEL
+   // is handled correctly and doesn't accumulate in the queue.
+   #if defined(__APPLE__)
+   {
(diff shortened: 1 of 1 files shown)


jsm174:
Thanks for this PR. 

I would like to get https://github.com/vpinball/vpinball/issues/3327 solved first as I think that is more related to current MacOS performance issues.

Also, we may have to look into removing the waitable swap chain only on MacOS if that is causing problems for others, but I will need @vbousquet help on that. (It's better to fix in BGFX than hack vpx).

I will leave this open for now.

vbousquet:
> Also, we may have to look into removing the waitable swap chain only on MacOS if that is causing problems for others, but I will need @vbousquet help on that. (It's better to fix in BGFX than hack vpx).

The Metal "waitable swapchain" was removed (note that it was not really a waitable swapchain but a proxy for it)

johnstotler:
Should we close this out then?

jsm174:
Yes. New release reverts bgfx back to the non waitable swapchain.

PR #3330, opened 2026-04-16
core.vbs: guard mFastTimer access in EnableUpdate

francisdb:
`cvpmTimer.EnableUpdate` accesses `mFastTimer.TimerEnabled` unconditionally, but `mFastTimer` is only set when `InitTimer` is called with a `PulseTimer` element. Classes that register their `NeedUpdate` before `PulseTimer_Init` runs hit "Object required" (800a01a8), silently swallowed by the On Error Resume Next. Guard with `IsObject` so the error doesn't occur in the first place.


francisdb:
(wrong assumptions)

PR #3351, opened 2026-04-27
audio: let parallel plugin sources play alongside the active source

### LINKED ISSUE #3305: PUP plugin audio is silently dropped - missing audio source registration
## Summary

The PUP plugin sends audio data via `AudioUpdateMsg` / `CTLPI_AUDIO_ON_UPDATE_MSG`, but never registers itself as an audio source via `CTLPI_AUDIO_GET_SRC_MSG`. As a result, when PinMAME (or AltSound) is the active audio source, all PUP audio is silently discarded by the player's `OnAudioUpdated` filter.

## Steps to reproduce

1. Linux standalone, 3-screen cab (playfield, DMD, backglass), no topper
2. Load any table with a PUP Pack (e.g. Monster Bash `mb_106b`)
3. PUP videos render correctly on the backglass
4. PUP audio is completely silent — PinMAME audio works fine on the same backglass sound device

## Root cause

In `src/core/player.cpp` lines 2307-2311, `OnAudioUpdated` filters incoming audio messages:

```cpp
if (me->m_activeAudioSourceId == 0)
   me->UpdateActiveAudioSource();

if (me->m_activeAudioSourceId != 0 && msg.id.id != me->m_activeAudioSourceId)
   return; // <-- PUP audio is dropped here
```

`UpdateActiveAudioSource()` queries all plugins via `GetAudioSrcMsg`. PinMAME responds (registers with `overrideId=0` as root source). AltSound also responds (registers as override of PinMAME). **PUP never responds** — it does not subscribe to `CTLPI_AUDIO_GET_SRC_MSG`.

So `m_activeAudioSourceId` is set to PinMAME (or AltSound), and every `AudioUpdateMsg` from PUP fails the `msg.id.id != me->m_activeAudioSourceId` check and is silently ignored.

## Comparison

**AltSound** (works) — `plugins/altsound/AltSoundPlugin.cpp`:
- Subscribes to `CTLPI_AUDIO_GET_SRC_MSG` (line 254)
- Finds PinMAME as root source and sets `audioSrcDef.overrideId = pinmameAudioId` (line 86)
- Responds to `OnGetAudioSrc` with its `AudioSrcId` (lines 52-58)
- Result: `UpdateActiveAudioSource()` walks the override chain and selects AltSound as active

**PUP** (broken) — `plugins/pup/PUPPlugin.cpp`:
- Only gets `CTLPI_AUDIO_ON_UPDATE_MSG` to send audio (line 258)
- Never subscribes to `CTLPI_AUDIO_GET_SRC_MSG`
- Never declares an `AudioSrcId`
- Sends `AudioUpdateMsg` with valid PCM data that is never played

## Possible fix

The PUP plugin should register as an audio source, similar to AltSound. However, the current architecture only supports a single active source (override chain), while PUP audio should ideally play **alongside** PinMAME/AltSound (PUP is supplemental video/audio content, not a replacement for the game audio). This may require either:

1. Allowing multiple concurrent audio sources in the player, or
2. Having PUP register as an override of PinMAME and forward/mix PinMAME audio alongside its own

## Environment

- VPX v10.8.1 Beta (Rev. 4893 (e2488aa), linux BGFX 64bits)
- PUP plugin: `plugin-pup.so`
- Tables tested: Monster Bash (mb_106b), Demolition Man (dm_lx4)

Le-Syl21:
## Summary

Fixes #3305 — the PUP plugin's audio was silently dropped whenever PinMAME (or AltSound) was the active audio source.

The host's audio routing assumed a single active source resolved through the override chain. Any `AudioUpdateMsg` whose id did not match `m_activeAudioSourceId` was dropped at `player.cpp:2310`. PUP, however, is supposed to play *in parallel* with the ROM audio (video soundtracks / music / SFX layered on top), not replace it — so its samples were silently discarded for any ROM-based table.

## Fix

Two coordinated changes:

- **Player** — replace the single-active filter with an explicit "overridden set". `UpdateActiveAudioSource()` now collects every `src.overrideId.id != 0` value into `m_overriddenAudioSourceIds`. `OnAudioUpdated` drops messages whose id is in that set. Sources outside the chain (PUP) and the chain tip (PinMAME, or AltSound when overriding it) coexist. Streams of newly-overridden sources are closed when the chain changes; streams that leave the set reopen lazily on the next audio update.

- **PUP plugin** — register itself as an audio source via `CTLPI_AUDIO_GET_SRC_MSG` with `overrideId = 0`, satisfying the API contract and making PUP visible to other plugins. The umbrella registration is metadata only; PUP's per-video sub-streams keep allocating their own ids.

No API/ABI change to `ControllerPlugin.h`. AltSound and PinMAME are unchanged — AltSound still overrides PinMAME (PinMAME ends up in the overridden set, silenced).

## Behavior matrix

| Scenario | Before | After |
|---|---|---|
| PinMAME only | plays | plays |
| PinMAME + AltSound (override) | AltSound plays, PinMAME silenced | unchanged |
| PinMAME + PUP | PUP silenced (bug) | both play in parallel |
| PinMAME + AltSound + PUP | only AltSound | AltSound + PUP, PinMAME silenced |
| PUP only (no-ROM table, e.g. JP's Deadpool) | plays | plays (no regression) |

## Test plan

- [x] Build passes on all CI targets (linux-x64, windows x86/x64, macos arm64/x64, mingw, RPI/RK3588)
- [x] Manual test on Linux 3-screen pincab (Demolition Man `dm_lx4`): PinMAME audio + PUP music screen 4 (`MusicOnly`) play simultaneously
- [x] Manual test on JP's Deadpool (PUP-only, no ROM): no regression
- [ ] Cross-check on Windows builds with AltSound+PinMAME+PUP combo

### DIFF
--- a/plugins/pup/PUPPlugin.cpp
+++ b/plugins/pup/PUPPlugin.cpp
@@ -159,8 +159,24 @@ void DeleteTexture(VPXTexture texture)
 //
 
 static unsigned int onAudioUpdateId;
+static unsigned int onAudioSrcChangedId;
+static unsigned int getAudioSrcId;
 static vector<uint32_t> freeAudioStreamId;
 uint32_t nextAudioStreamId = 1;
+
+// PUP advertises itself as a parallel audio source (overrideId = 0) so the host
+// does not silence it when PinMAME or AltSound is the primary source.
+// PUP's per-video sub-streams allocate their own ids on top of this; this entry
+// is the umbrella registration that satisfies the CTLPI_AUDIO_GET_SRC_MSG contract.
+static AudioSrcId pupAudioSrcDef = {};
+
+static void OnGetAudioSrc(const unsigned int msgId, void* userData, void* msgData)
+{
+   GetAudioSrcMsg* msg = static_cast<GetAudioSrcMsg*>(msgData);
+   if (msg->count < msg->maxEntryCount)
+      memcpy(&msg->entries[msg->count], &pupAudioSrcDef, sizeof(AudioSrcId));
+   msg->count++;
+}
 
 CtlResId UpdateAudioStream(AudioUpdateMsg* msg)
 {
@@ -269,6 +285,17 @@ MSGPI_EXPORT void MSGPIAPI PUPPluginLoad(const uint32_t sessionId, const MsgPlug
    msgApi->SubscribeMsg(endpointId, onGameEndId = msgApi->GetMsgID(VPXPI_NAMESPACE, VPXPI_EVT_ON_GAME_END), OnGameEnd, nullptr);
 
    onAudioUpdateId = msgApi->GetMsgID(CTLPI_NAMESPACE, CTLPI_AUDIO_ON_UPDATE_MSG);
+   onAudioSrcChangedId = msgApi->GetMsgID(CTLPI_NAMESPACE, CTLPI_AUDIO_ON_SRC_CHG_MSG);
+   getAudioSrcId = msgApi->GetMsgID(CTLPI_NAMESPACE, CTLPI_AUDIO_GET_SRC_MSG);
+
+   pupAudioSrcDef.id.endpointId = endpointId;
+   pupAudioSrcDef.id.resId = nextAudioStreamId++;
+   pupAudioSrcDef.overrideId = { 0, 0 };
+   pupAudioSrcDef.type = CTLPI_AUDIO_SRC_BACKGLASS_STEREO;
+   pupAudioSrcDef.format = CTLPI_AUDIO_FORMAT_SAMPLE_FLOAT;
+   pupAudioSrcDef.sampleRate = 48000;
+   msgApi->SubscribeMsg(endpointId, getAudioSrcId, OnGetAudioSrc, nullptr);
+   msgApi->BroadcastMsg(endpointId, onAudioSrcChangedId, nullptr);
 
    const unsigned int getScriptApiId = msgApi->GetMsgID(SCRIPTPI_NAMESPACE, SCRIPTPI_MSG_GET_API);
    msgApi->BroadcastMsg(endpointId, getScriptApiId, &scriptApi);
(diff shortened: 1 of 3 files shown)


vbousquet:
Sorry for the delay, I'm a bit low on free time lately.

I will have a look but as we are priorizing getting a first beta of 10.8.1 out, and plugins won't be a stable feature of next release as they still need quite some work to fully reach this state.

The idea regarding audio source is to be able to have a unique global mixer aware of plugin sources (per plugin) at application level (instead of per plugin setup / COM component setup + operating system level like we have today). This needs more work on the API which is not stabilized yet. When this will be done, plugin that would not fullfill the API would be asserted / rejected, not hacked in.

Le-Syl21:
Closing this in light of #3449 and the discussion above.

After re-reading your earlier note about moving to a unique global mixer aware of plugin sources at the application level (with strict API enforcement), and seeing #3449 merged for the wmp/altsound *override* path via miniaudio, it's clear the right long-term home for PUP's *parallel* coexistence with PinMAME audio is that future unified mixer — not a host-side routing patch tacked onto the current pipeline.

For the record, the issue itself (#3305) is still reproducible: when PinMAME is the active source, PUP's `AudioUpdateMsg` packets get dropped at `Player::OnAudioUpdated` because the active-source filter has no notion of parallel sources. So the underlying bug is real and worth keeping on the radar when the new plugin-aware mixer lands — at that point PUP just needs to register cleanly and the routing handles parallel sources by design.

In the meantime I'll carry the patch on my fork as a local stopgap so Linux cab users running PinMAME + PUP keep their audio.

Thanks for taking the time to explain the direction, and no pressure on this one. Looking forward to the new audio API whenever it's ready.

jsm174:
FWIW, in Discord we get asked a few times a week about why users hear no audio when using pup and pinmame, so this is definitely something we need to solved.

Le-Syl21:
@jsm174 Thanks for confirming the user-side impact — that matches what I see on the Linux cab side, and it's good to know it's reproducing widely enough to come up on Discord regularly.

For context on what the patch actually does, it's a 49/-5 diff scoped to two files:

- **`src/core/player.cpp`** — `Player::OnAudioUpdated` currently drops any `AudioUpdateMsg` whose id ≠ `m_activeAudioSourceId`. The patch tracks the set of explicitly-overridden source ids (built from the existing `CTLPI_AUDIO_GET_SRC_MSG` override chain) and only drops *those*. Result: PinMAME gets silenced when AltSound overrides it (existing behavior preserved), but a non-overriding source like PUP coexists with the active one. Streams of newly-overridden sources are closed when the chain changes; streams that are no longer overridden reopen lazily on the next audio update.
- **`plugins/pup/PUPPlugin.cpp`** — PUP registers itself as an audio source via `CTLPI_AUDIO_GET_SRC_MSG` with `overrideId = 0`, satisfying the API contract and making PUP visible to other plugins that may want to route around or override it. The umbrella registration is metadata; PUP's per-video sub-streams keep allocating their own ids.

It's been running on my fork master since April with no regressions I'm aware of (Linux x64, BGFX), and is part of the build I'm offering to beta-testers in #3366. I'm happy to:

- **Reopen this PR** if you want to take it as a stopgap until the unified mixer @vbousquet sketched out lands — fully understanding it would be obsoleted at that point and removed in favor of the new API.
- Or **open a fresh PR / issue** with the same diff if you'd prefer a clean tracking item rather than reviving this thread.
- Or just **leave the patch on my fork** if neither feels right upstream right now — happy users on Discord can point at my fork release in the meantime.

Whatever works best for you and Vincent's roadmap — just let me know and I'll handle the mechanics.

PR #3359, opened 2026-04-30
Fix build: add missing standard library includes

francisdb:
- imconfig.h: include <cstdint> for uint32_t (used by ImDrawIdx)
- PUPLabel.h: include <climits> for INT_MIN/INT_MAX

Newer libstdc++ no longer transitively pulls in these headers.


toxieainc:
Thanks!

PR #3366, opened 2026-04-30
Linux: fix borderless playfield placement and Wayland focus on multi-monitor cabs

Le-Syl21:
> 📦 **Builds on top of upstream `56ab1edc3` (BGFX: Readd fullscreen without video mode change).** This PR no longer ships its own fullscreen path — it auto-triggers Vincent's mechanism for the cabinet case, drops a couple of conflicting SDL window flags on Wayland, and adds focus handling that the protocol requires for newly-mapped fullscreen toplevels on Mutter Wayland.

## TL;DR for reviewers

This PR makes plain `vpinball -play table.vpx` land correctly on a multi-monitor Linux cab — both X11 and Wayland, with or without a launcher that mints an `xdg_activation_v1` token. **Four small, orthogonal commits**, all gated to Linux at runtime so no other platform is touched.

1. **diag** — universal: logs `XDG_CURRENT_DESKTOP` / `XDG_SESSION_TYPE` / `WAYLAND_DISPLAY` at startup, and SDL `WINDOW_MOVED` / `WINDOW_DISPLAY_CHANGED` events with the window title. Makes opaque "the window ended up on the wrong screen" reports self-describing.
2. **auto-enable BGFX fullscreen** — when a VPX window's requested size matches its target display in size (= cabinet layout), auto-toggle `m_fullscreen=true` on Linux. The actual SDL fullscreen setup is then handled by your `56ab1edc3` BGFX fullscreen path; we just auto-flip the toggle so users don't have to discover `[Player]PlayfieldFullScreen=true` in the INI. Drops `SDL_WINDOW_UTILITY|ALWAYS_ON_TOP|BORDERLESS` for that path because they conflict with the fullscreen state on Wayland (UTILITY maps to `xdg_toplevel.set_parent` → regroups same-app_id windows onto one output, undoing the pin).
3. **focused-keep** — report the Playfield as focused when it's a fullscreen toplevel that Mutter Wayland refuses to grant focus to (sibling fullscreen toplevels of the same app_id mapped on adjacent outputs is a known Mutter focus-stealing-prevention quirk). Without this, `Player::IsPlaying()` keeps the simulation paused until the user Alt+Tabs onto the Playfield — frozen cab on boot. Restricted to `VPXWINDOW_Playfield` + `SDL_WINDOW_FULLSCREEN`.
4. **XDG forward** — bundle the launcher's `XDG_ACTIVATION_TOKEN` and apply it explicitly to the Playfield's `wl_surface` from `Window::Show` via a small `WaylandActivation` helper that binds `xdg_activation_v1` from the registry. SDL3's built-in handling consumes the env var on the *first* window it shows, but VPX doesn't guarantee the Playfield is first. With this, launchers that mint a token (PinReady ≥ 0.13.2, GNOME Files, KDE Dolphin, `gtk-launch`, …) get Playfield focus immediately on Wayland instead of dock-badge flicker.

## What's the actual user-visible bug

Multi-monitor cabinet on a stock Linux desktop (no GNOME Kiosk, no `windowrulev2`, just `vpinball -play table.vpx`):

- **X11 + Mutter** (GNOME/Cinnamon/Pop!_OS/Unity/Budgie…): Mutter routes borderless screen-sized NORMAL toplevels to the focused display, ignoring `SDL_PROP_WINDOW_CREATE_X/Y`. Playfield ends up stacked behind Backglass / ScoreView on the wrong monitor, only briefly visible when those auxiliaries close.
- **Wayland + Mutter**: client positioning is forbidden by design (`SDL_SetWindowPosition` is a no-op). Even if each window is initially mapped to the correct output, Mutter applies an auto-management policy a few hundred ms after mapping that regroups same-`app_id` toplevels onto a single output. The `SDL_WINDOW_UTILITY` flag (which SDL3 maps to `xdg_toplevel.set_parent`) accelerates the regrouping. Borderless screen-sized windows also trigger SDL3 issue [libsdl-org/SDL#8186](https://github.com/libsdl-org/SDL/issues/8186). And once placed, newly-mapped fullscreen siblings can be denied focus by Mutter without an `xdg_activation_v1` token.

Both desktop-server modes show the same end-user symptom (windows on the wrong monitor + frozen cab) and affect a large share of Linux Standalone users — anyone running on a regular Linux desktop or a distro-based cab.

## Why this approach (and what changed vs the earlier revision)

The fundamental observation is that on both X11 and Wayland, **the only protocol-level primitive that pins a toplevel to a specific monitor is the fullscreen one**:
- Wayland: `xdg_toplevel.set_fullscreen(output)`.
- X11: `_NET_WM_STATE_FULLSCREEN` + `_NET_WM_FULLSCREEN_MONITORS` (Mutter X11 honors these reliably, unlike position hints for borderless toplevels).

When the user originally opened this PR, the BGFX backend hard-coded `m_fullscreen = false` and skipped the entire fullscreen setup, so the patch had to spin up its own fullscreen path. With **`56ab1edc3` (Vincent, 2026-06-23)** re-introducing the BGFX fullscreen path (no video mode change, exactly the same `SDL_SetWindowFullscreenMode(SDL_GetCurrentDisplayMode(selectedDisplay.display))` pattern), there's no longer any reason to duplicate it. **The patch now reuses that path** and only adds:

- An auto-toggle so cabinet users (window size matches display size) don't have to know about the new `[Player]*FullScreen=true` INI knobs you re-added — the zero-config goal stays.
- A drop of `SDL_WINDOW_UTILITY|ALWAYS_ON_TOP|BORDERLESS` for the auxiliary windows on the Linux cabinet path. Without it, Mutter Wayland regroups Backglass / ScoreView under the Playfield via `set_parent` regardless of the fullscreen state, and the cabinet placement is undone after a few hundred ms.
- The focus and activation handling that the protocol requires when you do this on Mutter Wayland (the last two commits).

If `56ab1edc3` lands first as expected, this PR is essentially "the Linux cabinet glue around your BGFX fullscreen mechanism".

## Out of scope

- This is not about *exclusive* fullscreen with a video mode change — we always reuse the current desktop mode. No display blink, no DPI dance, no resolution change. It's the SDL3 equivalent of a borderless windowed-fullscreen pinned to a specific output.
- Aux window flag drops only apply when the auxiliary is itself sized to its target display (cabinet layout). Users running with a 800×600 Backglass on a desktop are not affected — runtime check on size + SDL video driver.
- For dedicated kiosk installations, [GNOME Kiosk 49+](https://gitlab.gnome.org/GNOME/gnome-kiosk) provides a compositor-side equivalent via `window-config.ini`. This PR's target is the rest of the Linux user base: plain desktop sessions, distro-based cabs, no Kiosk, no manual `windowrulev2`.

## Compatibility matrix (call for beta testers)

The user's Linux cab runs Ubuntu 24.04 + GNOME 46 on both X11 and Wayland — those two are covered. I'd love confirmations or failure reports for the others before merge.

| Compositor / WM    | Session   | `XDG_CURRENT_DESKTOP` | Result            |
|--------------------|-----------|-----------------------|-------------------|
| Mutter (GNOME)     | Wayland   | `GNOME`               | ✅ Fixed (Le-Syl21) |
| Mutter (GNOME)     | X11       | `GNOME`               | ✅ Fixed (Le-Syl21) |
| COSMIC             | Wayland   | `COSMIC`              | ✅ confirmed (@francisdb in earlier comment) |
| Muffin (Cinnamon)  | X11       | `X-Cinnamon`          | ❓ untested        |
| KWin (KDE Plasma)  | Wayland   | `KDE`                 | ❓ untested        |
| KWin (KDE Plasma)  | X11       | `KDE`                 | ❓ untested        |
| Hyprland           | Wayland   | `Hyprland`            | ❓ untested        |
| sway               | Wayland   | `sway`                | ❓ untested        |
| labwc (Batocera)   | Wayland   | `labwc`               | ❓ untested        |
| Xfwm (XFCE)        | X11       | `XFCE`                | ❓ untested        |

A prebuilt build (`linux-x64` + `linux-aarch64`) is at https://github.com/Le-Syl21/vpinball/releases — `VPinballX_BGFX-*-linux-*-Release.tar.gz`. The diag commit included in this PR means test reports automatically capture compositor, session type, requested vs WM-placed display, and every subsequent `MOVED` / `DISPLAY_CHANGED` event.

## Test plan

**Linux compositors — stock config (no `windowrulev2`, no Kiosk):**
- [x] Mutter / GNOME — Wayland (Le-Syl21)
- [x] Mutter / GNOME — X11 (Le-Syl21)
- [x] COSMIC — Wayland (@francisdb)
- [ ] Muffin / Cinnamon — X11
- [ ] KWin / KDE Plasma — Wayland
- [ ] KWin / KDE Plasma — X11
- [ ] Hyprland — Wayland
- [ ] sway — Wayland
- [ ] labwc / Batocera — Wayland
- [ ] Xfwm / XFCE — X11

**Other:**
- [ ] Single-display setup — no-op expected (window size == display size + only one display = same place anyway)
- [x] Windows / macOS / iOS / Android — code path gated out at runtime via `SDL_GetCurrentVideoDriver()` and `__linux__ && !__ANDROID__` guards (covered by fork CI matrix)


dekay:
Reported as not working in Debian on Discord. This patch might need more work.

https://discord.com/channels/652274650524418078/1076655472846831667/1500896775773487174

francisdb:
Can we keep the discussion going here instead of pointing to discord?
I can confirm the issue, vpinball is not fullscreen with a default config and window decorations are visible.

Tested on COSMIC with default config (+ `SyncMode 0`) - the current patch fixes the issue for me.

- default (XWayland): true fullscreen + 60 fps (XWayland's broken high-rate refresh detection, unrelated to this PR)
- `SDL_VIDEO_DRIVER=wayland`: true fullscreen + 220 fps

_Small remark: walls of text on PRs can scare reviewers away - might be worth condensing. The summary is that there are normally 3 window modes (Windowed / Fullscreen borderless / Exclusive fullscreen-deprecated), and on Linux/Mac vpinball doesn't currently request either fullscreen mode from the windowing system. Windows auto-promotes a borderless display-sized window for you; X11/Wayland don't, so the compositor leaves the panel and decorations on top._

Le-Syl21:
Hey @dekay — my bad on this one, I should have pointed to my build from the start.

The latest release bundles Vincent's two upstream Vulkan fixes from his bgfx force-push: lighter per-frame VSync on/off, and waitable swapchain support for reduced latency. Release notes spell out exactly what's in.

A retest against that build would be much appreciated: https://github.com/Le-Syl21/vpinball/releases

Le-Syl21:
Thanks for the COSMIC confirmation @francisdb 🙏

Point taken on the wall of text — I'll keep future descriptions tighter.

Side observation while we're here: my impression is VSync doesn't lock onto the playfield display. It seems to sync against *all* attached displays and settle on the slowest one. A 144 Hz playfield then caps at the backglass's 60 Hz.

Which on a pincab is the flipper equivalent of those knockoff SNES pads with the "slow" button — the one spamming Start at 10 Hz to fake slow-motion.

francisdb:
@Le-Syl21 should this stay in draft? I'd like to see this one merged.

Le-Syl21:
> @Le-Syl21 should this stay in draft? I'd like to see this one merged.

Hey @francisdb ! @vbousquet moved it to draft, so I'll leave it in his hands to decide. Not my repo to manage. 😄

francisdb:
> Reported as not working in Debian on Discord. This patch might need more work.
> 
> https://discord.com/channels/652274650524418078/1076655472846831667/1500896775773487174

@superhac since you were the one reporting that this pr breaks things, can you add some more info here? What breaks on what setup? We need something actionable instead of "it does not work"  to push this through.

superhac:
> > Reported as not working in Debian on Discord. This patch might need more work.
> > https://discord.com/channels/652274650524418078/1076655472846831667/1500896775773487174
> 
> @superhac since you were the one reporting that this pr breaks things, can you add some more info here? What breaks on what setup? We need something actionable instead of "it does not work" to push this through.

I'm Debian 13 with Hyprland.  When the patch is applied the windows no longer move to their respective placement.   Instead they stack on the primary monitor.  

my window position rules on hyprland:
```
# VPinball
windowrule {
  name = dmd-vpinball-screen
  match:initial_title = ^Visual Pinball Score View$
  workspace = 3 silent
  fullscreen = 1
}

windowrule {
  name = bg-vpinball-screen
  match:initial_title = ^Visual Pinball Backglass$
  workspace = 2 silent
  fullscreen = 1
}

windowrule {
  name = table-vpinball-screen
  match:initial_title = ^Visual Pinball Player$
  workspace = 1 silent
  fullscreen = 1
  #match:focus  = true
  #stay_focused = on
  #allows_input = on
  #focus_on_activate = on
}
```

At least in my IMHO I would focus on getting more people to test.  Besides @Le-Syl21 and yourself I haven't heard from anyone else on this.  Either way don't let me hold this up, I can work around anything now.

superhac:
I will also add there used to be props for this:
```
PlayfieldFullScreen
BackglassFullScreen
ScoreViewFullScreen
```

The issue popped up when these were removed.  So if I understand the commit correctly this is just hardcoding what these options used to do?

Le-Syl21:
Thanks @superhac, that's the actionable detail we needed.
  
  One correction first: those props weren't removed — PlayfieldFullScreen / BackglassFullScreen / ScoreViewFullScreen still exist in Settings_properties.inl. What changed is their meaning: today *FullScreen=true is exclusive fullscreen (a real video-mode change), which is discouraged. This patch doesn't restore
  the old behavior — it adds a separate path, a desktop-mode fullscreen (no resolution change), triggered automatically when the playfield is sized to its display.

  And that "automatically" is exactly what bites you. There's no portable Wayland way to say "put this window on output X" — it's compositor policy:

  - Mutter (GNOME/COSMIC): ignores borderless positioning, only honors a client fullscreen-on-output → the patch fixes placement.
  - wlroots (Hyprland/sway): honors your title/workspace rules, but a client-issued set_fullscreen overrides them and lands on the focused monitor → the patch breaks your layout.

  So a single hardcoded strategy can't win on both.

  The version that should satisfy both: keep the borderless path, and fall back to desktop-fullscreen only if the window actually ended up off its target display (check SDL_GetWindowPosition after SDL_SyncWindow). On Mutter placement fails → we correct it; on Hyprland your rules already placed it right → no
  fallback, no regression.

  I'm stretched thin on other work right now, so I'll leave the draft status to @vbousquet — happy for anyone to pick up the auto-detect approach above, or I can push it later if there's appetite to land this.

superhac:
> One correction first: those props weren't removed — PlayfieldFullScreen / BackglassFullScreen / ScoreViewFullScreen still exist in Settings_properties.inl. What changed is their meaning: today *FullScreen=true is exclusive fullscreen (a real video-mode change), which is discouraged. This patch doesn't restore the old behavior — it adds a separate path, a desktop-mode fullscreen (no resolution change), triggered automatically when the playfield is sized to its display.
> 
> And that "automatically" is exactly what bites you. There's no portable Wayland way to say "put this window on output X" — it's compositor policy:

Those options no longer show up in my VPinballX.ini?  Can someone else confirm they actually have these options in their `ini`?

francisdb:
They are only removed for the `BGFX` build

https://github.com/vpinball/vpinball/blob/bd65c64fe04b41f2ecfdb00ecdc6aef1eaf05706/src/core/Settings_properties.inl#L25-L31

superhac:
> They are only removed for the `BGFX` build

So if you switch to OpenGL build does your problem go away?

francisdb:
> > They are only removed for the `BGFX` build
> 
> So if you switch to OpenGL build does your problem go away?

No idea, I don't switch to the OpenGL build, that's deprecated as far as I know.

superhac:
> No idea, I don't switch to the OpenGL build, that's deprecated as far as I know.

Me neither.  Yes that what I heard also.  

So the gist is we've confirmed that the options were indeed removed and this is hardcoding what was removed?

francisdb:
In the BGFX build `m_fullscreen` is hardcoded to false with the exclusive-fullscreen props compiled out, so the playfield is always a plain SDL_WINDOW_BORDERLESS window sized to the display - VPX never issues a fullscreen request to the windowing system on any platform. It only looks fullscreen on Windows because DWM auto-promotes a borderless window that exactly covers a monitor (taskbar hidden); Wayland never does that - a surface is fullscreen only if the client explicitly calls xdg_toplevel.set_fullscreen, otherwise the panel and decorations stay on top.

I think only @vbousquet can answer why the code was changed that way.

https://github.com/vpinball/vpinball/blob/bd65c64fe04b41f2ecfdb00ecdc6aef1eaf05706/src/renderer/Window.cpp#L77-L79

Le-Syl21:
Pushed a reworked version (rebased on current `master`).

**Approach changed per the discussion above** — detection instead of an unconditional fullscreen:

- The playfield is created as a normal `SDL_WINDOW_BORDERLESS` window, as before.
- On the first `Show()` (i.e. once the window is actually *mapped* — `SDL_SyncWindow`), it compares `SDL_GetDisplayForWindow()` against the intended display.
- **Only if it landed on the wrong display** does it fall back to a desktop-mode fullscreen on the target display (no video-mode change).

So compositors that already place the window correctly (wlroots: sway, Hyprland, …) hit **no fallback and keep their user/workspace rules** — this should fix @superhac's regression. Mutter, which mis-places the borderless window, gets corrected. Using `SDL_GetDisplayForWindow()` rather than the window position is deliberate: the position is unqueryable on Wayland, the output association is not.

Also: the branch is now rebased on upstream `master`, so it inherits Vincent's BGFX VSync / waitable-swapchain fixes natively (`339c0c1cc`). The "test against my build" caveat from the earlier Debian report no longer applies — a plain `master` build already has them.

**Validation:**
- ✅ Mutter / Wayland (GNOME, multi-display cab): playfield now lands on the correct monitor; backglass/scoreview unchanged.
- ⚠️ Still needs testing on **non-Mutter**, especially **wlroots (Hyprland/sway)**, to confirm the no-fallback path leaves placement untouched. @superhac — a retest on your Hyprland setup would confirm the regression is gone.

Leaving the draft status to @vbousquet.

dekay:
Would be nice to test this under labwc which is the Wayland compositor used by Batocera.

superhac:
@Le-Syl21 if you use the OpenGL build does your problem go away?

Le-Syl21:
@superhac I don't run the OpenGL build (it's deprecated), so I can't tell you. But the thing to keep in mind: this is about the **WM placing the window**, not about what gets rendered *inside* it. SDL is the windowing layer — shared by both the BGFX and OpenGL builds — and it's what talks to the compositor; BGFX/OpenGL only draw into the surface SDL hands them. So the renderer is orthogonal to placement: switching it wouldn't change where the compositor decides to put the window.

The only build-specific bit is the `#ifndef ENABLE_BGFX` that keeps the fullscreen request compiled into the OpenGL path — that's a code path, not a rendering difference.

Le-Syl21:
@dekay fair point, but I only test on Ubuntu (Mutter/GNOME) — I don't have a Batocera/labwc setup here. That said, labwc is wlroots-based, so it's in the same family as sway/Hyprland: if it places the window correctly (as wlroots compositors do), the detection sees it on the right output and applies **no fallback** — so it should be a no-op there. Would need someone running Batocera to confirm, though.

Le-Syl21:
For anyone who'd like to validate this without building from source, I've published a prebuilt fork release containing this fix on top of current upstream `master`:

**https://github.com/Le-Syl21/vpinball/releases/tag/v10.8.1-5096-f22b239**

Relevant artifact (the fix is Linux-only): [`VPinballX_BGFX-...-linux-x64-Release.tar.gz`](https://github.com/Le-Syl21/vpinball/releases/download/v10.8.1-5096-f22b239/VPinballX_BGFX-10.8.1-5096-f22b239-linux-x64-Release.tar.gz)

**Mutter multi-monitor (GNOME/Cinnamon):** already confirmed working on my 3-display cabinet (playfield + FullDMD + backglass) — the borderless playfield lands on the intended display instead of being relocated by the compositor.

The one thing left to confirm is on **any other compositor** (KDE/KWin, wlroots-based ones like sway/Hyprland/labwc/Batocera, …): since the detection is behaviour-based — it only kicks in when the window ends up on the *wrong* output — it should be a true **no-op** wherever the compositor already places the playfield correctly. Feedback from non-Mutter setups especially welcome 🙏

superhac:
> For anyone who'd like to validate this without building from source, I've published a prebuilt fork release containing this fix on top of current upstream `master`:
> 
> **https://github.com/Le-Syl21/vpinball/releases/tag/v10.8.1-5096-f22b239**
> 
> Relevant artifact (the fix is Linux-only): [`VPinballX_BGFX-...-linux-x64-Release.tar.gz`](https://github.com/Le-Syl21/vpinball/releases/download/v10.8.1-5096-f22b239/VPinballX_BGFX-10.8.1-5096-f22b239-linux-x64-Release.tar.gz)
> 
> **Mutter multi-monitor (GNOME/Cinnamon):** already confirmed working on my 3-display cabinet (playfield + FullDMD + backglass) — the borderless playfield lands on the intended display instead of being relocated by the compositor.
> 
> The one thing left to confirm is on **any other compositor** (KDE/KWin, wlroots-based ones like sway/Hyprland/labwc/Batocera, …): since the detection is behaviour-based — it only kicks in when the window ends up on the _wrong_ output — it should be a true **no-op** wherever the compositor already places the playfield correctly. Feedback from non-Mutter setups especially welcome 🙏

One important detail for anyone testing on Hyprland/wlroots though: because Hyprland applies monitor/window rules after map time, some applications can still appear to “race” briefly during initial mapping depending on whether they are native Wayland, SDL/X11, or XWayland. This patch can cause adverse interaction with existing placement behavior.

So the sequence is usually more like:

App creates window
Window appears/mapped
Hyprland evaluates rules
Hyprland moves/resizes/floats/fullscreens it
App may fight back afterward (depending on toolkit/app)

Le-Syl21:
Thanks, that's a fair point about the ordering on wlroots — rules being evaluated *after* map is exactly the kind of timing the detection has to respect.

The direction I'd take from this is to **not** evaluate placement synchronously on the first map, but to let the compositor settle first — defer the check by a frame / a short event-pump cycle so `SDL_GetDisplayForWindow()` reflects the *final* output association rather than an intermediate one. On Wayland that association only becomes meaningful once `wl_surface.enter` has fired anyway, so reading it too early is the real risk here. Once settled, a compositor that places the playfield correctly (as wlroots does) reports the right output → **no fallback**, so there's nothing to fight.

That said, before I rework the timing: **did you actually get to test the prebuilt build on your Hyprland setup?** And if so, what did you observe in practice — does the playfield actually get disturbed, or is it the no-op I'd expect? That'd tell us whether this is a real race in the current code or a theoretical one, and it'd shape how aggressively I need to defer the check.

superhac:
> Thanks, that's a fair point about the ordering on wlroots — rules being evaluated _after_ map is exactly the kind of timing the detection has to respect.
> 
> The direction I'd take from this is to **not** evaluate placement synchronously on the first map, but to let the compositor settle first — defer the check by a frame / a short event-pump cycle so `SDL_GetDisplayForWindow()` reflects the _final_ output association rather than an intermediate one. On Wayland that association only becomes meaningful once `wl_surface.enter` has fired anyway, so reading it too early is the real risk here. Once settled, a compositor that places the playfield correctly (as wlroots does) reports the right output → **no fallback**, so there's nothing to fight.
> 
> That said, before I rework the timing: **did you actually get to test the prebuilt build on your Hyprland setup?** And if so, what did you observe in practice — does the playfield actually get disturbed, or is it the no-op I'd expect? That'd tell us whether this is a real race in the current code or a theoretical one, and it'd shape how aggressively I need to defer the check.

I tested this on Hyprland and I’m seeing a regression.

With the patch applied, all VPinball windows are now getting stuck on one screen instead of being placed across my intended cabinet displays. So at least on Hyprland, this does not appear to be a no-op.

This may be related to timing: Hyprland applies its window/monitor rules after the window is mapped, and this patch may be detecting/intervening before Hyprland has completed its own placement/rule handling. In that case, the behavior-based detection could think the compositor placed the window wrong, even though Hyprland would have corrected it right after mapping.

So I’d be careful calling it non-Mutter-safe yet. It may need to be restricted to Mutter/GNOME/Cinnamon, or delayed enough to avoid racing compositor window rules on Hyprland/wlroots.

Le-Syl21:
Quick clarification needed to interpret the regression correctly: **does your `hyprland.conf` already contain `windowrulev2` (or equivalent) routing the VPX windows (Playfield / Backglass / DMD / ScoreView…) to specific monitors or workspaces?**

The reason I ask: the bug this patch is meant to fix is the *out-of-the-box* case where the user has done **no** WM-side configuration — exactly the scenario where Mutter mis-places a borderless screen-sized window with no recourse. On vanilla Hyprland I'd expect the same misplacement (Wayland generally ignores the position hint SDL passes), and the patch's fallback should correct it.

But if your config already has rules that explicitly route VPX windows, you're in a different scenario: your rules already solve the placement, and the patch is colliding with them rather than helping. That's not the regression case — that's the "two solutions for the same problem fighting each other" case. The deferred-check rework I was planning would let them coexist (the patch backs off when the WM has already placed things correctly), but the *primary* validation should be against a vanilla Hyprland with no VPX-specific rules.

Could you share the relevant chunk of your `hyprland.conf`, or confirm whether it's vanilla? That'd let us tell which of the two we're looking at.

Le-Syl21:
I've refreshed the PR description with two changes:

1. The **Fix** section now describes the current auto-detect approach (the original text still described the pre-rework unconditional-fullscreen version).
2. Added a **compatibility matrix** to track validation across compositors / sessions / vanilla-vs-custom-config — including a column that distinguishes "the WM mis-places out of the box" (where the patch is supposed to help) from "the user already has WM-side rules" (where the patch should be a no-op, currently subject to a timing race on Hyprland).

If you're testing on a setup not yet ticked in the matrix, please drop a comment with the row + result (✅/❌) and a short note on your config — especially: vanilla WM with no VPX-specific rules vs. custom `windowrule`/equivalent. That'll help triage whether any further ❌ is a patch bug or a timing collision with user-side placement rules.

Le-Syl21:
Force-pushed an update to this PR with two changes:

1. **Rebased on top of #3433** (the diagnostic logging PR). The branch now contains two commits: the diag commit (identical to #3433), followed by the placement-fix commit on top. If #3433 lands first, this PR will auto-reduce to a single commit via patch-id deduplication.

2. **Cleanup made possible by the rebase**: the placement-fix no longer carries its own `m_linuxTargetDisplay` / `m_linuxPlacementChecked` members — both were redundant with the universal `m_targetDisplayId` / `m_placementLogged` introduced by #3433. The `Show()` method now does **a single** `SDL_SyncWindow` + `SDL_GetDisplayForWindow` shared between the diag log and the fullscreen-fallback decision (was two separate blocks before).

A fresh prebuilt fork build is available with both commits applied: **https://github.com/Le-Syl21/vpinball/releases/tag/v10.8.1-5098-877f05b**

This is especially useful for **@superhac** to retest on Hyprland — the new diag logs from #3433 will make the timing race you reported directly visible in your log output. Specifically you should see something like:

```
[INFO] Linux desktop session: XDG_CURRENT_DESKTOP=Hyprland XDG_SESSION_TYPE=wayland WAYLAND_DISPLAY=wayland-1
[INFO] Window #0 first mapped: requested display=N | WM placed display=M at (...) ...
[INFO] Playfield was mis-placed by the compositor; corrected via desktop-mode fullscreen on the target display.
[INFO] Window 'Playfield' moved to display ID K     ← Hyprland's windowrulev2 firing late
[INFO] Window 'Playfield' moved to (...)            ← and again as it settles
```

That trace will tell us precisely (a) whether your `hyprland.conf` is what's repositioning the window, (b) the temporal order between our fallback and your rules — which is the missing piece we need to decide between *deferred detection* and *opt-out gating*. If you can grab the full session log when you retry, that'd close out the diagnosis for that compositor.

superhac:
@Le-Syl21 

GNOME Kiosk 49/50 added a bunch of new window placement controls specifically because Mutter/Wayland normally does not allow arbitrary window positioning like X11 did. The important changes for multi-monitor kiosk setups are:

- set-on-monitor
- lock-on-monitor
- lock-on-monitor-area
- guaranteed initial placement/fullscreen application on first map
- improved auto-fullscreen behavior

These landed across GNOME Kiosk 49 and 50.

The big thing is: GNOME Kiosk now owns placement policy itself instead of relying on clients calling SDL_SetWindowPosition() or X11 positioning hints.

Example window-config.ini:
```
[all]
set-fullscreen=false

[playfield]
match-title=VPinballX
set-on-monitor=HDMI-1
lock-on-monitor=true
set-fullscreen=true

[backglass]
match-title=Backglass
set-on-monitor=DP-1
lock-on-monitor=true
set-fullscreen=true
```

GNOME Kiosk 50 also specifically fixed cases where:

- fullscreen/size/position were ignored during initial map 
- windows got relocated unexpectedly by Mutter
- fullscreen detection behaved incorrectly with monitor-sized windows

For pinball cabinet style setups, lock-on-monitor=true is especially important because it prevents a window from migrating to another display if monitor topology changes or focus changes.

Another relevant detail:
GNOME Kiosk is now Wayland-only because Mutter dropped the X11 backend in newer versions.

Also regular GNOME 50 completely dropped the X11 backend. The GNOME Shell, Mutter window manager, and GDM login manager no longer contain any X11 code. This means you cannot log into a native GNOME X11 session, though legacy X11 apps continue to work natively through built-in XWayland.

Le-Syl21:
Thanks for the deep context — really useful to know GNOME Kiosk converged on `lock-on-monitor` and explicit per-window placement.

What this PR does is the client-side equivalent of that approach: when a VPX window's requested size matches its target display, switch to `xdg_toplevel.set_fullscreen(target_output)` (via `SDL_SetWindowFullscreenMode + SDL_SetWindowFullscreen`) at creation. On Wayland that's the only protocol primitive that pins a toplevel to a specific output. On X11 the same SDL call uses `_NET_WM_FULLSCREEN_MONITORS`, which Mutter X11 honors reliably (unlike position hints for borderless screen-sized toplevels, which Mutter routes to the focused display).

Rolling out a simplified single-mechanism version right now — both X11 and Wayland share the same proactive pin, no more detect-then-fallback in `Show()`. Recent testing on a 3-display NVIDIA cab confirms it gives the same lock-on-monitor effect Kiosk 49/50 implements compositor-side: no post-mapping relocation, no musical chairs, on both X11 and Wayland.

Re: GNOME 50 dropping the Mutter X11 backend — that's huge and reinforces the direction. The same code path covers both the legacy XWayland case (still relevant short-term) and the Wayland-native case (long-term default).

GNOME Kiosk + a `window-config.ini` remains the cleanest answer for dedicated cab installations that deploy Kiosk specifically. The goal here is for plain `vpinball -play table.vpx` to land correctly out-of-the-box on standard GNOME desktops — which it now does on both sessions with a single client-side mechanism. Will mention the Kiosk path in the PR description as a complementary option for fixed installations.

superhac:
@Le-Syl21
 
I think this is the correct long-term direction, especially given where the Linux desktop stack is headed.

The important point here is that relying on X11-era positioning semantics is becoming increasingly fragile because X11 itself is now effectively legacy infrastructure. GNOME has already removed the native Mutter X11 backend in GNOME 50, and other compositors are steadily reducing investment in X11-specific behavior as Wayland becomes the primary path forward. At this point, any solution built around “classic” borderless-window positioning and post-map relocation is solving yesterday’s compositor model, not tomorrow’s.

What makes your approach compelling is that it aligns with the protocol-native mechanism Wayland compositors actually support:

- xdg_toplevel.set_fullscreen(target_output) on Wayland
- _NET_WM_FULLSCREEN_MONITORS on X11 as the compatibility bridge

That is fundamentally different from trying to fight the compositor with placement hints that modern compositors are intentionally free to ignore.

The GNOME Kiosk comparison is particularly telling because it demonstrates the same architectural conclusion from the compositor side:

- explicit monitor affinity
- fullscreen-bound output targeting
- no dependency on arbitrary window coordinates
- no “repair after map” logic

So VPX moving to a proactive “pin-to-output-at-creation” model is not just a workaround for Mutter quirks — it is aligning the application with the modern compositor contract.

And honestly, that matters more now than preserving perfect behavior for increasingly legacy X11 workflows. XWayland compatibility is still important in the short term, but it should be treated as a transitional layer, not the architectural foundation. The reality is:

- native X11 support across desktops will continue shrinking
- compositor behavior around X11 hints will diverge more over time
- Wayland-native fullscreen/output APIs are the stable future surface

Given that, using one unified fullscreen/output-selection path for both X11 and Wayland makes a lot of sense. It reduces special-case logic, matches compositor expectations, and avoids depending on behavior that desktop environments are actively moving away from.

The biggest win here is probably predictability: no more “focused monitor roulette,” no post-show reposition attempts, and no compositor-specific heuristics trying to reverse-engineer placement after mapping.

Le-Syl21:
@superhac — thanks, that's a really nice framing. *"Transitional layer, not foundational"* and *"modern compositor contract"* capture exactly what we wanted this PR to be, and that vocabulary is much cleaner than what I was reaching for.

Just force-pushed the unified version (commit `9bb6eef72`): **a single proactive pin path covers both X11 and Wayland at the source**, no post-map detection, no fallback, no `Show()` observer. The code is around 30 lines including comments, and the diff is entirely in `Window.cpp`'s constructor:

- Runtime check on `SDL_GetCurrentVideoDriver()` to detect Linux desktop (X11 or Wayland)
- When the requested window matches its target display size: skip `SDL_WINDOW_BORDERLESS` + `SDL_WINDOW_UTILITY|ALWAYS_ON_TOP`, set `SDL_WINDOW_FULLSCREEN`, and pass the target display's desktop mode to `SDL_SetWindowFullscreenMode + SDL_SetWindowFullscreen(true)`
- That single SDL call resolves to `xdg_toplevel.set_fullscreen(output)` on Wayland and `_NET_WM_STATE_FULLSCREEN + _NET_WM_FULLSCREEN_MONITORS` on X11 — both protocol-blessed primitives that the compositor cannot ignore or override post-mapping

Validated on the 3-display NVIDIA cab in **three sessions**:

| Session | SDL driver | Result |
|---|---|---|
| Pure X11 (GNOME 49 X11 login) | x11 | ✅ all 3 windows pinned at creation, no `MOVED` event |
| Wayland session + XWayland (SDL default) | x11 | ✅ same |
| Wayland session + Wayland native (`SDL_VIDEODRIVER=wayland`) | wayland | ✅ all 3 pinned, no post-map regrouping |

In all three cases the diag log from #3433 shows `Window #X first mapped: requested display=Y | WM placed display=Y` with no subsequent `MOVED` / `DISPLAY_CHANGED` events. The "focused monitor roulette" and the Mutter Wayland musical chairs are both gone, in one shot, with the same code path.

I'll cut a new pre-release build on the fork as soon as CI clears so anyone wanting to retest (especially non-Mutter setups — KWin, Hyprland, sway, labwc) has an up-to-date binary; will post the link here once it's published.

superhac:
@Le-Syl21 

This looks much cleaner architecturally now that the legacy reposition/fallback logic is gone entirely and everything routes through the compositor-approved fullscreen/output path from creation time.

I still think the key point here is that we should avoid designing around legacy X11 placement semantics going forward. X11 is already in maintenance-mode territory across the Linux desktop ecosystem:

- GNOME removed the native Mutter X11 backend in GNOME 50
- Wayland-native compositors continue becoming the default
- XWayland is increasingly treated as a compatibility bridge, not a primary target platform

Because of that, depending on borderless-window positioning behavior, post-map coordinate corrections, or WM-specific X11 hint interpretation is only going to become less reliable over time.

What your unified implementation does instead is target the actual fullscreen/output-selection primitives both stacks expose:

- Wayland → xdg_toplevel.set_fullscreen(output)
- X11/XWayland → _NET_WM_FULLSCREEN_MONITORS

Those are the only mechanisms modern compositors are really expected to preserve consistently because they map to explicit compositor policy rather than legacy client-controlled positioning.

The fact that the same path now works:

- native Wayland
- XWayland
- pure X11
- without special handling is a strong indication the implementation is aligned with the compositor model instead of fighting it.

Honestly, from a maintenance perspective, this is probably the biggest win:

- no WM-specific relocation heuristics
- no “fix it after mapping”
- no MOVED-event observers
- no compositor behavior probing
- no divergence between Wayland and X11 code paths

That is exactly the direction Linux desktop applications need to move as X11 continues fading out of active platform support.

At this point I would prioritize correctness on Wayland-native compositors over preserving historical X11 window-management expectations. XWayland compatibility should remain functional, but it should be viewed as transitional compatibility behavior, not the architectural baseline VPX is optimized around.

Le-Syl21:
Pre-release published — same commit (`9bb6eef72`-equivalent) packaged for every platform:

**https://github.com/Le-Syl21/vpinball/releases/tag/v10.8.1-5098-ac787e4**

Linux x64 BGFX artifact (direct link, what most cab users want): [`VPinballX_BGFX-10.8.1-5098-ac787e4-linux-x64-Release.tar.gz`](https://github.com/Le-Syl21/vpinball/releases/download/v10.8.1-5098-ac787e4/VPinballX_BGFX-10.8.1-5098-ac787e4-linux-x64-Release.tar.gz)

15 assets total (Linux BGFX/GL, macOS arm64/x64, Windows x86/x64 Release+Debug + mingw). The build also bundles the diagnostic logs from #3433, so any compositor-specific feedback will come with full session/placement/move context out of the box. Especially looking for **non-Mutter** validation (KWin, Hyprland, sway, labwc) to tick the matrix above.

Le-Syl21:
## 📣 Call for beta testers

This PR is now ready for review. The branch handles **both X11 and Wayland natively and respectfully** — no hacks like X11-only window-manager hints or forced `SDL_VIDEODRIVER=x11`:

- On **X11**, screen-sized playfields are pinned to their target display through `_NET_WM_STATE_FULLSCREEN` + `_NET_WM_FULLSCREEN_MONITORS` (Mutter X11 honors these reliably).
- On **Wayland**, the Playfield is pinned via `xdg_toplevel.set_fullscreen(output)` — the only protocol primitive that lets a client request a specific output. To play well with Wayland's focus-stealing prevention, the launcher's `XDG_ACTIVATION_TOKEN` (set per [xdg-activation-v1](https://wayland.app/protocols/xdg-activation-v1)) is captured at startup and explicitly handed to the Playfield surface so the compositor grants it focus immediately instead of flashing it in the background. This requires a launcher that mints a token (e.g. PinReady ≥ 0.13.2, GNOME Files, KDE Dolphin, `gtk-launch`, …). Without a token, behavior simply falls back to whatever the compositor decides — no regression.

### How to help

I'm looking for testers on compositors I don't run locally to fill in the compatibility matrix in the description. A few minutes is enough — just launch a table on a multi-monitor cab and confirm whether the Playfield lands on the right display and gets focus on first map.

**Please reply with:**
- OS + distribution version (e.g. `Fedora 41`, `Arch`, `Ubuntu 24.04`, `Batocera 41`, `KDE Neon`)
- Display server (`X11` or `Wayland`) — `echo $XDG_SESSION_TYPE`
- Compositor / desktop (`echo $XDG_CURRENT_DESKTOP`)
- Whether placement + focus worked correctly
- If something went wrong, attach the relevant lines from the diagnostic logs (the build includes the diag patch from #3433 — every `MOVED` / `DISPLAY_CHANGED` event is logged with the requested vs WM-placed display)

### Prebuilt binary

Latest fork release with this patch (and the #3433 diag logs) for `linux-x64` and `linux-aarch64`:
👉 **https://github.com/Le-Syl21/vpinball/releases**

Grab the `VPinballX_BGFX-*-linux-x64-Release.tar.gz` (or `-aarch64-Release.tar.gz`) asset from the latest release, drop the binary in place of your current `VPinballX_BGFX`, and run as usual.

Thanks in advance to anyone who can spare a session 🙏

vbousquet:
Latest changes should have solve the issue this PR addressed with a broader and more complete refactor:
- window positioning is now only proposed where supported (https://github.com/vpinball/vpinball/commit/4fe457a3b49123eb4513d8be168e48f61ae5facd)
- borderless fullscreen is now properly supported (https://github.com/vpinball/vpinball/commit/8446b31595b62803db35af6a38a4de99e8256228 ans follow ups)
- focus management has been cleaned up (https://github.com/vpinball/vpinball/commit/42e502acc003370d18dd9fa3911721345fed5589)
- logical units have been dropped from the UI & settings (https://github.com/vpinball/vpinball/commit/c446c2f7e7cd824a1a7df21e7860c5d913d1999c)

Therefore, I will close it for now.

Le-Syl21:
@vbousquet — tested the new fullscreen/positioning refactor (`8446b315` + `4fe457a3` + `42e502ac`) on a 3-display cabinet (5251 build, Mutter/Wayland). The **placement** side is solved: with `PlayfieldFullScreen=1` (borderless), the playfield and ancillary windows now land on their assigned outputs. 👍

There's one **focus** case left that `42e502ac` doesn't cover — because it's the opposite problem. That commit makes ancillary windows `NOT_FOCUSABLE` so they don't *steal* focus, but nothing makes the playfield *take* keyboard focus when it's mapped as a fullscreen toplevel under Mutter/Wayland.

When VPX is started by a frontend (here PinReady), the child process has no usable `xdg_activation_v1` token, so Mutter denies the freshly-mapped fullscreen playfield keyboard focus. Then in the player loop:

```cpp
const bool willPlay = m_playing && m_playfieldWnd->IsFocused();   // player.cpp
```

`IsFocused()` is false → the table stays **paused** on launch. The confusing symptom: PinMAME keeps emulating, so you get **sound but a frozen image** — the ScoreView/DMD shows its static pixel matrix with no animation, the backglass doesn't animate, and the user has to Alt+Tab onto the playfield to unfreeze it. Easy to misdiagnose as a DMD/render bug when it's really the sim being paused.

We're going to open a **small, focused PR** for just this: report the playfield as focused when it's a fullscreen toplevel (so the sim keeps running), and forward the launcher's `XDG_ACTIVATION_TOKEN` to the playfield surface so it actually grabs keyboard focus. Strictly scoped to `VPXWINDOW_Playfield` + fullscreen so focus semantics elsewhere are untouched. Will link it here.

vbousquet:
I'm not sure to understand 2 things in this:
- PinMAME is supposed to be suspended when VPX main window is not focused, so if PinMAME is running, then VPX main window is likely focused
- `XDG_ACTIVATION_TOKEN` is already handled by SDL at least for the first created and activated window which happens to be the playfield

There seems to be something more here that doesn't fit

PR #3386, opened 2026-05-09
Add string and bool element type support to plugin script arrays

Pyrrvs:
## Summary
- Adds `TYPEID_STRING` and `TYPEID_BOOL` support to the array marshaller in `DynamicScript.cpp` (both 1D and 2D code paths), replacing `assert(false)` stubs with working conversions
- Strings are packed as `const char*` pointers in the `ScriptArray` data area and converted to `VT_BSTR` via `MultiByteToWideChar`/`SysAllocStringLen`; bools are packed as `char[]` and mapped to `VT_BOOL`
- Extends the `helloscript` plugin with 1D/2D string and bool array examples so plugin authors have a reference
- Adds a doctest case exercising the 2D marshalling through the full `DynamicTypeLibrary::Invoke` pipeline

## Why
Plugins that use string-identified hardware (e.g. MPF's `"l-1"`, `"coil1"`) need to return data as 2D SAFEARRAYs where column 0 is a string ID and column 1 is a state value. Without string array support, plugin authors targeting macOS/Linux can't match the data formats that existing VBScript tables expect from the Windows COM bridge, splitting the ecosystem.

PinMAME's plugin uses int32 element types because PinMAME's hardware is numbered (lamp 1, coil 5). MPF uses string identifiers, which is why this limitation is hit.

## What changed

**`src/core/DynamicScript.cpp`** — Added `TYPEID_STRING` and `TYPEID_BOOL` cases to `ScriptToCOMVariant` for both 1D and 2D arrays. The string conversion matches the existing scalar string pattern (line ~595). For 2D strings, `VariantClear` is called after `SafeArrayPutElement` to free the local BSTR copy.

**`plugins/helloscript/helloscript.cpp`** — Added 4 example getter functions (`StringArray1D`, `StringArray2D`, `BoolArray1D`, `BoolArray2D`) demonstrating the packing convention for both element types.

**`tests/test-plugin-arrays.cpp`** — New doctest case that creates a standalone `DynamicTypeLibrary`, registers array types and a test class, invokes property getters via `DynamicDispatch`, and verifies SAFEARRAY dimensions and element values (38 assertions).

**`make/vpx-test.vcxproj`** — Added the new test file to the VS test project.

## Memory ownership convention
- Plugin packs `const char*` pointers into the `ScriptArray` data area pointing to plugin-owned strings
- Marshaller copies each string to a BSTR via `SysAllocStringLen`; the SAFEARRAY owns those BSTRs
- After marshalling, `ReleaseScriptVariant` calls `sv.vArray->Release(sv.vArray)` — no cross-boundary ownership transfer

## Test plan
- [x] New doctest case passes on Windows (38/38 assertions, verified via CI)
- [x] All 20 desktop build targets pass (Windows DX9/GL/BGFX, macOS, Linux)
- [x] Mobile builds pass (iOS, Android, Quest)
- [x] SBC builds pass (RPI, RK3588)
- [ ] Existing render test suite (manual — requires GPU)

### DIFF
--- a/make/vpx-test.vcxproj
+++ b/make/vpx-test.vcxproj
@@ -36,6 +36,7 @@
     </ClCompile>
     <ClCompile Include="../../tests/vpx-test.cpp" />
     <ClCompile Include="../../tests/test-base-render.cpp" />
+    <ClCompile Include="../../tests/test-plugin-arrays.cpp" />
   </ItemGroup>
   <ItemGroup>
     <ClInclude Include="../../tests/vpx-test.h" />
--- a/plugins/helloscript/helloscript.cpp
+++ b/plugins/helloscript/helloscript.cpp
@@ -4,9 +4,10 @@
 #include "plugins/VPXPlugin.h"
 #include "plugins/ScriptablePlugin.h"
 #include <cstring>
+#include <cstdlib>
 
 namespace HelloScript {
-   
+
 static const MsgPluginAPI* msgApi = nullptr;
 static VPXPluginAPI* vpxApi = nullptr;
 static ScriptablePluginAPI* scriptApi = nullptr;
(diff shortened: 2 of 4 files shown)


vbousquet:
Sorry for the delay, I'm a bit low on free time lately. I had a first look and it looks overall awesome.

One thing is that it does not fully conform to the overall data marshalling:
- `ScriptToCOMVariant` should be a mirror of `COMToScriptVariant` (here only one half is implemented, this is something I would need to add test for)
- Marshalling is supposed to be applied between `ScriptVariant` and `COM Variant` while here you are using a custom string type (the good old C `char*`) with custom data lifecycle, instead of the `ScriptString` type (which was created to support all use cases by including the lifecycle `Release` hook)

Pyrrvs:
Switched the string array marshalling to `ScriptString` (with its per-element `Release` hook) instead of the raw `char*` layout. Both the `helloscript` examples and the test now follow the scalar string convention, and `ReleaseScriptVariant` walks the array to release each element before freeing the block.

On the mirror point: `ScriptToCOMVariant` (plugin->script return path, which is what this feature needs for MPF) now handles string/bool for 1D+2D, but `COMToScriptVariant` (the script→plugin argument path) currently only handles 1D arrays of a subset of integer types (`INT`/`UINT`/`INT8`/`INT16`/`UINT8`/`UINT16`): no 2D, and no float/double/int32/int64/string/bool. So the two directions were already asymmetric before this change.

How would you like to scope closing that gap: add the COM->Script direction for string/bool (1D+2D) in this PR to match, or handle the broader symmetry pass separately since you mentioned wanting to add tests for it?

toxieainc:
@vbousquet How should we follow up on this one? Seems like its close to being done..

francisdb:
> New doctest case passes on Windows (38/38 assertions, verified via CI)

Are you sure about "verified via CI"? CI builds everything through CMake, and the doctest project (vpx-test.vcxproj) only lives in the Visual Studio solution - nothing in the CMake builds or the workflows compiles or runs it, and there's no ctest step.

No objection to the test itself - I'd actually love to see a test framework wired into CMake and run on CI.

vbousquet:
> How would you like to scope closing that gap: add the COM->Script direction for string/bool (1D+2D) in this PR to match, or handle the broader symmetry pass separately since you mentioned wanting to add tests for it?

Ideally, there would be symmetry and tests, but I'm just lacking time to do everything that should be done. This is especially true for the tests which are not yet part of the CMake workflow, not cross-platform and not integrated anywhere in our CI. For example, the tests now fail (and this is likely the problem reported here: https://github.com/vpinball/vpinball/issues/3525) but we did not detect which commit broke the light rendering while just running the tests would have told us.

I emphasized the symmetry as I think it was needed to validate the memory allocation scheme and it is especially important for object & strings. It would still be nice to have it for other data types but I don't have the time to implement and validate it (in fact implementing it is obvious, the real time consuming task is not providing the code but validating it).

So, I would say, let's go ahead with your changes and we will improve later

PR #3392, opened 2026-05-12
Restore SDL video driver log

francisdb:
Logs the SDL3 video driver (e.g. "wayland", "x11", "windows") right after the SDL_INIT_VIDEO subsystem init in WinMain. Useful when diagnosing display-related issues - e.g. SDL3's heuristic of falling back to XWayland on Wayland sessions.

Originally added in ee2f55651, removed in 88baaccfc

### DIFF
--- a/src/core/main.cpp
+++ b/src/core/main.cpp
@@ -203,6 +203,9 @@ extern "C" int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE /*hPrevInstance*/, 
          // FIXME this is not correct as we may be running something else than the player (extract vbs, ...)
          exit(1);
       }
+      if (const char* const drv = SDL_GetCurrentVideoDriver()) {
+         PLOGI << "SDL video driver: " << drv;
+      }
 
       // Run the application
       if (cmdLine.m_command)


jsm174:
@francisdb, hate to ask, but can you surround plog with braces. Even though single line, for some reason plogs throw compile warnings on mac.

jsm174:
ie, 

Screenshot 2026-05-12 at 5 25 41 PM

francisdb:
_You know what my response is going to be. Make CI fail on it. Oh and make sure there is CI for PR's. This is the only project I know that asks PR writers to set up their own CI._

Anyway, will fix it in the morning.

jsm174:
- If we failed builds for warnings in a C/C++ project, we'd never get any builds.

- As for the CI on PR, we've already discussed this numerous times. We are already at our budget. Every PR would evict the cache and cause hour long builds.

Screenshot 2026-05-12 at 8 13 10 PM

- I took care of the change. Sorry I asked.

francisdb:
> * If we failed builds for warnings in a C/C++ project, we'd never get any builds.
> 
>     * As for the CI on PR, we've already discussed this numerous times. We are already at our budget. Every PR would evict the cache and cause hour long builds.

Both points are real but solvable: linters can disallow use of specific constructs and proper cache management can fix the cache. There are ways to request a bigger cache if you can prove there's nothing that can be improved.

I can have a look at setting that up but only if there's willingness to improve things.

PR #3404, opened 2026-05-15
Don't save invalid table version (still keeping the hacky version section though)

### LINKED ISSUE #3403: ini Version section contains ` = `
vpinball master

Not sure how this happened but my ini has this:

```ini
[Version]
; VPX Version: VPX version that saved this file [Default: '10819999']
VPinball = 10819999
...
 = 
```

My parser fails on this as it's strictly not valid ini I think.

vbousquet:
Fix https://github.com/vpinball/vpinball/issues/3403


francisdb:
> frontend are not supposed to read internal settings file

That's among others for finding out where windows are configured and how pinmame is set up

vbousquet:
> > frontend are not supposed to read internal settings file
> 
> That's among others for finding out where windows are configured and how pinmame is set up

Yes, but this stays a very hacky and bad design that leads to problems, difficult setup & learning path. This is something that should be solved, not maintained. But still, on my side, this is low priority

PR #3406, opened 2026-05-16
altsound: fix audio buffer underruns on windows mingw

### LINKED ISSUE #3376: altsound is garbled with bgfx windows mingw build only
**Describe the bug**

I recently noticed that altsound is garbled with the bgfx windows _mingw_ build only. If I switch to the bgfx windows build (non-mingw) then altsound works perfectly. Also, if I disable altsound then bgfx windows _mingw_ build works perfectly.

**To Reproduce**
* Play a table with altsound with the bgfx windows _mingw_ build
  * game [Fish Tales (Williams 1992)](https://virtualpinballspreadsheet.github.io/games?game=tV1GotAP)
  * table [9FuFBUxfy8](https://vpuniverse.com/files/file/18293-fish-tales-vpw)
  * altsound [MYsBNDl3](https://vpuniverse.com/files/file/5937-altsound-20-%E2%80%93-fish-tales-williams-1992-reorchestrated)
* Wait for table attract mode
* Sound will be "garbled"

**Expected behavior**

Sound should be clear

**Versions & Settings**
 - VPX: Any version since _mingw_ build introduced from 10.8.1-4839-afbf695 (latest tried 10.8.1-5041-aa243b)
 - Windows version 11; 25H2 (Build 26200.8328)

gitfool:
Fixes #3376 (following up with learnings from https://github.com/vpinball/libaltsound/pull/11)

## Summary
- Increase AltSound buffer size for MinGW builds to reduce underruns and garbled playback.
- Keep existing smaller buffer size for non-MinGW builds to avoid added latency.

## Risk/Side Effects
- MinGW builds may have slightly higher audio latency due to larger buffer size.
- Non-MinGW builds are unchanged.

## Testing
- Windows MinGW (BGFX Release): AltSound playback now clean.
- Played Fish Tales before and after changes
- See attached files for more validation

https://github.com/user-attachments/assets/9d4e697a-a182-488c-bc0b-8d2b8a73150d

https://github.com/user-attachments/assets/bfe96cc9-d287-4ed4-9868-0c3bbc199f9b

### DIFF
--- a/plugins/altsound/AltSoundPlugin.cpp
+++ b/plugins/altsound/AltSoundPlugin.cpp
@@ -128,7 +128,9 @@ static void StartAltSound(const string& gameId, const string& basePath, uint64_t
 
     LOGI(std::format("Initializing AltSound for game: {}, basePath: {}", gameId, basePath));
 
-    if (AltSoundInit(basePath, gameId, 44100, 2, 128)) {
+    // Avoid audio buffer underruns on windows mingw builds by using a larger buffer size
+    // (128 samples is too small for some reason; tried 256/512 but only 1024 works fine)
+    if (AltSoundInit(basePath, gameId, 44100, 2, 1024)) {
         AltSoundSetAudioCallback(AudioCallback, nullptr);
         AltSoundSetHardwareGen(static_cast<ALTSOUND_HARDWARE_GEN>(hardwareGen));
 


vbousquet:
@jsm174 is the one who really knows the AltSound stuff.

As much as possible, we should avoid the `#define` per platform without a clear knowledge of the code (it makes the code less readable, less testable, and supporting error report quite difficult since it it become highly platform dependent). Is there a reason to increase the buffer size only for MINGW and not other platforms ?

gitfool:
@vbousquet I was being cautious, trying not to impact any other builds. I'd rather remove the condition but I wasn't sure about the impact in terms of latency/performance.

jsm174:
I agree that we shouldn't really try do platform specific changes in this area. I would hope we could come up with one value that would provide the least latency on all platforms.  We did have issues with Android in the past, but that was more on the SDL side.

gitfool:
@jsm174 With libaltsound testing it started at 768 but was still slightly garbled. I didn't try in between, say 896. Where did 128 come from in the first place? I could just remove the condition so all targets use 1024 as a starting point, then we wait and see what happens in the field. Thoughts?

gitfool:
Removed the target platform conditional since that's undesirable.

gitfool:
Could another option be to make this configurable and default to pre-existing value of 128?

vbousquet:
I just had a look to this issue and this other one: https://github.com/vpinball/vpinball/pull/3441 / https://github.com/vpinball/vpinball/issues/3440. For these 2 situations we are using miniaudio but not its engine layer, implementing a custom, very rough mixer thread.

I would say that a fix would be to replace the basic mixer in altsound.cpp/AudioMixingThread by using the default miniAudio mixing (ma_engine, like in VPX AudioPlayer). This should allow a more robust and cleaner result as we would benefit from miniAudio quality without needing to recreate and maintain everything on our side.

PR #3420, opened 2026-05-21
fix: only start ball-control drag on a real in-play ball

francisdb:
Ball control registers the most recently plunger/kicker-touched ball as the drag target without checking it can actually be dragged, so it could latch onto a ball that isn't in play:
- an invisible helper ball (e.g. Rollercoaster Tycoon's hidden SpinnerBall),
- a ball held in a kicker such as the trough/saucer/lock (which the physics engine already refuses to move, HitBall::UpdateVelocities).

Only start a drag when the target is visible and not held in a kicker, re-checked each frame since the registered ball's state changes. This also stops a click with no valid target from getting stuck "dragging" nothing, which blocked further selection until a flipper press.

_I have another pr coming that visually shows what ball is selected which improves things even more. But that needs more discussion._

### DIFF
--- a/src/ui/live/BallControl.cpp
+++ b/src/ui/live/BallControl.cpp
@@ -35,6 +35,15 @@ void BallControl::SetMode(Mode mode)
    m_mode = mode;
 }
 
+bool BallControl::IsDraggedBallSelectable() const
+{
+   // A ball is only a valid drag target while it is in active play: visible and not held in a kicker
+   // (trough/saucer/lock). Checked at use time rather than registration because a ball's state changes
+   // (e.g. trough balls are briefly unlocked while being shuffled along the trough). Matches the physics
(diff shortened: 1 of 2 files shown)


francisdb:
I was debugging `Rollercoaster Tycoon` and could not get the main ball to be selected because the invisible one was selected. This was all very confusing as you have no clue why ball control is not working.

I still find it annoying that the ball needs to be plunged / kicked before you can control it.

vbousquet (commented):
Thanks for the PR. I have just a very small request: if I understand well, the right name for the new function should be `IsSelectedBallDraggable()` not `IsDraggedBallSelectable()`. Could you update it (I can do it otherwise) ?

vbousquet:
Thanks!

PR #3421, opened 2026-05-21
feat: show a ring on the ball-control drag target

francisdb:
Ball control silently steers the most recently plunger/kicker-registered ball, which is easy to mistake (e.g. a trough/captured ball). When the FPS/stats overlay is shown, draw a colored ring over the current drag target so it is clear which ball is grabbed: yellow = grabbable, red = locked in a kicker/trough (cannot be dragged), green = being dragged.

I for example found it strange that I needed to plunge for a ball to be selected.

It's currently linked to `When the FPS/stats overlay is shown` because I was not sure what mode should show this. Could be it's own button or linked to the show balls mode `o`.


francisdb:
Updated to only temporarily show the ring

francisdb:
🪦

PR #3424, opened 2026-05-22
feat: log script file paths when read or loaded

francisdb:
Make it clear which script files (and from which resolved path) a table actually pulls in:

- `GetTextFile` logs `"Reading script: <path>"` for every .vbs it reads (the ExecuteGlobal GetTextFile include chain: controller/core/sega/...). It says "reading" because the caller is not guaranteed to execute it.
- `LoadScriptOverride` logs `"Loading script: <path>"` for an external <table>.vbs sidecar that overrides the embedded table script.

The `fso.OpenTextFile` editor-fallback path lives in the VBScript runtime, not our source, so it is not covered.


francisdb:
This should help log-based debugging to find what scripts are in use and if the user has a sidecar script.

francisdb:
@vbousquet this is 4-line PR, should not take long to review.

PR #3426, opened 2026-05-22
standalone: log method name and HRESULT constant in proxy Invoke warn…

francisdb:
…ings

Port vpinball/libwinevbs#21 to the standalone IDispatch proxy generator. The <Class>::Invoke fallback log only showed raw dispId/hres numbers; it now resolves the dispId to its method name and the HRESULT to its symbolic constant, e.g.:

  Option (dispId=230 0x000000e6), wFlags=3, hres=0x80004005 (E_FAIL)

IDLParserToCpp emits a per-class <Class>_dispid_name() (with "::" in nested class names sanitized to "_") and a shared hresult_name() helper, and uses both in the Invoke log. Regenerated vpinball_standalone_i_proxy.cpp.


francisdb:
looks like this was already taken care of in #3425

PR #3433, opened 2026-05-28
diag: log desktop session, window placement and move events

Le-Syl21:
## Why this PR

When a Linux user reports a window-placement issue (« mes fenêtres ne s'affichent pas sur le bon écran », « la playfield apparaît derrière la backglass »…), the very first thing any maintainer needs is *context that isn't in the code*: which compositor, which session type, what did SDL actually do, where did the WM end up putting the window. Today that information has to be teased out of the user one round-trip at a time.

This PR is a **small, behavior-free diagnostic patch** that makes future bug reports immediately triagable. It adds **four log lines** and **zero functional code**.

### 1. Desktop session at startup (Linux-only, ~1 line of log)

Right after the existing `"SDL video driver: ..."` line in `main.cpp`:

```
[INFO] SDL video driver: wayland
[INFO] Linux desktop session: XDG_CURRENT_DESKTOP=GNOME XDG_SESSION_TYPE=wayland WAYLAND_DISPLAY=wayland-0
```

That single line tells a reviewer the *family* of WM (`GNOME` / `KDE` / `Hyprland` / `sway` / `X-Cinnamon` / `XFCE` / `labwc`…) and whether it's an X11 or Wayland session — the two pieces of information that drive most placement-bug triage. Gated to non-Windows / non-Apple / non-mobile since `XDG_*` doesn't exist elsewhere.

### 2. Requested vs WM-placed position, once per window (universal)

On the first `Show()` of each window, after `SDL_SyncWindow`:

```
[INFO] Window #0 (Playfield) first mapped: requested display=2 | WM placed display=2 at (3840,0) 1920x1080
[INFO] Window #1 (Backglass) first mapped: requested display=1 | WM placed display=1 at (1920,0) 1920x1080
```

Symmetrical complement to the already-existing `"Window #X was created on display Y"` log (which prints the *intent* at creation time, before the window is mapped). The new line prints the **reality** after the WM has placed it. When the two displays match → no bug. When they diverge → support sees the misplacement in plain text, no need to ask the user to confirm anything.

Works on all SDL platforms — placement bugs aren't unique to Linux, just more frequent there.

### 3. Runtime move / display-change events (universal)

Two new cases in `Player::ProcessOSMessages`:

```
[INFO] Window 'Playfield' moved to (1920,0)
[INFO] Window 'Playfield' moved to display ID 3
```

Previously *neither* `SDL_EVENT_WINDOW_MOVED` nor `SDL_EVENT_WINDOW_DISPLAY_CHANGED` was observed anywhere in VPX. Any post-mapping relocation — by the WM applying its rules late (e.g. Hyprland `windowrulev2`), by a user dragging, by an external tool (`wmctrl`, `hyprctl dispatch`…), or by a monitor hotplug — went silently unrecorded. Now every such event is a one-line trace.

## What it costs

Effectively nothing:

- **Startup**: 1 extra log line (Linux) + 1 line per window first mapped (typically 1–4 windows). Total: 2–5 lines per app launch, spread across init.
- **Runtime**: one log line per real move/display-change event — these are *user-initiated* or *WM-initiated* events, so they fire at human-scale frequency (seconds to minutes apart), not per-frame.
- **Hot path**: nothing. The event-loop additions only fire on the corresponding event types, which are inert during normal gameplay.
- **PLOG level**: all `PLOGI` (info), filterable like any other info log if needed.

## What it doesn't do

This PR is intentionally narrow:

- **No behavior change.** No window is moved, resized, fullscreened, or placed differently. Pure observation.
- **No new dependencies.** Uses only `SDL_getenv`, `SDL_GetCurrentVideoDriver`, `SDL_GetWindowPosition`, `SDL_GetDisplayForWindow`, `SDL_SyncWindow`, `SDL_GetWindowFromID`, `SDL_GetWindowTitle` — all already in use elsewhere.
- **No PII.** `XDG_CURRENT_DESKTOP` and friends are user-config labels (`GNOME`, `KDE`, `Hyprland`…); `WAYLAND_DISPLAY` is a socket name (`wayland-0`). No paths, no usernames.
- **No platform-specific code in universal sections.** Only item (1) is `#ifdef`'d to Linux; items (2) and (3) are pure SDL and useful on every platform that builds with SDL.

## Why this is worth merging *now*

A current open PR (#3366) is debugging a Mutter multi-monitor placement issue, and the conversation keeps stalling on « please tell me what compositor and session you're on », « did the WM move your window after creation, and when », « did our code's correction actually fire ». With these four log lines, every future report carries that context by construction — both for this bug and for whatever the next exotic compositor surfaces.

## Test plan

- [x] Built locally (`make vpinball`, BGFX Linux x64) — no warnings, no functional change.
- [x] Logs verified to format correctly (`PLOGI` already in use throughout the codebase).
- [ ] Smoke test on Windows / macOS via CI — purely additive, no behavior path touched.

### DIFF
--- a/src/core/main.cpp
+++ b/src/core/main.cpp
@@ -200,6 +200,17 @@ extern "C" int WINAPI WinMain(HINSTANCE hInstance, HINSTANCE /*hPrevInstance*/, 
       // The video subsystem is initialized lazily when a window is created (see VPX::Window), so
       // headless commands (info, script/POV export, audit, tournament validation) run without a
       // display or a working video driver.
+#if !defined(_MSC_VER) && !defined(__APPLE__) && !defined(__ANDROID__)
+      {
+         const char* const xdgDesktop = SDL_getenv("XDG_CURRENT_DESKTOP");
+         const char* const xdgSession = SDL_getenv("XDG_SESSION_TYPE");
+         const char* const waylandDpy = SDL_getenv("WAYLAND_DISPLAY");
+         PLOGI << "Linux desktop session: "
+               << "XDG_CURRENT_DESKTOP=" << (xdgDesktop ? xdgDesktop : "(unset)")
+               << " XDG_SESSION_TYPE=" << (xdgSession ? xdgSession : "(unset)")
+               << " WAYLAND_DISPLAY=" << (waylandDpy ? waylandDpy : "(unset)");
+      }
+#endif
 
       // Run the application
       if (cmdLine.m_command)
--- a/src/core/player.cpp
+++ b/src/core/player.cpp
@@ -1489,6 +1489,19 @@ void Player::ProcessOSMessages(const bool isInitialized)
          break;
       }
 
+      case SDL_EVENT_WINDOW_MOVED: {
+         SDL_Window* const w = SDL_GetWindowFromID(e.window.windowID);
+         const char* const title = (w != nullptr) ? SDL_GetWindowTitle(w) : nullptr;
+         PLOGI << "Window '" << (title ? title : "(unknown)") << "' moved to (" << e.window.data1 << ',' << e.window.data2 << ')';
+         break;
+      }
+      case SDL_EVENT_WINDOW_DISPLAY_CHANGED: {
+         SDL_Window* const w = SDL_GetWindowFromID(e.window.windowID);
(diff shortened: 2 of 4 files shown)


vbousquet:
The codebase already log the display per window creation. As much as possible, we avoid to add platform specific code to keep the codebase as clean as possible. Therefoe, I will close this one

PR #3434, opened 2026-05-28
linux: restore fullscreen for BGFX

francisdb:
On Linux BGFX the playfield stopped going fullscreen: a screen-sized borderless window is not promoted to fullscreen by the compositor the way Windows does, so it ended up placed under panels and clipped off-screen.

Bring back the Fullscreen setting and toggle as fullscreen-desktop (no video mode change), default it on for the playfield, and leave Windows+BGFX unchanged.

To be clear, I agree that `exclusive fullscreen` (mode changes) can be dropped. But there should still be a way to indicate to the compositor on linux that vpinball wants to take up the whole screen.

Somewhat related to #3366 but only focusing on fullscreen and not doing the heuristics

Example of what it looks like for a default config + score/backglass set up in-window  (mind the missing part at the bottom)

image

### DIFF
--- a/src/core/Settings.cpp
+++ b/src/core/Settings.cpp
@@ -153,6 +153,12 @@ void Settings::UpdateDefaults()
          reg.Register(GetWindow_FSWidth_Property(i)->WithDefault(conf.width));
          reg.Register(GetWindow_FSHeight_Property(i)->WithDefault(conf.height));
          #endif
+         #if defined(ENABLE_BGFX) && !defined(_MSC_VER)
+         // On Linux/macOS a screen-sized borderless window is not promoted to fullscreen by the
+         // compositor, so default the playfield to fullscreen(-desktop) for a usable fresh install.
+         if (i == VPXWindowId::VPXWINDOW_Playfield)
+            reg.Register(GetWindow_FullScreen_Property(i)->WithDefault(true));
+         #endif
          reg.Register(GetWindow_Width_Property(i)->WithDefault(i == 0 ? conf.width : (conf.width / 4)));
          reg.Register(GetWindow_Height_Property(i)->WithDefault(i == 0 ? conf.height : min(conf.width * 4 / 9, conf.height)));
          break;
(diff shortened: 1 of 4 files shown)


jsm174:
fwiw, as of https://github.com/vpinball/vpinball/commit/e05c54cbf0246a7d96468ba8fb018da30f3fec7b, setting a window to screen dimensions does work on MacOS. There are no offsets, etc. The only issue is the menu bar and dock show. I am able to control them via `NSApplication.shared.presentationOptions` and was planning a patch, but I'll hold off until all this Linux stuff is resolved.

francisdb:
@jsm174 Would it not be better to not work around this and let sdl3 handle it by setting the proper fullscreen flag?

jsm174:
tbh, idk. MacOS fullscreen has always been clunky (not just in vpx). It's behavior seemed to change in 14, 15, and 26.

vbousquet:
Fullscreen as it was in VPX is really a legacy, deprecated thing. It used to make sense when there was no compositor, for mostly single app OS/use, with video mode/framerate changes. I don't think any modern OS compositor still fully support/allow this.

As far as I can see, the discussion on the Linux branch is how to patch Wayland with a solution to honor window positionning (which has nothing to do with fullscreen, we are just evaluating using its messages to get a not too hacky solution when the window covers the whole display).

So I would really avoid to enable it back as this would move us backward.

vbousquet:
To be more precise on my review: I have nothing against setting SDL3 fullscreen. What I think should be entirely avoid is fullscreen as a user experience option, fullscreen with mode change (as it will fail partly or fully more or less everywhere).

francisdb:
So the only option is detecting that the user did not set specific sizes and giving that a fullscreen flag?

Don't all games these days have a windowed vs full screen option?

image

image

vbousquet:
Fullscreen as a user option is legit if it actually allows to manage fullscreen, that is to say take ownership of the screen and manage its video mode (resolution and refresh rate). We do not support this for BGFX (DX9 does) so in our situation, the 'fullscreen' option does not actually manage fullscreen but in fact is a way to enable a hack around an unfixed bug regarding window positionning on a specific setup. What we need to do is solve this bug without any UX impact on any platform.

francisdb:
I'm not interested in modes, those don't exist on wayland. What condition should set the fullscreen flag on SDL that removes the decoration and allows the window over the menu bars / launcher?

vbousquet:
I have not yet taken the time to fully read and understand the discussion on this bug and corresponding PR https://github.com/vpinball/vpinball/pull/3366. So, with my so far limited understanding, I would guess it may be summed up like this:
- unlike all other OS/desktop compositors, Wayland honor size but won't honor display, nor position,
- when size match target display size, we may promote the window to a so called 'fullscreen' window (which in fact, it is not as I don't think it get any exclusive access to the display) as this will workaround Wayland limitations without side effects on other platforms.

Switching modes is the UX behavior I depict as deprecated. I think this is somewhat outdated as it gives bad user experience (to a point that some OS are prohibiting it). That's one of the reasons I removed it for BGFX

francisdb:
This was re-implemented by 6bba0bfb1344eb1dd4bb5d37fbeb06cbc0ac4b8f

PR #3437, opened 2026-05-28
flexdmd: fix plugin path resolution on Windows/MinGW

### LINKED ISSUE #3436: FlexDMD Windows MinGW Font Asset Bug
## Symptom

On Windows MinGW builds only, the FlexDMD plugin fails to load built-in font assets at startup for [Blood Machines Original (2022)](https://virtualpinballspreadsheet.github.io/games?game=Iox4EmTcwu):

```
2026-05-23 12:22:15.339 ERROR [38800] [VPXPluginAPIImpl::PluginLog@246] Asset not loaded: udmd-f7by13.fnt
```

Does **not** reproduce on MSVC (Windows non-MinGW) or macOS/Linux builds.

## Root Cause

In `plugins/flexdmd/common.h`, `PATH_SEPARATOR_CHAR` is defined as:

```cpp
#ifdef _MSC_VER
#define PATH_SEPARATOR_CHAR '\\'
#else
#define PATH_SEPARATOR_CHAR '/'
#endif
```

MinGW defines `_WIN32` but **not** `_MSC_VER` (it uses GCC targeting Windows), so `PATH_SEPARATOR_CHAR` is `'/'` on MinGW.

In `GetPluginPath()` (`plugins/flexdmd/common.cpp`):

```cpp
const size_t lastSep = pathBuf.find_last_of(PATH_SEPARATOR_CHAR);  // searches for '/'
if (lastSep == string::npos)
    return string();   // ← always hit on MinGW
```

`GetModuleFileNameA` is a Win32 API that **always** returns backslash-separated paths (`C:\...\plugins\flexdmd\FlexDMD.dll`) regardless of compiler. So `find_last_of('/')` finds nothing, `lastSep == string::npos`, and `GetPluginPath()` returns `""`.

In `AssetManager::Open` the FlexResource path then becomes:
```
"" + "assets" + "/" + "udmd-f7by13.fnt"  →  "assets/udmd-f7by13.fnt"   (no directory)
```
which cannot be found.

**Why the other builds were unaffected:**
- **MSVC:** `_MSC_VER` defined → `PATH_SEPARATOR_CHAR == '\\'` → matches the backslash paths `GetModuleFileNameA` returns.
- **macOS/Linux:** `dladdr`/`realpath` return forward-slash paths → `PATH_SEPARATOR_CHAR == '/'` → matches correctly.

## Fix

`plugins/flexdmd/common.cpp`, end of `GetPluginPath()` — accept either separator, matching the pattern already used in the `pinup-events` and `pup` plugins:

```cpp
    // Accept both separators: POSIX paths use '/', Win32 APIs (GetModuleFileNameA) always
    // return '\' regardless of compiler, and PATH_SEPARATOR_CHAR is '/' on MinGW.
    const size_t lastSep = pathBuf.find_last_of("\\/");
    if (lastSep == string::npos)
        return string();

    return pathBuf.substr(0, lastSep + 1);
```

An earlier iteration used `#ifdef _WIN32` to hard-code `'\\'` on Windows. That also fixes the bug but the `"\\/"` form is simpler and consistent with the rest of the codebase.

## Other Plugins Audited

All other plugins were checked for the same pattern:

| Plugin | `GetPluginPath` / path resolution | Affected? |
|---|---|---|
| `flexdmd` | `find_last_of(PATH_SEPARATOR_CHAR)` → **fixed** to `"\\/"` | ✓ fixed |
| `pinup-events` | `find_last_of(_T("\\/"))` | safe |
| `scoreview` | `std::filesystem::path::parent_path()` | safe |
| `wmp`, `vni`, `serum` | no `GetPluginPath`; use `PATH_SEPARATOR_CHAR` only in `normalize_path_separators` for user-supplied paths, not for Win32 API output | safe* |
| `altsound`, `pup` | `std::filesystem::path::parent_path()` | safe |

`flexdmd` was the only affected plugin.

## Follow-up: `normalize_path_separators` in `wmp` / `vni` / `serum`

\* The `normalize_path_separators` helper in these plugins replaces one separator with the other based on `PATH_SEPARATOR_CHAR`, then collapses runs of duplicate separators. This works for typical relative or absolute paths but has at least two known gaps:

- **POSIX paths on MinGW** — `PATH_SEPARATOR_CHAR` is `'/'` on MinGW, so backslashes in a user-supplied Windows path (e.g. `tables\mygame\rom`) would be left as-is instead of being converted, potentially causing `std::filesystem` lookups to fail in the same way as this bug.
- **UNC paths** (`\\server\share\...`) — collapsing duplicate separators will corrupt the leading `\\`, turning `\\server\share` into `\server\share`.

These should be fixed in a follow-up: either replace `normalize_path_separators` with `std::filesystem::path` construction (which handles both cases natively), or special-case the UNC prefix and apply the same `"\\/"` dual-separator approach used here.

gitfool:
Fixes #3436 

GetModuleFileNameA (Win32 API) always returns backslash-separated paths regardless of the compiler or runtime. PATH_SEPARATOR_CHAR is defined as '/' on MinGW builds, so find_last_of(PATH_SEPARATOR_CHAR) failed to find the directory separator, causing GetPluginPath() to return an empty string and the FlexDMD plugin to fail to locate its assets.

Fix by searching for both separators with find_last_of("\\\\/"), which correctly handles both Win32 backslash paths and POSIX forward-slash paths without conditioning on the build environment.

Logs before and after fix:
* [vpinball-windows-mingw-blood-machines-err.log](https://github.com/user-attachments/files/28369120/vpinball-windows-mingw-blood-machines-err.log)
* [vpinball-windows-mingw-blood-machines-fix.log](https://github.com/user-attachments/files/28369121/vpinball-windows-mingw-blood-machines-fix.log)

### DIFF
--- a/plugins/flexdmd/common.cpp
+++ b/plugins/flexdmd/common.cpp
@@ -206,7 +206,9 @@ string GetPluginPath()
     if (pathBuf.empty())
         return string();
 
-    const size_t lastSep = pathBuf.find_last_of(PATH_SEPARATOR_CHAR);
+    // Accept both separators: POSIX paths use '/', Win32 APIs (GetModuleFileNameA) always
+    // return '\' regardless of compiler, and PATH_SEPARATOR_CHAR is '/' on MinGW.
+    const size_t lastSep = pathBuf.find_last_of("\\/");
     if (lastSep == string::npos)
         return string();
 


vbousquet:
Closed as the fix was done by using the modern codepath used in other plugins (see https://github.com/vpinball/vpinball/commit/6050b604b7fa21d6e2656aeecb76208fac9c1edb) and keeping std::string here was not a good thing

PR #3448, opened 2026-05-31
flexdmd: don't prepend base path to absolute asset paths

francisdb:
AssetManager::ResolveSrc unconditionally prefixed the base path (./) to file asset sources, so an absolute path from the script (e.g. FlexPath built from fso.GetAbsolutePathName) became .//home/... and failed to load. Skip the prefix when the source is already absolute, matching the original .NET FlexDMD Path.Combine behavior.

Fixes loading of https://vpuniverse.com/files/file/27952-the-grinchs-how-to-steal-christmas/ on standalone


vbousquet:
Thanks!

PR #3455, opened 2026-06-01
Fix normalize_path_separators destroying UNC path prefix

### LINKED ISSUE #3454: `normalize_path_separators` destroys UNC paths
## Summary

`normalize_path_separators` collapses the leading `\\` of a UNC path (e.g. `\\server\share\...`) to a single `\`, making the path invalid. Any caller that passes a UNC path through this function will silently fail to resolve the file.

## Root cause

The function uses `std::unique` to deduplicate consecutive path separator characters. On Windows `PATH_SEPARATOR_CHAR` is `'\\'`, so the predicate also deduplicates the *intentional* double-separator that forms a valid UNC prefix, turning `\\server\share` into `\server\share`:

```cpp
auto end = std::unique(szResult.begin(), szResult.end(),
   [](char a, char b) { return a == b && a == PATH_SEPARATOR_CHAR; });
```

## Observed symptom

Reproduced with the [Blood Machines (Original 2022)](https://virtualpinballspreadsheet.github.io/games?game=Iox4EmTcwu) table when VBScript sets the music path using:

```vbscript
MusicDir = MusicDirectory("Blood Machines OST")
```

`MusicDirectory` returns a fully-qualified UNC path (e.g. `\\nas.local\Emulators\Pinball\Tables\Blood Machines (Original 2022)\music\Blood Machines OST\`) which is then passed as-is to `WMPlayer.OCX.URL`. The WMP plugin fails silently to load audio in this case. The log shows:

```
WMP Core URL set to: \\nas.local\Emulators\...\song.mp3
Volume updated: 80 -> 0.80
```

But **no** "Loading audio file:" line follows. `WMPCore::SetURL` calls `find_case_insensitive_file_path`, which calls `normalize_path_separators`, gets back a broken path, fails `filesystem::exists`, then fails the recursive parent walk, and returns empty — so `LoadAudio` is never called.

Relative paths work because they don't start with `\\`, so deduplication is harmless.

Network-mapped drive letters (e.g. `Z:\path\to\file.mp3`) also trigger this bug. The drive letter is resolved to its underlying UNC path (`\\nas.local\Emulators\...`) by `std::filesystem::weakly_canonical()` inside `CommandLineProcessor::GetPathFromArg` ([src/core/AppCommands.cpp](src/core/AppCommands.cpp#L391)):

```cpp
path = std::filesystem::absolute(path);
path = std::filesystem::weakly_canonical(path);   // Z:\ → \\server\share\
```

This runs on the table's file path at launch time. The result is stored as `table->m_filename`. `FileLocator::GetTablePath` then derives the music folder from `table->m_filename.parent_path()`, so `MusicDirectory()` already returns a UNC path even when the user opened the table via a drive letter. The symptom is identical: `normalize_path_separators` receives a string beginning with `\\` and destroys the prefix.

## Affected files (all copies are identical)

| File | Line |
|---|---|
| `src/core/def.cpp` | 592 |
| `plugins/wmp/common.cpp` | 24 |
| `plugins/vni/common.cpp` | 57 |
| `plugins/serum/common.cpp` | 57 |
| `plugins/pup/common.cpp` | 155 |
| `plugins/flexdmd/common.cpp` | 83 |

## Fix

In each copy, start the `std::unique` iterator after the first two characters when the path begins with a double separator (UNC prefix):

```cpp
string normalize_path_separators(const string& szPath)
{
   string szResult = szPath;

   #if '/' == PATH_SEPARATOR_CHAR
      std::ranges::replace(szResult.begin(), szResult.end(), '\\', PATH_SEPARATOR_CHAR);
   #else
      std::ranges::replace(szResult.begin(), szResult.end(), '/', PATH_SEPARATOR_CHAR);
   #endif

   // Skip the UNC prefix (\\server or //server) to avoid collapsing it to a single separator
   const bool isUNC = szResult.size() >= 2 && szResult[0] == PATH_SEPARATOR_CHAR && szResult[1] == PATH_SEPARATOR_CHAR;
   auto end = std::unique(isUNC ? szResult.begin() + 2 : szResult.begin(), szResult.end(),
      [](char a, char b) { return a == b && a == PATH_SEPARATOR_CHAR; });
   szResult.erase(end, szResult.end());

   return szResult;
}
```

gitfool:
Fixes #3454

std::unique with the PATH_SEPARATOR_CHAR predicate was collapsing the intentional leading \\\\ of UNC paths (\\\\server\\share\\...) to a single \\, making the path invalid and silently breaking file resolution on network shares and mapped drives that resolve to UNC via weakly_canonical().

Logs before and after fix:
* [vpinball-windows-bloodmachines-err.log](https://github.com/user-attachments/files/28448697/vpinball-windows-bloodmachines-err.log) (wmp; silent)
* [vpinball-windows-bloodmachines-fix.log](https://github.com/user-attachments/files/28448701/vpinball-windows-bloodmachines-fix.log)
* [vpinball-windows-strangerthings-err.log](https://github.com/user-attachments/files/28448705/vpinball-windows-strangerthings-err.log) (flexdmd)
* [vpinball-windows-strangerthings-fix.log](https://github.com/user-attachments/files/28448707/vpinball-windows-strangerthings-fix.log)

### DIFF
--- a/plugins/flexdmd/common.cpp
+++ b/plugins/flexdmd/common.cpp
@@ -90,7 +90,8 @@ string normalize_path_separators(const string& szPath)
       std::ranges::replace(szResult.begin(), szResult.end(), '/', PATH_SEPARATOR_CHAR);
    #endif
 
-   auto end = std::unique(szResult.begin(), szResult.end(),
+   const bool isUNC = szResult.size() >= 2 && szResult[0] == PATH_SEPARATOR_CHAR && szResult[1] == PATH_SEPARATOR_CHAR;
+   auto end = std::unique(isUNC ? szResult.begin() + 2 : szResult.begin(), szResult.end(),
       [](char a, char b) { return a == b && a == PATH_SEPARATOR_CHAR; });
    szResult.erase(end, szResult.end());
 
--- a/plugins/pup/common.cpp
+++ b/plugins/pup/common.cpp
@@ -162,7 +162,8 @@ string normalize_path_separators(const string& szPath)
       std::ranges::replace(szResult.begin(), szResult.end(), '/', PATH_SEPARATOR_CHAR);
    #endif
 
-   auto end = std::unique(szResult.begin(), szResult.end(),
+   const bool isUNC = szResult.size() >= 2 && szResult[0] == PATH_SEPARATOR_CHAR && szResult[1] == PATH_SEPARATOR_CHAR;
+   auto end = std::unique(isUNC ? szResult.begin() + 2 : szResult.begin(), szResult.end(),
       [](char a, char b) { return a == b && a == PATH_SEPARATOR_CHAR; });
    szResult.erase(end, szResult.end());
 
--- a/plugins/serum/common.cpp
+++ b/plugins/serum/common.cpp
@@ -9,7 +9,7 @@
 #endif
 
 namespace Serum {
-   
+
 #ifdef _WIN32
 #ifndef WIN32_LEAN_AND_MEAN
 #define WIN32_LEAN_AND_MEAN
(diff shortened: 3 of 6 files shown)


jsm174:
UNC paths are a Windows-specific concept which is counter-intuitive to VPX becoming more cross platform.  

I'll defer to @vbousquet because of all the updated filesystem logic.

vbousquet:
The path handling used to be `char*` based then `std::string` but it really needs to be `std::filesystem::path` based (for cross platform, fix all the bugs with custom characters in path, cleaner code). In a perfect world, we should never have to deal with `PATH_SEPARATOR_CHAR` and `normalize_path_separators` should not even exists but it is always a balance between merging a quick fix and going deeper toward a clean codebase.

Perhaps one thing that we should define more clearly is that everything coming from or to the script must be Windows style based path, but anything else must be native, `std::filesystem::path` based (the code, the persisted settings, ...).

I will have a look when I get some time off.

PR #3457, opened 2026-06-01
fix: ignore out of range BackglassMode values

francisdb:
put_BackglassMode computed m_viewMode = pVal - DESKTOP without validation. A script passing a raw 0/1 index (as Thunderbirds does) instead of the DESKTOP..FULL_SINGLE_SCREEN enum planted an out of range value into m_viewMode, causing out of bounds reads of m_BG_image / mViewSetups: intermittent crash or grey non-rendering background. Ignore values outside BG_DESKTOP..BG_FSS.

With this PR applied instead of a crash the table works and you get these logs:

```
2026-06-01 11:12:35.798 ERROR [2013793] [PinTable::put_BackglassMode@7150] BackglassMode is deprecated
2026-06-01 11:12:35.798 ERROR [2013793] [PinTable::put_BackglassMode@7156] Ignoring out of range BackglassMode value 0
```

### DIFF
--- a/src/parts/pintable.cpp
+++ b/src/parts/pintable.cpp
@@ -7147,8 +7147,7 @@ STDMETHODIMP PinTable::get_BackglassMode(BackglassIndex *pVal)
 
 STDMETHODIMP PinTable::put_BackglassMode(BackglassIndex pVal)
 {
-   PLOGE << "BackglassMode is deprecated";
-   m_viewMode = (ViewSetupID)(pVal - DESKTOP);
+   PLOGE << "BackglassMode is deprecated and ignored, this call has no effect";
    return S_OK;
 }
 


vbousquet:
Thanks for the report and fix!

I just had a look and I think modifying `m_viewMode` will never end in a good situation (as it would need to regenerate all static prerendering, and the use case is somewhat obscur to me, beside I don't think this ever worked on  a running player as it used to be even more designed against static prerendering). I would say the best would be to not change `m_viewMode` at all, and just report deprecation in the log. What do you think ?

francisdb:
I looked at this from a narrow view. Just fixing the crash. Don't think I have enough context to make a decision here. Any solution is ok as long as this table does not crash.

vbousquet:
Ok. I will remove the set to viewmode this evening (or merge this PR if you are able to update before)

francisdb:
@vbousquet now only logs a deprecated error and indicates it has no effect

vbousquet:
Perfect, thanks !

PR #3459, opened 2026-06-01
fix: don't crash loading an auto-aspect image decal

francisdb:
`Decal::Load` calls `EnsureSize`, but for an image decal with auto aspect sizing that dereferences `m_ptable` to look up the image. `m_ptable` is only set later in `AddPart` (and the image is not loaded yet at that point), so loading such a decal crashed.

Guard the image lookup against a missing table, and (re)compute the size in `RenderSetup` once the decal is added and its images are available.

https://www.vpforums.org/index.php?app=downloads&showfile=13406

### DIFF
--- a/src/parts/decal.cpp
+++ b/src/parts/decal.cpp
@@ -325,7 +325,8 @@ void Decal::EnsureSize()
 
       if (m_d.m_decaltype == DecalImage)
       {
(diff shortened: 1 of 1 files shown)


vbousquet:
Thanks again!

PR #3462, opened 2026-06-01
report a meaningful error for a value-returning member used as a statement

francisdb:
Using a value-returning member (a property get, or a function) as a bare
statement on a plugin-provided object returned `E_POINTER`, which VBScript
can't map. Now returns `DISP_E_MEMBERNOTFOUND`, matching the error native
Windows gives.

Before:
```
Runtime error on line 411, col 32: Description unavailable
```

After:
```
Runtime error on line 411, col 32: Object doesn't support this property or method
```

Still an error, just reported properly. A TODO notes that a
value-returning *function* used this way should run instead, but it can't be
told apart from a property getter here.

https://vpuniverse.com/files/file/22310-medieval-castel-v10/
Shows this issue here, `Controller.Pause` should be assigned
```vbscript
Sub table1_unPaused
    If Isobject(Controller)Then Controller.Pause
End Sub
```

### DIFF
--- a/src/core/DynamicScript.cpp
+++ b/src/core/DynamicScript.cpp
@@ -900,7 +900,12 @@ HRESULT DynamicTypeLibrary::Invoke(const ScriptClassDef * classDef, void* native
    }
    const ScriptClassMemberDef& memberDef = classDef->members[memberIndex];
    if ((memberDef.type.id != TypeID::TYPEID_VOID) && (pVarResult == nullptr))
-      return E_POINTER;
+      // A value-returning member invoked with no result buffer was used as a statement (e.g. a bare
+      // property get). Report it as native COM does (-> VBScript error 438).
+      // TODO a value-returning function used as a bare statement should run and discard its result (as
+      // native does), but a property getter and a function are indistinguishable here (identical member
+      // shape), so such a call fails with 438 too. Telling them apart needs a flag in ScriptClassMemberDef.
+      return DISP_E_MEMBERNOTFOUND;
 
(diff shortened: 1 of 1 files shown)


vbousquet:
Just had a look and fixed it by supporting discarding unused returned value both for function and getter as it seems to be legal in VBS (only illegal for VBA as I read across internet and Claude is saying too). See https://github.com/vpinball/vpinball/pull/3463

If we really want to error on Getter not being assigned, we can do it by using the Invoke flag. But it seems ok as this

francisdb:
I had a similar solution in https://github.com/vpinball/vpinball/pull/3461 which I then abandoned after some more testing on windows :

```
● Native Windows results:

  --- script-defined class, bare statements ---
  1) property get   f.Bar       : OK
  2) function       f.Fn        : OK
  3) function       f.Fn()      : OK
  4) function w/arg f.FnArg 5   : OK
  5) function w/arg f.FnArg(5)  : OK
  6) sub            f.Sb        : OK
  7) sub            f.Sb()      : OK
  --- COM object property, bare statements ---
  8) rw property    d.CompareMode: ERROR 438 (Object doesn't support this property or method)
  9) ro property    d.Count      : ERROR 438 (Object doesn't support this property or method)
```

I leave it to you to decide if this requires parity with windows.

vbousquet:
Make sense with my test (I said VBA does not allow it but you are right, this is COM does not allow it).
Being slighty more permissive seems the right choice to me as it doesn't break any existing table (the famous backward compatibility) and should not cause issues for the future

PR #3469, opened 2026-06-02
dmd: Fix standalone DMD crashes from rendering on consumer threads

francisdb:
What does this solve

* random crashes on exiting a table
* artifacts/flicker on backglass / dmd during gameplay
* choppy gameplay

The DMD source `GetRenderFrame` callback is documented as thread safe, but the
standalone implementations rendered / touched live render state on the calling
thread. The dmdutil worker thread and the main-thread scoreview/flasher pulls
drove that concurrently, causing intermittent (some Release-only) crashes:

- `FlexDMD::Render` -> `Group::Draw` (SDL blit): the dmdutil worker rendered the
  FlexDMD scene while the main thread mutated and drew it (e.g. the end-of-game
  match sequence), also producing DMD corruption.
- `BaseTexture::GetAlias` (shared_ptr release): the worker read the global/flasher
  DMD alias cache while the main thread rewrote the texture and cleared it.

### Changes

- FlexDMD renders only on the main thread (`OnPrepareFrame`) and publishes a
  double-buffered snapshot; `GetRenderFrame` just returns it, never touching the
  scene/surface off the main thread.
- Global/flasher DMD (`ControllerOnGetRenderDMD`): serialize the texture alias
  cache with a per-texture mutex, and keep the returned frame alive in a
  `thread_local shared_ptr` so it can't be freed before the consumer copies it.
- Skip redundant `FlexDMD.RenderMode` sets: some tables re-assert it every frame
  (the Thunderbirds match sequence does), which tore down the DMD surface and
  re-broadcast "display source changed" every frame. Now a no-op when unchanged,
  matching `SetWidth`/`SetHeight`.

<details>
<summary>Expand for crash backtraces encountered when trying to reproduce issues reported by users</summary>

FlexDMD scene rendered on the dmdutil worker thread:

```
SDL_BlitSurfaceScaled
Flex::SurfaceGraphics::DrawImage   SurfaceGraphics.cpp:36
Flex::Font::DrawText_              Font.cpp:201
Flex::Group::Draw                  Group.cpp:60
Flex::FlexDMD::Render              FlexDMD.cpp:79
Flex::GetRenderFrame               FlexDMDPlugin.cpp:529
DMDUtilPlugin::UpdateThread        DMDUtilPlugin.cpp:120
```

Same race seen from the main thread (scoreview/flasher pull) while the worker
renders concurrently:

```
Flex::Group::Draw                  Group.cpp:60
Flex::FlexDMD::Render              FlexDMD.cpp:79
Flex::GetRenderFrame               FlexDMDPlugin.cpp:529
ResURIResolver::GetDisplayState    ResURIResolver.cpp:318
Flasher::Render                    flasher.cpp:1317
Renderer::RenderItem               Renderer.cpp:1668
Player::PrepareFrame               player.cpp:2127
```

Global/flasher DMD alias cache read on the dmdutil worker thread:

```
std::_Sp_counted_base::_M_release            shared_ptr_base.h:414
BaseTexture::GetAlias                         Texture.cpp:589
VPXPluginAPIImpl::ControllerOnGetRenderDMD    VPXPluginAPIImpl.cpp:630
DMDUtilPlugin::UpdateThread                   DMDUtilPlugin.cpp:126
```

</details>

Reproduce on these tables, with on-playfield backglass/scoreview, requires multiple game playthroughs sometimes. (I know that it's not the best coded tables but still should not crash vpinball)
https://vpuniverse.com/files/file/9417-thunderbirds-homepin-2018/

### DIFF
--- a/plugins/dmdutil/DMDUtilPlugin.cpp
+++ b/plugins/dmdutil/DMDUtilPlugin.cpp
@@ -110,42 +110,48 @@ static void UpdateThread()
       // Fixed update at 60 FPS
       std::this_thread::sleep_for(std::chrono::microseconds(16666));
 
-      std::lock_guard<std::mutex> lock(sourceMutex);
+      // Copy the shared source/display state under the lock, then release it before rendering.
+      // The state can be reset (e.g. the source removed during teardown, clearing GetRenderFrame)
+      // concurrently with this thread, so skip until it is valid. The render itself can take a
+      // millisecond or more (FlexDMD renders its scene in GetRenderFrame); holding sourceMutex
+      // across it would stall the main thread, which takes the same lock in onDmdSrcChanged.
+      DisplaySrcId dmdId;
+      {
+         std::lock_guard<std::mutex> lock(sourceMutex);
+         if (pDmd == nullptr || selectedDmdId.id.id == 0 || selectedDmdId.GetRenderFrame == nullptr)
+            continue;
+         dmdId = selectedDmdId;
+      }
 
-      // Read the shared source/display state under the lock: it can be reset (e.g. the source removed
-      // during teardown, clearing GetRenderFrame) concurrently with this thread. Skip until it is valid.
-      if (pDmd == nullptr || selectedDmdId.id.id == 0 || selectedDmdId.GetRenderFrame == nullptr)
-         continue;
-
-      const DisplayFrame frame = selectedDmdId.GetRenderFrame(selectedDmdId.id);
+      const DisplayFrame frame = dmdId.GetRenderFrame(dmdId.id);
       if (lastFrameID == frame.frameId)
          continue;
       lastFrameID = frame.frameId;
 
-      switch(selectedDmdId.frameFormat) {
+      switch(dmdId.frameFormat) {
          case CTLPI_DISPLAY_FORMAT_LUM32F:
          {
             const float* const __restrict luminanceData = static_cast<const float*>(frame.frame);
-            uint8_t* const __restrict rgb24Data = new uint8_t[selectedDmdId.width * selectedDmdId.height * 3];
+            uint8_t* const __restrict rgb24Data = new uint8_t[dmdId.width * dmdId.height * 3];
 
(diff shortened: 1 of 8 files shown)


francisdb:
@jsm174  Can I please get a review on this one, this is a real issue where there is visible flicker and tables crash

vbousquet:
I have started to review it but did not find time to finish as this is a complex change that needs careful review.

My main concern is that we must not put hack or fixes around the base threading contract. This contract is supposed to be simple to implement if it is not, we need to understand why and see what needs to be changed in it. This contract is that `GetRenderFrame` must supports async requests as long as the requests are guaranteed to occur between the display advertisement. If there is a problem, then it must be fixed to fulfill this contract. If the contract is the problem, then it must be clearly identified and all plugins must be changed.

I'm still very low on free time and have no way to review multiple PR per day. @jsm174  @toxieainc please feel free to process as much as you can. My main concern is to avoid as many 'hacks' and 'fixes' and instead favor refactors as this is what we have done so far, allowing us to build a more future aware architecture.

francisdb:
~~My pr is supposed to adhere to the contract where the previous code did not, it's been 2 weeks so let me also do a review.~~

toxieainc:
I'm a bit confused here: @francisdb Will there be another follow up commit, or is this PR still current as-is?

francisdb:
Don't plan any further work on this. Feel free to do whatever you want with it.

vbousquet:
I had a look to this PR but did not have time to do a proper full review. The thing is that the plugin threading has never been clean from the start. It was quickly setup to allow moving forward and avoiding the standalone fork, but needed a full review once the API would stabilize.

We are slowing moving toward this state where this review can be performed and the changes proposed in this PR should be part of this review, as they have to fit in. That is to say, before fixing a threading problem, the root of it must be reviewed (advertisement process for example which must be cleaned up first), then it must be reviewed against what we have searched to have on data dispatch (ideal being lock & copy free, then if impossible lock free with copy, worst being locked and copy).

francisdb:
🪦

PR #3471, opened 2026-06-02
render: fix bloom bleeding into embedded ancillary windows

### LINKED ISSUE #3467: Embedded backglass / scoreview strobe (flicker)
The embedded backglass and scoreview windows show a fast strobe/flicker. The playfield (same window/surface) is unaffected. Both aux windows flicker independently and at the same rate.

- Master
- VPX standalone, BGFX renderer
- OS / compositor: Linux, Wayland (SDL3), COSMIC
- GPU / driver: AMD Strix Halo
- Monitor refresh rate: 240 Hz

### Key observations

- Only the embedded aux windows strobe; the playfield never does.
- A **static** backglass image strobes too -> it is not a content/update issue.
- The flicker is **much less frequent (but not gone) when the window is focused /
  clicked**, and once when unfocused again I have the impression it does not get worse.
- The Scoreview DMD shader enhances the flicker, so it must come from the underlying buffer.

### Suspected cause (not confirmed)

The aux regions are cleared early in the frame and redrawn in a later pass
(after bloom/AO, before post-process), while the playfield is drawn early. The
present does not appear to hard-depend on that late aux redraw completing, so it
can be presented before the redraw finishes on the GPU -> the just-cleared (black)
region shows for that frame.

Question

Anyone else - especially on a **high-refresh display and/or Wayland** - seeing
the embedded backglass/scoreview strobe while the playfield is stable? Does
focusing/clicking the window or changing the sync mode change it for you?

Was testing with this table, both Backglass and ScoreView enabled and pretty big in left / right corner

https://vpuniverse.com/files/file/9417-thunderbirds-homepin-2018/
or

francisdb:
Embedded backglass/scoreview content is drawn into the shared back buffer after bloom has sampled it with the region cleared (ClearEmbeddedAncillaryWindow runs before UpdateBloom). The content pass had no dependency on the bloom pass, so the frame sorter could place it before the bloom sample on frames where the pass set changed (e.g. ball motion blur), making bloom bleed the ancillary content into the region: an intermittent brightness strobe on embedded ancillary windows.

Make the embedded ancillary content pass depend on the bloom pass so it is always ordered after the bloom sample.

fixes #3467
That table had a 1.8 bloom value that was affecting backglass and score view

### DIFF
--- a/src/renderer/Renderer.cpp
+++ b/src/renderer/Renderer.cpp
@@ -3396,6 +3396,12 @@ void Renderer::RenderAncillaryWindow(VPXWindowId window, const VPX::RenderOutput
    if (outputRT == nullptr)
       return;
 
+   // Keep the embedded ancillary content ordered after bloom: it writes the back buffer that bloom
+   // already sampled with this region cleared, so without this dependency the sorter may run it before
+   // the bloom sample and bloom bleeds the content into the region. No-op if bloom did not run.
+   if (output.GetMode() == VPX::RenderOutput::OM_EMBEDDED)
+      rd->AddRenderTargetDependency(GetBloomBufferTexture());
+
    rd->ResetRenderState();
    if (output.GetMode() == VPX::RenderOutput::OM_WINDOW)
       rd->Clear(clearType::TARGET | clearType::ZBUFFER, 0x00000000);


francisdb:
Simple fix, lot's of debugging required to find out it was this 😅. Learned a lot along the way (eg renderdoc).

PR #3473, opened 2026-06-02
ui: fix graphics backend dropdown listing unusable backends

### LINKED ISSUE #3466: Graphics backend reported Direct3D11 on Linux?
Master
Linux
Wayland

`F12`

Image

francisdb:
The in-game Graphics Backend dropdown reported e.g. `Direct3D11` on Linux when nothing was selected, listed a phantom "Default" backend, and offered backends that cannot run on the platform.

- Add WebGPU to bgfxRendererNames: it was added to bgfx after Vulkan, so the array was misaligned - WebGPU got mislabeled "Default" and the trailing Count slot was an empty string.
- Drop backends that are not usable here: `WebGPU` (web/experimental) everywhere, and Direct3D on non-Windows (Direct3D is Windows only), alongside the existing `Noop`/`Direct3D12` exclusions.
- Add a "Default" entry at index 0 so an unset/unknown GfxBackend shows as Default instead of resolving to the first supported backend.

fixes #3466


francisdb:
merged in #3474

PR #3474, opened 2026-06-02
render: clean up graphics backend selection, validation and UI

### LINKED ISSUE #3466: Graphics backend reported Direct3D11 on Linux?

### LINKED ISSUE #3470: GfxBackend setting not validated
```ini
GfxBackend = opengl
```

```
2026-06-02 22:08:39.790 INFO  [2666931] [RenderDevice::RenderDevice@1163] Using graphics backend:  (available: Noop, Direct3D11, Direct3D12, OpenGL, Vulkan, Default)
2026-06-02 22:08:39.857 ERROR [2666989] [RenderDevice::SelectBackBufferFormat@270] Driver did not report any supported backbuffer format for Visual Pinball Player. Defaulting to BGRA8
2026-06-02 22:08:39.885 INFO  [2666989] [RenderDevice::RenderThread@438] BGFX initialized using Vulkan backend (3840x2160 BGRA8)
```

The correct backend is `OpenGL` with the correct casing. Maybe this could log a warning/error about the fallback

francisdb:
Consolidates the graphics-backend list into one helper,
`RenderDevice::GetSelectableBackendNames()`, shared by the `GfxBackend`
validation/log and the in-game graphics dropdown so they can't drift.

- **Warn on an invalid `GfxBackend`** - it's case sensitive (`opengl` != `OpenGL`)
  and unknown values silently fell back to the platform default. The warning lists
  the valid values (`Default` plus the backends bgfx reports as supported,
  excluding Noop, WebGPU, and Direct3D12 in release).
- **Report fallback** - if a requested backend can't initialize, bgfx falls back.
  The selection logs as `Requested graphics backend: ...`, and the render thread
  warns when the actual backend (`getRendererType`) differs.
- **Fix the name table** - it was missing `WebGPU` (added to bgfx after Vulkan),
  mislabeling it as `Default` and leaving the fallback name empty. A `static_assert`
  keeps it in sync with `bgfx::RendererType`.
- **Show `Default` for an unset backend** in the dropdown (was the first backend,
  e.g. `Direct3D11` on Linux) and the log.

fixes #3466
fixes #3470

### DIFF
--- a/src/renderer/RenderDevice.cpp
+++ b/src/renderer/RenderDevice.cpp
@@ -298,6 +298,8 @@ colorFormat RenderDevice::BGFXtoVPXTextureFormat(bgfx::TextureFormat::Enum forma
 }
 
 
+static const string& bgfxRendererName(const bgfx::RendererType::Enum type);
+
 void RenderDevice::RenderThread(RenderDevice* rd, bgfx::Init init)
 {
    SetThreadName("RenderThread"s);
@@ -420,6 +422,14 @@ void RenderDevice::RenderThread(RenderDevice* rd, bgfx::Init init)
    {
       PLOGE << "BGFX initialization failed";
       exit(-1);
+   }
+
+   // A specific backend was requested but BGFX created a different one (init.fallback let it fall back to
+   // the next best because the requested backend failed to initialize), so make that explicit.
+   if (init.type != bgfx::RendererType::Count && bgfx::getRendererType() != init.type)
+   {
+      PLOGW << "Requested graphics backend " << bgfxRendererName(init.type) << " is unavailable; BGFX fell back to "
+            << bgfx::getRendererName(bgfx::getRendererType());
    }
 
    if (init.platformData.nwh)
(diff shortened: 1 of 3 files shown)


vbousquet:
bgfx offers `bgfx::getRendererName`. Perhaps we should use this instead of creating our own ?

Regarding DirectX under Linux, my understanding is that it may be available and that BGFX supports it (at least WSL, DXVK). You are forcing it to be disabled. Is this because there is a bug in BGFX that would wrongly report it as available ? Or is there another reason ? My question is because I would prefer to avoid hacks or patch in VPX that do not belong to it (if it is a BGFX bug, it is better to fix it there, contributing to the main project as we have already done quite a bit)

francisdb:
On getRendererName: it returns versioned display strings ("Direct3D 11", "OpenGL 4.6") and asserts on Count, so it can't be the persisted GfxBackend key - kept the table with a static_assert.

On Direct3D/Linux: not a bug - bgfx compiles it deliberately and getSupportedRenderers only reports compile-time support (no runtime probe), so the D3D11 backend just dlopen("d3d11.dll")s and falls back to Vulkan on native Linux.
So instead of a platform filter I left the list as bgfx reports it and added a warning when the requested backend isn't the one actually used.

PR #3475, opened 2026-06-03
pup: render backglass for packs with a positioned screen 2

francisdb:
Packs that put backglass content on screen 2 *with* a custom position (and
leave screen 6 empty) rendered a black backglass: PUP rejected any
custom-positioned screen as root and fell back to the empty screen 6/1.

- Root selection: only skip a custom-positioned screen when it has no
  children. One that hosts children is a valid canvas, so screen 2 (frame,
  video, callouts) renders.
- Back-tier z-order: draw the popup video before the overlay so a full-window
  frame stays above the background video.

Tested with the https://vpuniverse.com/files/file/23628-big-lebowski-pup-pack/


jsm174:
I tested against 

```
dd
ij4
matrix
koth
b66_orig
cupheadpro
diehard
fs
futurama
gotg
mb
stle
tf
tna
```

No regressions, so good to go.  Thanks.

Issue #2133, opened 2025-02-07
standalone: whirlwind backglass locks table


francisdb:
Vpinball BGFX 10.8.1 branch

The table plays fine when I remove the backglass.

Table: https://www.vpforums.org/index.php?app=downloads&showfile=16337
Backglass: https://vpuniverse.com/files/file/10812-whirlwind-williams-1990-b2s-with-full-dmd/

However when I add this blackglass the table locks up like below.
This backglass has an animated fan that rotates which works, maybe it's consuming the whole gui thread?

![Image](https://github.com/user-attachments/assets/c2ca80d3-c9a5-40c4-9da6-693113011986)

![Image](https://github.com/user-attachments/assets/c0d37015-9de3-40d5-8c03-613398779e3d)

jsm174:
since you can debug on linux, can you pause it, and get the stack where it's locked up?

francisdb:
Found out this only hangs when enabling `SDL_RENDER_DRIVER="vulkan"`

francisdb:
Main thread locks up here

```
futex_wait 0x00007ffff129e230
VP::WindowManager::ProcessUpdates WindowManager.cpp:151
Standalone::ProcessUpdates Standalone.cpp:99
operator() pintable.cpp:2630
operator() player.cpp:1843
Player::MultithreadedGameLoop player.cpp:1886
Player::GameLoop player.cpp:1866
PinTable::Play pintable.cpp:2671
VPinball::DoPlay vpinball.cpp:1059
VPApp::Run main.cpp:1041
WinMain main.cpp:1233
main main.cpp:1281
```

Issue #1004, opened 2023-10-27
Framerate limiter doesn't work on some tables


Crrispy:
**Describe the bug**
The maximum framerate in the graphics setup doesn't work. I've set it 120 and vsync off in the nvidia driver, it still goes over to 140/150fps (as reported by F11 screen) and then framerate fluctuates a lot and the ball stutters.

**To Reproduce**
Set "Maximum framerate" to the max refresh rate of your display and vsync off on the gfx driver.
Start a table which is not limited by your hardware, press F11 and check that the framerate goes over the value set in the settings.

**Expected behavior**
Framerate should be capped to the value entered in settings.

**Versions & Settings**
 - VPX: 10.8 build 1635 , GL x64
 - Windows version: 11
 - GPU: RTX2080 , latest Nvidia driver

[VPinballX.log](https://github.com/vpinball/vpinball/files/13191423/VPinballX.log)

Crrispy:
Edit (I can't edit the text body anymore, only title): the issue does not happen with all tables. To reproduce it, use VPW's Johnny Mnemonic: https://vpuniverse.com/files/file/15487-johnny-mnemonic-williams-1995/
With other tables like Spongebob, the frame limiter works fine.

Crrispy:
Also doesn't work on Totan 4K  https://vpuniverse.com/files/file/12652-tales-of-the-arabian-nights-4k-williams-1996/

francisdb:
Aha, that sounds like the issue where vsync is sometimes forced by tables user settings. Try overriding the value in the table specific ini to see if that works.

JM is known to force vsync overriding the default. I need to set it to 0 (no sync) to have the table run properly.

Crrispy:
> Aha, that sounds like the issue where vsync is sometimes forced by tables user settings. Try overriding the value in the table specific ini to see if that works.
> 
> JM is known to force vsync overriding the default. I need to set it to 0 (no sync) to have the table run properly.
> 
> `Johnny Mnemonic (Williams 1995) VPW v1.1.ini`
> 
> ```ini
> [Player]
> SyncMode = 0
> ```

Hello, well, this is quite the opposite, I have vsync off in the nvidia control panel, but the frame limiter set to 120 in VPX. This works fine for some tables, not for some others. But there it's rather disabling the global VPX frame limiter parameter since it goes well over 120. I will try your suggestion though.

francisdb:
For some more testing you can also add this to the player section

```ini
MaxFramerate = 120
[...]

PR #3477, opened 2026-06-03
 table: drop obsolete pre-10.8 table-embedded user settings

francisdb:
Pre-10.8 tables can embed user settings in the `.vpx` itself, imported on load
when no user ini exists. With 10.8 (settings in the ini everywhere, unified
settings UI) these are obsolete, and worse, they silently override the user's
global configuration - making a single table render or perform unexpectedly with
no clear way to tell why (see #2133, #1004).

This drops the import for the render/perf settings: anti-aliasing (`UAAL`), FXAA
(`UFXA`), synchronization (`AVSY`), detail level (`ARAC`/`OGAC`) and day/night
(`OGDN`). They are still read past for file-format compatibility but no longer
applied. `ARAC`/`OGAC` are the detail-level value and its "override global"
flag, so they go as a pair.

Ball trail (`BTRA`/`BTST`) is intentionally left importing as-is: unlike the
others it has a real ongoing purpose (table visual style and an accessibility
aid). It needs a proper redesign - the table declares whether it uses trails
(respected), plus a user force on/off/strength override like the ball-image
override - rather than a plain drop. Tracked as follow-up.

The first revision of this PR warned about these fields; per review we drop them
instead (the log should not be part of the user experience).

_Out of scope: the legacy `.pov`-file setting remaps apply the same kind of
obsolete settings from an explicit user import and could be cleaned up
separately._

### DIFF
--- a/src/parts/pintable.cpp
+++ b/src/parts/pintable.cpp
@@ -2413,11 +2413,10 @@ void PinTable::Load(IObjectReader& reader)
             break;
          case FID(BPRS): m_ballPlayfieldReflectionStrength = reader.AsFloat(); break;
          case FID(DBIS): m_defaultBulbIntensityScaleOnBall = reader.AsFloat(); break;
-         case FID(UAAL):
-            // Before 10.8, user tweaks were stored in the table file (now moved to a user ini file), we import the legacy settings if there is no user ini file
-            if (const int useAA = reader.AsInt(); useAA != -1 && !hasIni)
-               m_settings.SetPlayer_AAFactor(useAA == 0 ? 1.f : 2.f, true);
-            break;
+         // Obsolete pre-10.8 render settings (anti-aliasing, FXAA, synchronization) that were embedded in
+         // the .vpx and imported when no user ini existed, silently overriding the global configuration
+         // (see #2133, #1004). Read past for file format compatibility but no longer applied.
+         case FID(UAAL): reader.AsInt(); break;
          case FID(UAOC):
             // Before 10.8, this setting could be set to -1, meaning override table definition using video options instead
             m_enableAO = reader.AsInt() != 0;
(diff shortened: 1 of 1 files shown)


vbousquet:
I guess, the best move would be to just drop all of these 'hard coded into VPX file' user settings. I persisted them when moving from 10.7 to 10.8 as all the ini stuff was new and people would need to time to port things. Now that we are moving a quite large step forward, with ini everywhere and unified UI for all settings, these legacy fields are deeply obsolete. Beside I think the log should never be part of the user experience at all, just debugging purpose for (very) advanced users (and some AIs nowadays).

I can do the change but I'm quite low on free time so this may have to wait a bit. If you want to update the PR for removing all these, I would be ok (and happy) with that.

francisdb:
@vbousquet that is indeed the better solution, on it

francisdb:
For debugging table/script issues on standalone the log is very important. It's the first thing we ask the user to provide or look at. That's why we want all required info in there, hence my other logging improvement PRs. These are mostly small papercuts.

The opposite is also true, errors that are logged that are not really errors cause our users to report non-issues or make them think their issue is related to something it is not.

francisdb:
@vbousquet Should any of the remaining pre-10.8 table-embedded fields also be dropped here?
Ball trail (`BTRA`/`BTST`), detail level (`ARAC`/`OGAC`) and the day/night override (`OGDN`) are still imported when no user ini exists. Ball trail in particular may be relied on as a table feature, so I left them all in for now - let me know which (if any) are safe to drop.

vbousquet:
I would say that OGAC/OGDN should be removed too as they are overrides of the table definition (so the table define something then override it). It is wrong with regards to the current setting model.

Balltrail would need a proper cleanup, that is to say table should define if it includes balltrail and this should be respected. A setting allowing the user to force them on (with a strength) or off should be added (to help people with visual defficiency like the ball image override).

francisdb:
Done. Dropped `OGDN` and `OGAC` (plus its coupled `ARAC` detail-level value) alongside AA/FXAA/sync.

Left ball trail importing for now since it needs the table-declares + user-force-override redesign you described.

vbousquet:
Thanks!

PR #3478, opened 2026-06-03
shaders: fix SMAA crash and broken render on Vulkan

### LINKED ISSUE #2133: standalone: whirlwind backglass locks table

francisdb:
On the Vulkan (SPIR-V) build the SMAA shaders went through the HLSL path, which
declares standalone `SamplerState` objects (`LinearSampler`/`PointSampler`).
bgfx does not add those to the pipeline layout, so the SPIR-V references sampler
bindings the layout never declares. On RADV this crashes while compiling the
pipeline (`lower_immediate_samplers`); when it does not crash, the AA samples an
invalid descriptor and renders dark/washed. OpenGL is unaffected because it
already uses the combined-sampler GLSL path.

Route the SPIR-V build through a custom SMAA porting (`SMAA_CUSTOM_SL`) that
samples via bgfx combined samplers, like the working OpenGL build, while keeping
bgfx's HLSL-style types (the plain GLSL path would redefine `bvec2`/`float2` and
clash). Validation (`VUID-VkGraphicsPipelineCreateInfo-layout-07988`,
`VUID-vkCmdDraw-None-08114`) is clean afterwards and the SPIR-V no longer
references the standalone samplers.

Only the SPIR-V path changes; Direct3D and Metal are untouched.

⚠️ _Direct3D 11/12 hit the same standalone-sampler problem - `Shader.cpp` already
notes it ("Direct3D 11 & 12 have issues with SMAA and DMD display shaders") and
carries a debug-build workaround that skips the `Point`/`Linear` SMAA samplers
during uniform validation (`Shader.cpp:1369`). This PR does not change the HLSL
path, so D3D is left as a known follow-up; the same `SMAA_CUSTOM_SL`
combined-sampler approach should fix it there too (after which that skip can go)._

Only the three SMAA `spv` blobs in `bgfx_antialiasing.h` are regenerated (the
source change affects only the SPIR-V path). They were verified byte-identical
to a full regeneration from the `vpinball-bgfx-shaders` CI workflow, so the
header is canonical for this change. The other targets' blobs are intentionally
left untouched: regenerating them would only introduce non-deterministic
DXBC-hash churn unrelated to this fix.

fixes #2133

Table overrides SMAA, see #3477 

Before below or black screen (F12 works and overlay visible)

image

After

image

### DIFF
--- a/src/shaders/bgfx/SMAA.sh
+++ b/src/shaders/bgfx/SMAA.sh
@@ -11,6 +11,22 @@ uniform vec4 w_h_height;
 
 #if BGFX_SHADER_LANGUAGE_GLSL
 #define SMAA_GLSL_3
+#elif BGFX_SHADER_LANGUAGE_SPIRV
+// Vulkan: the HLSL path declares standalone SamplerState (Linear/PointSampler) that bgfx does not
+// add to the pipeline layout, so RADV crashes compiling the pipeline and the AA renders broken.
+// Use a custom porting that samples through bgfx combined samplers (like the working OpenGL build),
+// while keeping bgfx's HLSL-style types (the GLSL path would redefine bvec2/float2 and clash).
+#define SMAA_CUSTOM_SL
+#define SMAATexture2D(tex) sampler2D tex
+#define SMAATexturePass2D(tex) tex
+#define SMAASampleLevelZero(tex, coord) texture2DLod(tex, coord, 0.0)
+#define SMAASampleLevelZeroPoint(tex, coord) texture2DLod(tex, coord, 0.0)
+#define SMAASampleLevelZeroOffset(tex, coord, offset) texture2DLodOffset(tex, coord, 0.0, offset)
+#define SMAASample(tex, coord) texture2D(tex, coord)
+#define SMAASamplePoint(tex, coord) texture2D(tex, coord)
+#define SMAASampleOffset(tex, coord, offset) texture2DLodOffset(tex, coord, 0.0, offset)
+#define SMAA_FLATTEN
+#define SMAA_BRANCH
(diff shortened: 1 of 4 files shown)


jsm174:
> A full shaders.sh regeneration on a
complete toolchain should replace the header before merge.

fwiw, you can manually run the [vpinball-shaders](https://github.com/vpinball/vpinball/blob/master/.github/workflows/vpinball-shaders.yml) workflow. This will artifact a zip file of all the shaders.

francisdb:
> > A full shaders.sh regeneration on a
> > complete toolchain should replace the header before merge.
> 
> fwiw, you can manually run the [vpinball-shaders](https://github.com/vpinball/vpinball/blob/master/.github/workflows/vpinball-shaders.yml) workflow. This will artifact a zip file of all the shaders.

Thanks, will update the pr

francisdb:
> > A full shaders.sh regeneration on a
> > complete toolchain should replace the header before merge.
> 
> fwiw, you can manually run the [vpinball-shaders](https://github.com/vpinball/vpinball/blob/master/.github/workflows/vpinball-shaders.yml) workflow. This will artifact a zip file of all the shaders.

Updated PR description on this topic. Short: no extra shader compilation needed.

Issue #1277, opened 2023-12-15
Haunted Hotel very dark


francisdb:
**Describe the bug**

This is what the table looks like when it's just started. I don't remember it being so dark in previous vpinball versions.

![image](https://github.com/vpinball/vpinball/assets/161305/c1172f9e-9cb2-4c1b-a9a3-7f2902e55dce)

vpinball 10.8.1 branch standalone

https://www.vpforums.org/index.php?app=downloads&showfile=16218

wiesshund:
@francisdb Hi
Did you try cycling through the LUT levels?

Yes, that is way darker than i made the table.
I made it dark and creepy, but not blind LoL

I wonder if the LUT levels will help at all in the mean time?

francisdb:
This is max brightness by switching LUT's

![image](https://github.com/vpinball/vpinball/assets/161305/da1fb552-c28c-417e-8b15-ae5c9678e7c6)

francisdb:
I extracted the table and compared its global lighting against the bundled example table to see what actually makes it so dark.

| setting | Haunted Hotel | example table |
|---|---|---|
| `light_emission_scale` | 80,000 | 4,000,000 |
| `light_height` | 9,340 | 5,000 |
| `env_image` | `Environment Moody` (dark) | `environment` (bright) |
| `env_emission_scale` | 0.3 | 2.0 |
| `light_ambient` | #151515 | #000000 |
| `global_emission_scale` | 0.52 | 0.52 |
| `tone_mapper` | none | none |

The darkness comes from two stacked factors, both about an order of magnitude. The direct lighting is tiny: `light_emission_scale` is 50x lower than the example and the scene light sits almost twice as high above the playfield. The fill lighting is tiny too: the environment image is used for diffuse scene lighting (it is prefiltered into the irradiance map that lights every material and the ball, not just reflections), and here it is a dark custom map at `env_emission_scale` 0.3 versus a bright map at 2.0.

The things people usually reach for are not the cause: the tonemapper is unset in both (changing it makes no visible difference), `global_emission_scale` is identical, and the ambient is actually higher on Haunted Hotel, not lower. Raising camera exposure brightens it but only flatly, because it is compensating for genuinely low input brightness rather than a wrong curve.

[...]

PR #3479, opened 2026-06-03
pinmame: fix hang on table exit for time-fence tables

francisdb:
Exiting a PinMAME table could hang the whole app: `Controller::Stop()` ->
`PinmameStop()` blocks forever joining the emulation thread.

`Stop()` disabled the time fence first (`PinmameSetTimeFence(0.0)`). While a
fence is active the PinMAME OSD throttle bails out, so its frame timing sits
stale; disabling it makes the throttle busy-spin inside `updatescreen()`, which
never returns, so the emulation thread never sees the stop flag.

Fix: leave the fence alone - the throttle keeps bailing and the emulation stays
fence-paced, so it observes the stop and exits cleanly. libpinmame resets the
fence itself on stop.

Notes:
- The `PinmameSetTimeFence(0.0)` line dates back to the plugin's first commit
  (a16e5d224) and was never a deliberate fix.
- Not table-specific: `controller.vbs`/`core.vbs` enable the time fence on the
  standalone plugin, so any PinMAME table is exposed; some slip through, others
  (Haunted Hotel) hang every time.
- The Windows/COM path has the same bug upstream
  (vbousquet/pinmame `src/win32com/Controller.cpp` `put_TimeFence(0.0)` in
  `CController::Stop()`, commit `169239a4`) and needs the same fix there?

Reproduced with 

* Haunted Hotel (LTD) (a generally messy table, also #1277):
   https://www.vpforums.org/index.php?app=downloads&showfile=16218
* Alien warrior (same author)
   https://www.vpforums.org/index.php?app=downloads&showfile=15941

### Where it hangs

```
  Main thread - blocked joining the PinMAME emulation thread:
  Player::~Player()                          (player.cpp:839, "Closing player...")
    PinTable::FireDispID()                   -> script runs Controller.Stop on exit
      B2SLegacy::PinMAMEAPI::HandleCall()    
        PinMAME::Controller::Stop()          (Controller.cpp:127)
          PinmameStop()
            std::thread::join()              <- parked here forever

  The emulation thread it's waiting for (Thread 116) - stuck spinning, never returns to check the stop flag:
  StartGame -> run_game -> cpu_run -> ... -> updatescreen()
    update_video_and_audio()
      osd_update_video_and_audio()
        usleep()                             <- busy-spinning in the OSD throttle
```

### DIFF
--- a/plugins/pinmame/Controller.cpp
+++ b/plugins/pinmame/Controller.cpp
@@ -126,7 +126,11 @@ void Controller::Run(long hParentWnd, int nMinVersion)
 
 void Controller::Stop()
 {
-   PinmameSetTimeFence(0.0);
+   // Do not disable the time fence here (the previous PinmameSetTimeFence(0.0)): while a fence is
+   // active the PinMAME OSD throttle bails out, but disabling it makes the throttle (or, unthrottled,
+   // the audio update) run away inside updatescreen(), which then never returns - so the emulation
+   // thread never re-checks the stop flag and PinmameStop() hangs joining it. Leaving the fence at its
+   // last value keeps the throttle bailing and the emulation fence-paced, so it observes the stop.
    if (PinmameIsRunning())
    {
       PinmameStop();


francisdb:
@vbousquet remember why you added that line?

vbousquet:
The fence disabling was added to avoid dead locking :) So the exact reason you are removing it.

I would need to remember and search more deeply but the problem was that the emulation thread could lock waiting on the TimeFence and never join. Note that disabling TimeFence should never block the emulation thread, so if you are finding an issue here, I guess we have a deeper problem than the workaround of not disabling the TimeFence. I will try to have a look when I get some time off

francisdb:
Found another table by the same author with the same issue:
https://www.vpforums.org/index.php?app=downloads&showfile=15941 (Alien Warrior)

Applying the PR fixed it, tested with 12 runs.

vbousquet:
I just had a look and as expected, the bug is not on VPX side, the problem is how LibPinMAME handles the synchronization in `throttle_speed_part`. Setting the timefence would just be a hack to bypass the bug in it. I will close as this PR does not address the core issue

PR #3480, opened 2026-06-04
cli: clear error for an invalid table argument

francisdb:
Running `-play <file>` with a file that exists but is not a Visual Pinball table
- e.g. accidentally pointing at the table's `.ini` sidecar instead of the `.vpx`
- loaded into a black screen with only a dialog written to the log, no useful
feedback. It now reports a clear command line error:

```
Command Line Error

'/path/Table.ini' is not a Visual Pinball table (.vpx)
```

A VPX table is an OLE compound file (CFB), so the file is validated by its
signature. This also catches a file that has a `.vpx` extension but is not
actually a table, which an extension-only check would miss.

The same validation applies to every argument that names a table - a bare table
argument, `-play`, `-povedit`, `-audit`, `-pov`, `-extractvbs`, `-captureattract`
and `-tournament`. A table argument now **plays or fails identically on all
platforms**: a missing, directory or non-table argument is an error, not a silent
no-op nor a fall-back to the editor.

A bare launch with no arguments now shows the usage instead of starting up and
immediately exiting. (On Windows a no-command launch still opens the editor.)

### DIFF
--- a/src/core/AppCommands.cpp
+++ b/src/core/AppCommands.cpp
@@ -4,6 +4,7 @@
 #include "AppCommands.h"
 
 #include <iostream>
+#include <fstream>
 
 #include "extern.h"
 #include "core/TournamentFile.h"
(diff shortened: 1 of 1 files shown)


francisdb:
If interest I can after this merges create a pr that shows a SDL3 File open dialog for a bare launch like the macos wrapper does?

vbousquet:
Looks nice, thanks!

I don't think we should fallback to the editor on any platform. When asking to play a table, the application should play or fail on all platform with the exact same behavior. Could you update the PR to avoid any `#ifdef` and have a unified behavior across platforms ?

The only special situation where a fallback seems expected is when starting to edit a table `-edit file` where we could just start the editor without any table (it's a windows only command but that's not the point here).

PR #3484, opened 2026-06-04
b2slegacy: render the embedded DMD with the ScoreView glass overlay

francisdb:
The legacy B2S embedded DMD (FullDMD / backglass overlay) drew the dots without the glass cover the ScoreView DMD applies, so it looked flat and sharp while the ScoreView DMD rendered glowy. Load the same glass texture and use the same parameters (roughness, area, ambient) so the embedded DMD matches the ScoreView.

The glass image is read from the ScoreView plugin's layouts folder.

This unifies both, literally embedding the score view.

Before

image

After

image

_In the future we would probably want to make this tweakable from the ui, disabling, changing to non-dirty glass, style selection, save per table._

### DIFF
--- a/make/CMakeLists_plugin_B2S.txt
+++ b/make/CMakeLists_plugin_B2S.txt
@@ -17,6 +17,7 @@ set(B2S_PLUGIN_SOURCES
    plugins/b2s/B2SRenderer.cpp
    plugins/b2s/B2SServer.h
    plugins/b2s/B2SServer.cpp
+   plugins/plugins/DMDDisplay.h
    plugins/plugins/ResURIResolver.h
    plugins/plugins/ResURIResolver.cpp
    third-party/include/tinyxml2/tinyxml2.h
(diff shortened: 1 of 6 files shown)


jsm174:
Wouldn't you want to respect whatever scv file got loaded for the scoreview? That controls colors, layout, glass, etc.

francisdb:
> Wouldn't you want to respect whatever scv file got loaded for the scoreview? That controls colors, layout, glass, etc.

it's still on draft :-)

francisdb:
Closing this. Diagnosis is clear but the clean fix is bigger than it looks.

The embedded B2S DMD and the ScoreView DMD use the same shader; the ScoreView just composites a glass cover (Glass-Dirty-1.webp + roughness/reflection/depth) and the legacy one draws bare dots - so one looks like a real DMD behind glass and the other like a flat render.

Quick fix (b2slegacy applies the same glass) works but reaches into the ScoreView plugin's asset and hardcodes params. The proper fix is rendering the legacy DMD through the ScoreView's .scv engine, but that restructures the ScoreView plugin and hit a few landmines, so it's better driven by whoever owns that code:
  - the engine + its helpers must move to shared plugins/plugins/ and compile into both;
  - LOGE/LOGD are per-plugin LPI macros (shared TU has to resolve logging per plugin);
  - GetPluginPath hardcodes a plugin-specific Windows symbol;
  - the engine positions from the layout, but the legacy DMD positions via black-square detection, so it needs an explicit-rect entry point.
  
Positioning stays as-is regardless - this is only about how the DMD is filled. Leaving it as a writeup if someone wants to pick up the engine-sharing.

PR #3491, opened 2026-06-05
ci: support PRs (@francisdb), combine workflows, switch to sha short

jsm174:
This PR will supersede https://github.com/vpinball/vpinball/pull/3458.

Along time ago we split the CI into desktop, mobile, and sbc builds. The reason for this was sbc builds would take 4+ hours and we didn't want to hold up desktop builds. But with the linux-arm64 runners, this is no longer an issue. In fact, iOS, android, and all SBC builds are much faster than any desktop builds. 

This takes @francisdb suggested approach of building a subset of artifacts (BGFX x64/arm64/release, 1 GL, 1 DX) using the existing caches. (Also no code signing / notarization, etc)


francisdb:
Thanks Jason! Very happy to see this implemented.

PR #3497, opened 2026-06-10
pup: keep self-positioned screens as window root (fixes #3487)

### LINKED ISSUE #3487: 10.8.1 BGFX/Win64 - PUP Plugin - Full DMD does not render on 3-screen with custom position defined
**Describe the bug**
In a multi-screen cabinet, some PUP packs when running their setup bats define custom positions for the FullDMD; known examples are Futurama and Die Hard.  If a custom position is defined, the FullDMD appears to not render anywhere (not z-fighting; scoreview monitor will still show empty Windows desktop, also does not appear to load associated audio).  If the custom position is cleared, it appears to render and behave normally.

**To Reproduce**
With Futurama, use Option 1 - Backglass DMD Separate.  The relevant line in screens.pup that is created is:
5,FullDMD,,,0,ForceBack,100,"5,0,0,100,100"
With this, the DMD will not render.  If the line is changed to:
5,FullDMD,,,0,ForceBack,100,
...then the table appears to function as expected.

With Die Hard, use Option 1 - Backglass DMD Separate.  The relevant line in screens.pup that is created is the same:
5,FullDMD,,,0,ForceBack,100,"5,0,0,100,100"
With this, the DMD will not render.  Again, if the line is changed to:
5,FullDMD,,,0,ForceBack,100,
...then the table appears to function as expected.

This appears to possibly be a related issue to what is described/addressed in #3475 
**Versions & Settings**
-VPX 10.8.1 BGFX Win64 5121
-Win 11

[vpinball.log](https://github.com/user-attachments/files/28619969/vpinball.log)

francisdb:
A screen whose CustomPos references its own screen number (e.g. 5,"5,0,0,100,100") has no parent, so treat it as a valid root canvas instead of skipping it; skip a root candidate only when it is positioned on another screen and hosts no children.

fixes #3487

### DIFF
--- a/plugins/pup/PUPManager.cpp
+++ b/plugins/pup/PUPManager.cpp
@@ -631,11 +631,12 @@ int PUPManager::Render(VPXRenderContext2D* const renderCtx, void* context)
       padBottom = pupTopperPadBottom_Get();
       break;
    case VPXWindowId::VPXWINDOW_Backglass:
-      // Prefer screen 2, falling back to 6. A screen with a custom position but no children is a
-      // positioned element rather than a backglass canvas, so skip it; a custom positioned screen
-      // that does host children (the overlays/videos) is a valid canvas and kept as root.
+      // Prefer screen 2, falling back to 6. A childless screen positioned on another screen is a
+      // positioned element rather than a backglass canvas, so skip it; a positioned screen that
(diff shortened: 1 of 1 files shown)


francisdb:
Somebody should write a proper manual on all this 🙄

It's silly. Screen says position me on myself at full width and height.
Worth a warning that this does nothing at all.

PR #3500, opened 2026-06-11
pinmame plugin: return VPinMAME defaults for settings

francisdb:
Table scripts read `Controller.Games(x).Settings.Value(...)` expecting the defaults VPinMAME ships on Windows (dmd_red=255, dmd_perc33=33, ...), but the plugin returned 0 for every key. This made FlexDMD-based tables that derive colors from these settings render a black DMD, and historically caused script crashes (vpx-standalone-scripts https://github.com/jsm174/vpx-standalone-scripts/pull/309 / https://github.com/jsm174/vpx-standalone-scripts/pull/445).

- Add a defaults table matching PinMAME src/windows/config.c and src/win32com/VPinMAMEConfig.cpp, used by Settings and GameSettings
- Store written values per instance so write-then-read round-trips
- Cache Game instances per name in the Controller so settings survive repeated Games(name) accesses
- Expose Controller.Settings (present in VPinMAME on Windows, used by tables like [Phantom of the Opera](https://github.com/jsm174/vpx-standalone-scripts/pull/137)) and proxy it in both B2S plugins

Verified on linux BGFX with a test table script: defaults returned, writes round-trip at both controller and game level.

### DIFF
--- a/plugins/pinmame/SettingsDefaults.h
+++ b/plugins/pinmame/SettingsDefaults.h
@@ -4,18 +4,9 @@
 
 #include "common.h"
 
-#include <algorithm>
-#include <cctype>
 #include <unordered_map>
 
 namespace PinMAME {
-
-inline string ToLowerKey(const string& key)
-{
-   string k(key);
-   std::transform(k.begin(), k.end(), k.begin(), [](unsigned char c) { return static_cast<char>(std::tolower(c)); });
-   return k;
-}
 
 // Default values matching VPinMAME settings that were never explicitly set,
 // so table scripts relying on VPinMAME defaults behave the same as on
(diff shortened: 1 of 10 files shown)


plugins/pinmame/SettingsDefaults.h
```diff
@@ -0,0 +1,79 @@
...
+
+#include "common.h"
+
+#include <algorithm>
+#include <cctype>
+#include <unordered_map>
+
+namespace PinMAME {
+
+inline string ToLowerKey(const string& key)
```
jsm174: FWIW, I know in a lot of the plugins, alot of these helper type string functions are in common.h

francisdb: Moved to common.h as string_to_lower (matching the b2s plugin naming) in f4b13dce4. Also pushed a fix for a Controller refcount leak the game-object cache introduced (21805fe8e).

Issue #1330, opened 2024-01-11
Standalone: Default right magna save button not set on mac?


francisdb:
**Describe the bug**
By default this is what you get on a mac when pressing F2 (on a table that supports printing keys)

```
Visual Pinball keys:
Left Shift      Left Flipper
Right Shift     Right Flipper
Left Ctrl       Left Magna Save
?    Right Magna Save
Enter   Launch Ball
1       Start Button
5       Insert Coin 1
4       Insert Coin 2
Q       Exit Game
T       Mechanical Tilt
Z       Nudge from Left
/       Nudge from Right
Space   Nudge forward
```

_I tried printing the key code and it seems to be 157? No idea what that corresponds to._

jsm174:
Ignoring the mac keyboard from above for a moment,

`Right Ctrl` is the Right Magna Save.

Lets walk through this:

In `pininput.h`, we have an `SDLK_TO_DIK` map: (SDL to DirectInput)
   
```
   { SDLK_LSHIFT, DIK_LSHIFT },
   { SDLK_RSHIFT, DIK_RSHIFT },
   { SDLK_LCTRL, DIK_LCONTROL },
   { SDLK_RCTRL, DIK_RCONTROL },
```

In `dinput.h` we have

```
#define DIK_RCONTROL        0x9D
```

`0x9D` hex is `157` decimal

in `core.vbs`, it looks like all the text names are defined in two arrays, and 157 is never handled.

```
' Convert keycode to readable string
Private keyNames1, keyNames2
keyNames1 = Array("Escape","1","2","3","4","5","6","7","8","9","0","Minus '-'",_
"Equals '='","Backspace","Tab","Q","W","E","R","T","Y","U","I","O","P","[","]",_
"Enter","Left Ctrl","A","S","D","F","G","H","J","K","L",";","'","`","Left Shift",_
"\","Z","X","C","V","B","N","M",",",".","/","Right Shift","*","Left Menu","Space",_
"Caps Lock","F1","F2","F3","F4","F5","F6","F7","F8","F9","F10","NumLock","ScrlLock",_
"Numpad 7","Numpad 8","Numpad 9","Numpad -","Numpad 4","Numpad 5","Numpad 6",_
"Numpad +","Numpad 1","Numpad 2","Numpad 3","Numpad 0","Numpad .","?","?","?",_
"F11","F12","F13","F14","F15")
[...]

PR #3508, opened 2026-06-12
standalone: skip the legacy key help text on standalone

francisdb:
see #1330

### DIFF
--- a/scripts/core.vbs
+++ b/scripts/core.vbs
@@ -2878,6 +2878,13 @@ End Function
 
 Private vpmSystemHelp
 Private Sub vpmShowHelp
+	' With the PinMAME plugin most listed keys do not apply (the VPM dialogs are gone and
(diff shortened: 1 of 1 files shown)


jsm174:
I think we've held off not having `isStandalone` as table authors would abuse this. I'll defer to @vbousquet.

```
Sub VPMVol
	' PinMAME plugin streams its sound through VPX, using VPX mixing and therefore volume settings
	if IsPluginPinMAME Then Exit Sub
	If IsStandalone Then
		MsgBox "VPinMAME Volume adjustment is not supported on standalone versions of Visual Pinball."
		Exit Sub
	End If
```

I'm wondering if standalone would ever not have the `IsPluginPinMAME`. ie, I would think it would never get there.

francisdb:
@jsm174 missed that part, removed `IsStandalone` and switched to `IsPluginPinMAME`

PR #3512, opened 2026-06-12
ui: transparent unused embedded displays, frame the adjusted one

francisdb:
Embedded ancillary windows were cleared to black every frame even when no renderer claimed them, leaving black squares (e.g. a table without b2s fulldmd/dmd). Skip the clear when nothing rendered on the last frame so the playfield stays visible. 

While a display settings page is open all regions are shown so they can be aligned, and the display being adjusted gets a green frame drawn over its content.

Here I have all windows setup embedded but the b2s only renders on one

Screenshot_2026-06-12_18-33-11

Here I am in the setup where all screens are shown and the one being configured highlighted (green border)

Screenshot_2026-06-12_18-33-01

### DIFF
--- a/src/renderer/Renderer.cpp
+++ b/src/renderer/Renderer.cpp
@@ -3337,17 +3337,12 @@ RenderTarget* Renderer::SetupAncillaryRenderTarget(
    return rd->GetCurrentRenderTarget();
 }
 
-void Renderer::ClearEmbeddedAncillaryWindow(VPXWindowId window, const VPX::RenderOutput& output, RenderTarget* embedRT)
+// Pages of the in game UI adjusting the displays, indexed by VPXWindowId
+static const string s_displaySettingsPages[]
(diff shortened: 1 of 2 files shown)


gitfool:
Another option I'd love to have for embedded ancillary windows is to be able to configure their opacity/transparency with global settings and per table overrides. The scenario I have in mind is for single screen, in particular mobile, devices that are screen space constrained, where you want to maximise playfield size to fill the screen but then still be able to enjoy an embedded backglass and scoreview overlay, without completely obscuring the playfield.

francisdb:
> Another option I'd love to have for embedded ancillary windows is to be able to configure their opacity/transparency with global settings and per table overrides. The scenario I have in mind is for single screen, in particular mobile, devices that are screen space constrained, where you want to maximise playfield size to fill the screen but then still be able to enjoy an embedded backglass and scoreview overlay, without completely obscuring the playfield.

I think that would belong in it's own issue as it's a whole new setting.

gitfool:
Sure. I just brought it up here since you were making useful improvements in this area. Created https://github.com/vpinball/vpinball/issues/3513.

Issue #2772, opened 2025-10-15
Can't move scoreview window


francisdb:
git master
linux
wayland

```ini
[Plugin.PUP]
Enable = 1

[ScoreView]
ScoreViewOutput = 2
```

When running vpinball in true Wayland mode (not XWayland), which is the default on newer Linux distributions, the score view window can't be placed using config file as Wayland does not allow placing windows. The scoreview ends up behind the playfield.

Therefore we wound need a way to move the window eg by dragging the mouse. This used to work by clicking in the window and dragging.

_On Gnome Alt-F7 can be used to move the window but my guess is that the average user won't know this. (I had to look it up myself)._

This is what you currently get:

<img width="2199" height="1375" alt="Image" src="https://github.com/user-attachments/assets/2492806c-e510-4f61-b367-c0de0b435f8b" />

vbousquet:
This part of the UI is being redesigned at the moment. The idea was to have a display section that would allow to enable/disable/embedd the ancillary (and main) windows, but also to set them up.

My question would be what is the correct way to open à window at à prefix de position on Linux ? Is there some docs somewhere (ideally using SDL) ? Did the initial quick hack for dragging window worked ? If so we could leverage the way it used to place window

francisdb:
> My question would be what is the correct way to open à window at à prefix de position on Linux ?

As far as I know there is no such thing in Wayland unless you force the use of X11 api which then is handled by XWayland which can place windows 🙄. It's absurd that they don't want to implement absolute window placement.
https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/264

Anyway, my issue is that I just want to have an easy way to move windows around. Maybe enabling window decorations would be enough for desktop use.

>  Did the initial quick hack for dragging window worked?

Which hack?

https://wiki.libsdl.org/SDL3/SDL_SetWindowHitTest with `SDL_HITTEST_DRAGGABLE` looks like a candidate to implement this.

jsm174:
fwiw we did use SDL_HITTEST_DRAGGABLE in the standalone branch (10.8.0) to move all the external windows.

francisdb:
This is still the case, position setting in F12 menu does nothing and always resets to 0. <-- I think this is expected in wayland
Window is not resizable at the corners
Resizing from within the `F12` menu does not resize and makes the whole window unresponsive (< 1fps)

francisdb:
And 2 more related PR's on BGFX

[...]

PR #3514, opened 2026-06-12
wayland: let floating ancillary windows be moved by dragging

francisdb:
The borderless windows could not be moved at all on Wayland, where applications may not position windows themselves. Declare them draggable so the window manager moves them when dragged anywhere, as the 10.8.0 standalone windows did.

Partial fix for #2772

This lets us at least test multi-window on wayland

If needed we could lock this to wayland only

### DIFF
--- a/src/renderer/Window.cpp
+++ b/src/renderer/Window.cpp
@@ -258,6 +258,15 @@ Window::Window(const string& title, const Settings& settings, VPXWindowId window
       SDL_DestroyProperties(props);
    }
 
+   // On Wayland applications may not position windows themselves, so the drag support of
+   // the display settings UI (see Player::ProcessOSMessages) cannot work; declaring the
+   // undecorated floating ancillary windows draggable lets the window manager move them
+   // when dragged anywhere instead. Other platforms keep the settings UI drag (a full
+   // surface hit test would swallow the mouse events it relies on, and make the windows
+   // movable outside of the settings UI, e.g. by stray touches on a cabinet).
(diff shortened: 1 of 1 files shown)


jsm174 (commented):
fwiw, floating ancillary windows are draggable on MacOS, and presumably windows, so this might need to be gated for wayland/linux?

francisdb:
> fwiw, floating ancillary windows are draggable on MacOS, and presumably windows, so this might need to be gated for wayland/linux?

The question is then would adding this break anything on windows / macos

jsm174:
Since it already works, I would not blindly add this for the other platforms. Would need to review SDL's internal implementation since wouldn't want the two fighting against each other if it's different.

francisdb:
@jsm174 true, gated this by `DL_GetCurrentVideoDriver() == "wayland"sv` since windows not being able to position themselves is a wayland problem.

vbousquet:
If we can, this would be nice to have a single path and UX interaction on all platforms. I'm fully ok with moving to SDL as this is one of the main reasons we chose SDL (cross platform, moving all the per platform out of VPX codebase).

francisdb:
Agree: no wl_*/X11 in VPX, use SDL only (so drop the vendored activation); but SDL_SetWindowPosition is a no-op on Wayland by design, so a small SDL-only `if(wayland)` strategy branch is irreducible — SDL can't abstract a capability the protocol doesn't have. Same with the fullscreen flag requirement and wayland that can only place on monitor for fullscreen windows (`set_fullscreen(output)`).

PR #3515, opened 2026-06-12
ui: apply window resizes asynchronously

francisdb:
Resizing a floating ancillary window from the in-game settings locked the game
below 1 fps and could crash.

`SetSize` assumed resizes are synchronous, but the
[SDL_SetWindowSize](https://wiki.libsdl.org/SDL3/SDL_SetWindowSize) docs state:

> On some windowing systems, this request is asynchronous and the new window size
> may not have been applied immediately upon the return of this function. [...]
> Additionally, as this is just a request, it can be denied by the windowing
> system (e.g. tiled window managers).
>
> When the window size changes, an SDL_EVENT_WINDOW_RESIZED event will be emitted
> with the new window dimensions.

On Wayland this is protocol level: sizes arrive as
[configure events](https://wayland.app/protocols/xdg-shell#xdg_surface:event:configure) the client acknowledges. So the immediate read-back left the cached size stale,
and the settings UI kept re-firing the request, each one recreating the swapchain.

The resize flow now works like the underlying protocol: send one request, wait for
the resize event to acknowledge it before sending the next (with a timeout for
window managers that never answer). The resize event does the bookkeeping and a
single swapchain recreate, which also covers resizes initiated by the window
manager itself and display scale changes - previously not handled at all.
(`SDL_SyncWindow` would be the blocking alternative, but blocking the render loop
per request is the wrong trade and it still cannot help when the request is
denied.)

While the swapchain is being recreated the window is skipped for a frame instead
of crashing on its freed backbuffer; that race predates this change, fully
sequencing the recreate against the frame loop is a possible follow-up.

Tested on Wayland with a floating score view: rapid slider resizing went from a
sub-1 fps lockup to tracking the slider.

Related to #2772

### DIFF
--- a/src/core/player.cpp
+++ b/src/core/player.cpp
@@ -1494,6 +1494,19 @@ void Player::ProcessOSMessages(const bool isInitialized)
          SetCloseState(Player::CloseState::CS_STOP_PLAY);
          break;
 
+      case SDL_EVENT_WINDOW_RESIZED:
+      case SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED:
+      {
+         SDL_Window* const wnd = SDL_GetWindowFromID(e.window.windowID);
+         if (m_playfieldWnd->GetCore() == wnd)
+            m_playfieldWnd->OnResized();
+         else
+            for (VPX::RenderOutput* const output : { &m_backglassOutput, &m_scoreViewOutput, &m_topperOutput })
+               if (output->GetMode() == VPX::RenderOutput::OM_WINDOW && output->GetWindow()->GetCore() == wnd)
+                  output->GetWindow()->OnResized();
+         break;
+      }
+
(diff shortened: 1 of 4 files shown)


toxieainc:
Should there be some additional locking/mutex/similar in place?

francisdb:
No extra locking needed. The resize state and `OnResized` all run on the main/logic thread (`SetSize` comes from the in-game UI during `PrepareFrame`, `OnResized` from `ProcessOSMessages`), so they can't race. The only thing the RenderThread touches is the backbuffer, and that's already serialized by the existing frame handoff (`m_frameMutex`/`m_framePending`) via the `AddEndOfFrameCmd` deferral - a separate mutex would just duplicate that. The brief swapchain-recreate gap is covered by the skip-a-frame null check, which is the bit I noted could be tightened later.

PR #3522, opened 2026-06-14
standalone: hide the in-player mouse cursor via SDL

### LINKED ISSUE #2698: VPX Standalone Linux version EnableMouseInPlayer = 0 not disable mouse pointer
Hello.
Thanks for reading my topic.
VPX Standalone Linux version GL 10.8.0 and Ubuntu 24.04.1
In VPinballX.ini selecting
EnableMouseInPlayer = 0
That settings not are disabling the mouse pointer in screen when loading an table.

Have an nice day.

francisdb:
ShowCursor is a no-op stub on standalone, so the player never actually hid the cursor. Drive it through `SDL_HideCursor`/`SDL_ShowCursor` instead.

fixes #2698

### DIFF
--- a/src/core/player.cpp
+++ b/src/core/player.cpp
@@ -826,6 +826,24 @@ Player::Player(PinTable *const table, const PlayMode playMode)
    }
 }
 
+// Show or hide the mouse cursor. The native build uses the Win32 ShowCursor
+// counter; the standalone build uses SDL on every platform, since there
+// ShowCursor is only a no-op stub.
+static void SetMouseCursorVisible(const bool visible)
+{
+#ifndef __STANDALONE__
+   if (visible)
+      while (ShowCursor(TRUE) < 0); // FIXME on a system without a mouse, it looks like this may result in an infinite loop
+   else
+      while (ShowCursor(FALSE) >= 0);
+#else
+   if (visible)
+      SDL_ShowCursor();
+   else
+      SDL_HideCursor();
+#endif
+}
+
 Player::~Player()
 {
    assert(g_pplayer == this && g_pplayer->m_closing != CS_CLOSED);
@@ -1075,8 +1093,7 @@ Player::~Player()
 
    m_ptable->Release();
 
-   while (ShowCursor(FALSE) >= 0);
-   while (ShowCursor(TRUE) < 0);
+   SetMouseCursorVisible(true);
    PLOGI << "Player closed.";
 
 #ifdef __LIBVPINBALL__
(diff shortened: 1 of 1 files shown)


src/core/player.cpp
```diff
@@ -826,6 +826,24 @@ Player::Player(PinTable *const table, const PlayMode playMode)
    }
 }
 
+// Show or hide the mouse cursor. The native build uses the Win32 ShowCursor
+// counter; the standalone build uses SDL on every platform, since there
+// ShowCursor is only a no-op stub.
+static void SetMouseCursorVisible(const bool visible)
```
vbousquet, line 832: The cleanup should be to use our cross platform layer (SDL) for all platform and avoid this type of function and all those horrible #ifdefs

src/core/player.cpp
```diff
@@ -1075,8 +1093,7 @@ Player::~Player()
 
    m_ptable->Release();
 
-   while (ShowCursor(FALSE) >= 0);
-   while (ShowCursor(TRUE) < 0);
+   SetMouseCursorVisible(true);
```
vbousquet, line 1096: Just `SDL_ShowCursor();` would be better here

src/core/player.cpp
```diff
@@ -1212,14 +1229,7 @@ void Player::ApplyPlayingState(const bool play)
...
 {
-   if (m_drawCursor || !IsPlaying())
-   {
-      while (ShowCursor(TRUE) < 0); // FIXME on a system without a mouse, it looks like this may result in an infinite loop
-   }
-   else
-   {
-      while (ShowCursor(FALSE) >= 0);
-   }
+   SetMouseCursorVisible(m_drawCursor || !IsPlaying());
```
vbousquet, line 1232: Avoiding the added function seems cleaner to me. Something like taht:
```
if (m_drawCursor || !IsPlaying())
   {
      SDL_ShowCursor();
   }
   else
   {
      SDL_HideCursor();
   }
```

vbousquet (commented):
A few notes, as these are occasions to improve all these pending things (and avoid adding more bloat code).

src/core/player.cpp
francisdb: Probably the windows code can be dropped?

francisdb: Want me to create a new pr that fixes this or are you on it?

vbousquet: > Probably the windows code can be dropped?

Yes, that's exactly the idea. This was on the pending list. The aim is to fully use SDL as our cross platform layer

PR #3523, opened 2026-06-14
player: drive the in-player cursor with SDL directly

francisdb:
The player window is an SDL window on every platform, so call SDL_ShowCursor/SDL_HideCursor at the two sites and drop the Win32 ShowCursor path and the SetMouseCursorVisible helper.

follow up on #3522


vbousquet:
Thanks!

PR #3527, opened 2026-06-15
pup: fix DMD capture quantization so capture triggers fire

### LINKED ISSUE #3524: Baywatch PUP doesn't trigger
**Describe the bug**
In the Baywatch PUP pack, the triggers never fire. Baywatch has a 192x64 DMD but the PUPCapture images are 128x32. 

**To Reproduce**
1. Load any version of Baywatch, with baywatch.zip rom
2. Load PUP pack from https://vpuniverse.com/files/file/6248-baywatch-puppack/
3. Play the game

If the bug appears when playing a specific table, give a link to a place where the exact version of the table file can be downloaded.

It's any Baywatch table as it's completely PUPCapture driven.

**Expected behavior**
PUP should trigger on B2S events. It's completely PUP driven.

**Versions & Settings**
 - VPX: tested on Windows BGFX mingw and Linux AARCH64 5141
 - Windows 11, OrangePi 5 Plus/ALP4k

**Additional context**
Issue also filed on libpupdmd https://github.com/PPUC/libpupdmd/issues/7

francisdb:
The 192x64/256x64 DMD downscale rounded the 2x2 box sum to the nearest level, but the pack capture images quantize a 4 level DMD at 3/6/11, not at the round-to-nearest 2/6/10. The mismatched edge levels meant the indexed frame hash never matched the reference images, so capture-driven triggers never fired (e.g. Baywatch). Quantize to match the capture.

fixes #3524

### DIFF
--- a/plugins/pup/PUPManager.cpp
+++ b/plugins/pup/PUPManager.cpp
@@ -474,53 +474,30 @@ int PUPManager::ProcessDmdFrame(const DisplaySrcId& src, const uint8_t* frame)
 
    if (src.width <= 256 && src.height == 64) // should be 192x64 or 256x64
    {
-      // Resize with a triangle filter to mimic what original implementation in Freezy's DmdExt (https://github.com/freezy/dmd-extensions)
-      // does, that is to say:
-      // - convert from luminance to RGB (with hue = 0, sat = 1)
-      // - resize using Windows 8.1 API which in turn uses IWICBitmapScaler with Fant interpolation mode (hence the triangle filter)
-      // - convert back from RGB to HSL and send luminance to PinUp
+      // The PuP capture reference images for these DMDs are 128x32: a 256x64 source maps to
+      // a full 128x32 frame, a 192x64 source to a 96x32 image centered in the 128x32 frame.
+      // Each output pixel is the 2x2 box sum of the source.
       //
-      // Some references regarding Fant scaling:
-      // - https://github.com/sarnold/urt/blob/master/tools/fant.c
-      // - https://photosauce.net/blog/post/examining-iwicbitmapscaler-and-the-wicbitmapinterpolationmode-values
-      //
-      // The Baywatch Pup pack was used to validate this (the filter is still a guess since Windows code is not available)
-      //
-      //
-      // This boils down to a 2x2 box filter in this case (256(or lower) x 64 -> 128 x 32), also no border checking needed then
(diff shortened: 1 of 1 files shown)


francisdb:
Across a full game we captured ~2000 distinct DMD frames and compared them to the pack's reference images. With the fix, the ones that should match now do (the rest are just mid-animation frames that aren't meant to trigger anything), and the themed videos play correctly.

I leave the final verdict to @evilwraith but this can already be merged.

mkalkbrenner:
shouldn't that rather be fixed in https://github.com/PPUC/libpupdmd ?

francisdb:
I think it should be fixed here as the existing code lives here. If this afterwards needs to be refactored and moved elsewhere I don't mind.

evilwraith:
> shouldn't that rather be fixed in https://github.com/PPUC/libpupdmd ?

I would have thought so as well. It seems like this is a solution for VPXS but a bug would still exist in the library itself.

francisdb:
> > shouldn't that rather be fixed in https://github.com/PPUC/libpupdmd ?
> 
> I would have thought so as well. It seems like this is a solution for VPXS but a bug would still exist in the library itself.

There's nothing to fix in libpupdmd here: it only ever receives the finished 128x32 indexed frame and hashes it. The 192x64 -> 128x32 downscale (and the wrong quantization) is vpinball-side code, so this is the correct place to fix it.

evilwraith:
> > > shouldn't that rather be fixed in https://github.com/PPUC/libpupdmd ?
> > 
> > 
> > I would have thought so as well. It seems like this is a solution for VPXS but a bug would still exist in the library itself.
> 
> There's nothing to fix in libpupdmd here: it only ever receives the finished 128x32 indexed frame and hashes it. The 192x64 -> 128x32 downscale (and the wrong quantization) is vpinball-side code, so this is the correct place to fix it.

Yeah after looking I realized that. I think your fix is good to go. We now have working triggers again. Thanks.

jsm174:
just curious, if vpx sent pupdmd the 192x64 frame, and pupdmd was updated to support 192x64, then we wouldn't need all this logic in vpx, correct?  (ie, I thought the purpose of pupdmd was to handle all this?)

francisdb:
Copy pasting my fix in libpupdmd and then removing the old code here would also work. But that's not a fix. That's a fix mixed with a refactor. Hard do backtrack later why things were done a certain way. Can somebody please validate this instead?

mkalkbrenner:
I think we should move it to libpupdmd.

PR #3531, opened 2026-06-15
Fix macOS crash sorting ball-reflection lights (#3528)

### LINKED ISSUE #3528: 10.8.1 BGFX Mac Arm64 - Table crashes in recent builds
This one here: https://www.vpforums.org/index.php?app=downloads&showfile=18804
Loads okay but crashes within seconds of launching the ball. Didn't see any errors in the log, but I could be mistaken.
Using 5165-2c83fd224 build on Mac Sonoma. Not sure which earlier build didn't crash, it was 1 or 2 months ago when I last opened it.

francisdb:
### Problem

On macOS (BGFX, arm64) tables crash with EXC_BAD_ACCESS within seconds of a ball being in play (#3528). The crash happens inside libc++'s sort, called from the ball-reflection light sort in Ball::Render.

### Root cause

Introduced by 54e23faaf (enabling -ffp-contract=fast). With contraction enabled, the squared-distance comparator used to sort the reflection lights can produce inconsistent results across the different places libc++ inlines it, so it is no longer a valid strict weak ordering. libc++'s sort relies on that and reads past the end of the array, dereferencing garbage.

This is not a data race (single writer, same thread). It is macOS/clang specific because libstdc++ tolerates the inconsistent comparator. It only reproduces at full speed.

### Fix

Make the reflection-light sort robust regardless of floating point flags, so the ordering stays consistent and total under contraction or fast-math. No build flags are changed.

### Testing

Reproduced reliably on macOS arm64 with Pinball Action (Tehkan 1985): crashed every attempt before the change, no crashes after playing the same scenario.

fixes #3528

### DIFF
--- a/src/parts/ball.cpp
+++ b/src/parts/ball.cpp
@@ -315,11 +315,22 @@ void Ball::Render(const unsigned int renderMask)
 
    // collect the x nearest lights that can reflect on balls
    vector<Light*>& reflectedLights = m_renderer->m_ballReflectedLights;
-   std::ranges::sort(reflectedLights.begin(), reflectedLights.end(), [this](const Light* const pLight1, const Light* const pLight2) {
-      const float dist1 = Vertex3Ds(pLight1->m_d.m_vCenter.x - m_hitBall.m_d.m_pos.x, pLight1->m_d.m_vCenter.y - m_hitBall.m_d.m_pos.y, pLight1->m_d.m_meshRadius + pLight1->m_surfaceHeight - m_hitBall.m_d.m_pos.z).LengthSquared(); //!! z pos
-      const float dist2 = Vertex3Ds(pLight2->m_d.m_vCenter.x - m_hitBall.m_d.m_pos.x, pLight2->m_d.m_vCenter.y - m_hitBall.m_d.m_pos.y, pLight2->m_d.m_meshRadius + pLight2->m_surfaceHeight - m_hitBall.m_d.m_pos.z).LengthSquared(); //!! z pos
-      return dist1 < dist2;
-   });
+   // Sort the candidate lights nearest-first. -ffp-contract=fast (CMakeLists) lets the compiler contract this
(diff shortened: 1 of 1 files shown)


jsm174:
I've seen this behavior recently too. Couldn't get it to replicate in debug mode, so this explains it.  Thanks for finding this.

@toxieainc - should we just go back to `-ffp-contract=off`. From a readability perspective this is a lot.

francisdb:
> I've seen this behavior recently too. Couldn't get it to replicate in debug mode, so this explains it. Thanks for finding this.
> 
> @toxieainc - should we just go back to `-ffp-contract=off`. From a readability perspective this is a lot.

If you roll that back make sure there is some clear documentation pointing here so that in a year we don't have another round of why did we have this flag, let's remove it and see.

I think it's better for the code to just work in both cases. The question is are there other affected code paths?

toxieainc:
This is interesting to say the least, especially given that clang and gcc use contraction by default nowadays. But maybe this is more aggressive if one also enables fast math.
So i would say to do both.

Goldilox2023:
This happens to me too with the Taxi table (VPW). It runs fine until it doesn't and crashes to the desktop. My log file is included. 
[LOG.txt](https://github.com/user-attachments/files/30592515/LOG.txt)

jsm174:
@toxieainc - I made a test build with `contract-fast` --> `contract-off`, and @Goldilox2023 reported this no longer happening:

https://github.com/vpinball/vpinball/commit/5a616b8fc588dfbe39c6360a9c3c4617a05b9503

Should I put this back in?

jsm174:
@Goldilox2023  - please check with latest. went back to ffp-contract=off:

https://github.com/vpinball/vpinball/commit/d6004882d4438dcfd62a51ded45ad8e000e0e6b7

Goldilox2023:
Thanks a lot, @jsm174!

PR #3533, opened 2026-06-15
texture: font texture atlas mip map issues

francisdb:
TNA renders it's desktop DMD like below on standalone. On windows these grey bars are not visible.

This is among others caused by the first square in the atlas being `FFFFFF00`

Screenshot_2026-06-16_00-11-10

Some images store transparent areas as opaque-coloured zero alpha. Bilinear filtering and mip generation then bleed that colour into the visible glyph edges as white seams (notably on the BGFX backend). Zero the rgb of fully transparent texels at load so filtering blends toward black; invisible since the texels have zero alpha.

`SF_NONE` is another option but that degrades now smooth looking white reels with numbers

*I'm not convinced any of these are the right fix but it at least starts the conversation*

This mostly fixes https://vpuniverse.com/files/file/14359-total-nuclear-annihilation-spooky-2017-vpw/

Screenshot_2026-06-16_00-00-43

After

Screenshot_2026-06-16_00-06-45

Still not perfect, low contrast lines remain

### DIFF
--- a/src/renderer/Texture.cpp
+++ b/src/renderer/Texture.cpp
@@ -412,6 +412,19 @@ std::shared_ptr<BaseTexture> BaseTexture::CreateFromFreeImage(FIBITMAP* dib, con
       FreeImage_Unload(dibResized);
    FreeImage_Unload(dib);
 
+   // Clear the colour of fully transparent texels. Some images (e.g. dot matrix reel font atlases)
+   // store transparent areas as opaque-coloured (often white), which bilinear filtering and mip
+   // generation then bleed into the visible edges as a halo/seam. Zeroing them makes any filtering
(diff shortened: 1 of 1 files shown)


francisdb:
If you don't like the solution, neither do I, but then propose how to fix this in a better way.

vbousquet:
Silently modifying the assets on one of the platform is a hack and I would really prefer avoiding something as bad as this.
I would say that the question is why it is different between Windows and other backends ? There is the question of the filtering but perhaps we introduced a bug on the texture coordinate evaluation somewhere (precision issue ?).

jsm174:
The lines and dots below the text have been visible in standalone since way before BGFX times, so probably not something we recently introduced.

francisdb:
Yes, this has always been like this in standalone.

PR #3534, opened 2026-06-16
standalone: add UserDefaultOnOff enum

francisdb:
Default/Off/On were absent from the generated dispatch proxy, so tables referencing them failed with 'Variable is undefined' compile errors on standalone. Enable the enum in the IDL parser and regenerate the proxy.

Some tables do this

```vbscript
SpeakerColor = Default          ' initialise to the "default/unset" marker (-1)

' lines 3678-3684 and 3702-3708 - inside the ball-drain logic (VR room)
If VRRoom > 0 Then
      If SpeakerColor = Default Then
          ' already default: do nothing
      Else '???
          SpeakerColor = Default   ' reset back to default
          SpeakerChangeColor
      End If
End If
```

*If it is provided some day it will be (ab)used*

Fixes these and possibly others:
* https://github.com/jsm174/vpx-standalone-scripts/blob/91043e45051021511c9e63397ee2ee19a9f5f460/Bud%20Spencer%20%26%20Terence%20Hill%20(Original%202024)/Bud%20and%20Terence%20(Original%202024)%20v1.6.vbs.patch
* https://github.com/jsm174/vpx-standalone-scripts/blob/91043e45051021511c9e63397ee2ee19a9f5f460/IT%20Pinball%20Madness%20(JP%20Salas%2CJoe%20Picasso)1.2/IT%20Pinball%20Madness%20(JP%20Salas%2CJoe%20Picasso)1.2.vbs.patch

Side note, these vpx-standalone-scripts patched vbs files now make the table throw errors. `Compile error: Name redefined`

### DIFF
--- a/standalone/idl/parser/src/main/java/org/vpinball/IDLParserToCpp.java
+++ b/standalone/idl/parser/src/main/java/org/vpinball/IDLParserToCpp.java
@@ -995,7 +995,7 @@ public class IDLParserToCpp {
 								"TriggerShape",
 								"RampType",
 								"PlungerType",
-								//"UserDefaultOnOff",
+								"UserDefaultOnOff",
 								//"FXAASettings",
 								//"PhysicsSet",
 								"TargetType",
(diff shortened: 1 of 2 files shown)


jsm174:
Nice find. Thanks!

francisdb:
Typical time sink rollercoaster all over the place (including a wine MR) to end up at something simple but unexpected

PR #3544, opened 2026-06-19
tests: add doctest unit test harness run on CI

francisdb:
Stand up a cross-platform doctest unit test target (vpx-test) and run it on the linux bgfx CI build. doctest is pulled at configure time via CMake `FetchContent` (header only) rather than vendored. The target is opt-in via -DVPINBALL_BUILD_TESTS and reuses the app build configuration.

Seed it with a test pinning the compiled-in Settings defaults, which the registry exposes without booting the player or a display.

see https://github.com/vpinball/vpinball/pull/3386#issuecomment-4749719478


vbousquet:
Nice, this is clearly a direction I would like to test and the reason I created vpx-test.

Regarding doctest, it is supposed to be part of the third-party lib. I just realize I forgot to git add it. We can fetch it, but then it would be more logical to have it asa part of our platform system (but since it is not platform dependent a simple integration in the third-party doc would do the job)

Regarding the tests, as much as possible, it would be nice to have a single cross-platform test suite (doctest allow to easily select what we want to run). The existing vpx-test already contains some 'integration' test that is to say, validation of the player output in a set of config. I made it this way since VPX does not fullfill design patterns that allows easy testing. I'm also moving things in this direction but there is some road to go before having a codebase that could have a significant test coverage.

francisdb:
Thanks!

- **doctest**: done - vendored 2.4.11 under `third-party/include/doctest` (with a `third-party/README.md` entry) and dropped the FetchContent step, matching the other header-only deps.
- **One suite**: agreed. The unit tests here can fold into the same binary as the existing integration tests (doctest can select what to run per invocation).
- **One restructuring I'd like your take on**: `vpx-test` currently recompiles the whole engine a second time (it's a separate target reusing the app's source list), so CI builds the engine twice in the test jobs. We could instead compile the engine once into a shared CMake `OBJECT` library that both the app exe and the test exe link - each TU built once, roughly halving the test job. The catch is it changes how the app target itself is assembled (linking the object library instead of listing sources), across the per-platform `make/CMakeLists_*` files - so only worth doing if you're on board with that direction.

francisdb:
@vbousquet any comments on that shared CMake OBJECT to avoid compiling twice?

vbousquet:
@jsm174 is the man for build script. This is not really my skill area. @jsm174 what do you think about this ?

francisdb:
🪦

PR #3545, opened 2026-06-19
build: auto-select the target, defaulting to bgfx for the host

francisdb:
The build currently requires copying the right `make/CMakeLists_<target>.txt` over the root `CMakeLists.txt` before configuring. This adds a small tracked top-level `CMakeLists.txt` that picks that file for you: it detects the host platform and architecture and defaults to the BGFX renderer, so a plain `cmake -B build` just works.

Anything else is an override via `-DTARGET` (renderer backend), `-DPLATFORM` and `-DARCH`. An unknown combination fails with the list of available targets, and if the third-party dependencies have not been fetched yet it fails early pointing at the platform `external.sh` (instead of a confusing missing-header error deep in the build).

The CI jobs are migrated to the dispatcher (explicit `-DTARGET`/`-DPLATFORM`/`-DARCH`). iOS/Android keep using `make/CMakeLists_bgfx_lib.txt` directly, since they must set their cross-compile toolchain before `project()`, which the dispatcher cannot.

The previous flow of copying a per-target file over the root still works as a fallback (the per-target files are otherwise untouched). This is meant as a step toward a single unified `CMakeLists.txt`.

### DIFF
--- a/.github/workflows/vpinball.yml
+++ b/.github/workflows/vpinball.yml
@@ -189,27 +189,26 @@ jobs:
           perl -i -pe"s/unknown/${{ needs.version.outputs.sha_short }}/g" src/core/git_version.h
       - name: Build
         run: |
-          TYPE=$(echo "${{ matrix.type }}" | tr '[:upper:]' '[:lower:]')
-          cp make/CMakeLists_${TYPE}-${{ matrix.platform }}-${{ matrix.arch }}.txt CMakeLists.txt
+          CMAKE_ARGS="-DTARGET=${{ matrix.type }} -DPLATFORM=${{ matrix.platform }} -DARCH=${{ matrix.arch }}"
           if [[ "${{ matrix.platform }}" == "windows-mingw" ]]; then
             CURRENT_DIR="$(pwd)"
             MSYSTEM=UCRT64 /c/msys64/usr/bin/bash.exe -l -c "
               cd \"${CURRENT_DIR}\" &&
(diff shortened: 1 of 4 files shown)


jsm174:
This is cool. 

I wonder if it should be more like all the supporting libraries we did, ie: `-DPLATFORM=android -DARCH=arm64-v8a`. and then with a `-DTARGET=BGFX`.

Thinking about this more, we are pretty close to having a single unified `CMakeLists.txt ` file (excluding plugins). 

(We always had separate files because in the beginning it was helpful to keep track of the differences in platforms. That really isn't the case today)

If we were to unify, it would be nice to be like libaltsound, libdmdutil, libdof, etc..

francisdb:
Good call - switched the override to `-DTARGET` / `-DPLATFORM` / `-DARCH` (all defaulting to BGFX for the host), matching the convention the iOS/Android builds already use, and migrated the CI jobs to it (keeping the explicit platform/arch). iOS/Android stay on the `bgfx_lib` file directly though: they set their cross-compile toolchain before `project()`, which the dispatcher (which has to call `project()` itself) can't do.

On the single unified `CMakeLists.txt`: agreed that's the end goal, and this dispatcher is the scaffolding for it - everything now goes through one entry point that selects among the per-platform files. I'd do the actual merge into one file (like libdmdutil/libdof/etc.) as a follow-up so this PR stays reviewable. Sound good?

PR #3554, opened 2026-06-22
Nudge refactor

vbousquet:
This PR provides an entire rewrite of how nudge is implemented from sensor, UI to physics.

This rewrite was made to fulfill:
- the need to support multiple sensors (gamepad + VR for example, switching when they are actually active),
- the need to support sensors tied to a machine not matching the emulated one (VR cab, gamepad, light or small pincab, ...),
- the will to move nearer a cabinet model (for the time simple rectangle without torque, but better than the previous box model and ready to be improved further more),
- the will to better support modern hardware that offer acceleration together with velocity,
- the opportunity to use recordings from real hardware,
- the will to improve the plumb simulation (from simple 3D newtonian to a pendulum),
- the will to have some live feedback when setting up a sensor as it can be quite challenging to adjust,
- the will to cleanup the code a bit more.

This needed a full rewrite as adding these features as is would have lead to a monster. I took the opportunity to move the cabinet physics appart from playfield physics. The cabinet physic uses standard SI units (no VPU, VPT,...) and is converted when used in the playfield physics (applying to ball).

It will likely need some testing to be validated against the large variety of setups that exist, but I hope it's finally good enough to be merged.


toxieainc:
Absolutely fantastic work, thanks!!

dekay:
Great work indeed!  However, I am running into a problem with my plunger not working when a velocity input is specified and have written up a bunch of detail in https://github.com/vpinball/vpinball/issues/3590.

Another thing I'm seeing is that Visual Nudge seems to be *very* sensitive compared to what I'd expect from the movement of the ball on the table. I had to adjust Visual Nudge strength down to 20% to get something more reasonable. This on a Pinscape Pico with an MMA8541Q accelerometer (same as KL25Z Pinscape) with both acceleration and velocity inputs specified.

PR #3564, opened 2026-06-24
Display list: show native pixel resolution instead of logical

francisdb:
On Wayland fractional scaling the display bounds are logical points (e.g. a 4K panel at 1.75x reports 2194x1234), so the monitor was listed at the scaled resolution. Use the largest fullscreen mode (reported directly in pixels) for the display label; window sizing keeps using the logical bounds.

### Before

image

### After

image

### DIFF
--- a/src/renderer/Window.cpp
+++ b/src/renderer/Window.cpp
@@ -450,6 +450,13 @@ vector<Window::DisplayConfig> Window::GetDisplays()
          displayConf.height = displayBounds.h;
          displayConf.isPrimary = primaryID != 0 ? displayIDs[i] == primaryID : (displayBounds.x == 0) && (displayBounds.y == 0);
          const SDL_DisplayMode* mode = SDL_GetDesktopDisplayMode(displayIDs[i]);
+         // The display bounds are in logical points (e.g. a 4K display at 1.75x reports 2194x1234). The
+         // native pixel resolution is the largest fullscreen mode, which is reported directly in pixels.
+         int modeCount = 0;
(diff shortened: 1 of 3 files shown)


francisdb:
macOS

This is a bit confusing with the window dimensions being logical, will add the scale

### Before

image

### After

image

francisdb:
Now looks like this, I think that's ok.

image

vbousquet:
This really looks wrong to me. The max here would not work with inverted scaling. It does not make really sense for Retina display either.

We should really have a clear definition of what is the logical/pixel sizes, including retina and DPI scaling, and this should not be part of VPX at all. This is supposed to be handled by SDL.

francisdb:
For me the first screenshot of the issue is unexpected, I have no displays with that resolution. Here is the native display settings page: 

image

> This really looks wrong to me. The max here would not work with inverted scaling. It does not make really sense for Retina display either.

What exactly looks wrong?

> We should really have a clear definition of what is the logical/pixel sizes, including retina and DPI scaling, and this should not be part of VPX at all. This is supposed to be handled by SDL.

Have you checked the last screenshot? That one has the scale in it.
Also this pr only changes the display line in case you are talking about the other stuff.

francisdb:
@vbousquet what exactly do you mean with "This really looks wrong to me."

vbousquet:
> @vbousquet what exactly do you mean with "This really looks wrong to me."

The code changed in between, so it's difficult to say now. If I remeber well, there was some processing of the values with a max of the size which looked wrong to me.

Anyway, the latest changes unify size handling by not exposing logical size at all anywhere in the UI or settings. Therefore I will close this as it is not useful anymore

PR #3566, opened 2026-06-24
window: use fullscreen desktop instead of exclusive mode on BGFX (#3451)

### LINKED ISSUE #3451: regression: "Fullscreen" windows no longer shown fullscreen
Linux
Master
Wayland / COSMIC
BGFX
Desktop setup, default config

By default the vpinball main/playfield window normally starts fullscreen

On Linux BGFX the playfield stopped going fullscreen since the recent removal of the fullscreen option in dbd2a00aaf32cedaae441f00f3de5a59b2218bb4. BGFX forces fullscreen false.
The window gets full screen dimensions but ends up being placed under panels and clipped off-screen.

This also affects some optimizations that can be applied only if the window is fullscreen.

Image

To fix this the window needs to be created using `SDL_WINDOW_FULLSCREEN`

As far as I know most windows games still have a window/fullscreen flag (next to the exclusive fullscreen that has mode-switching which is deprecated)

Other games I have run perfectly fullscreen (and they have the windowed toggle in the settings).

I created this issue as the related PR's don't get through (#3434 and #3366). No idea what the proper wanted solution would be.

_Please don't conflate this issue with the absolute window placement issues. That is [inherent to Wayland](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/247) and nothing much can be done except using hacks like compositor-specific window placement rules._

### LINKED ISSUE #3560: wayland: fullscreen not rendered hidpi
linux
master
bgfx
wayland

The table looks less detailed
Display resolution shown is the scaled resolution. This is a 4K display.
The config ui and fps debug info are:
* double as big (175% probably as that's my scaling factor)
* look blurry
* have more spacing, same as #2978 ?

### Fullscreen

Image

Image

### Not fullscreen

Image

Image

continuation of #3451

francisdb:
The post-creation fullscreen activation passed SDL_GetCurrentDisplayMode() to SDL_SetWindowFullscreenMode(), which requests exclusive fullscreen. On macOS the exclusive path does not take over the menu bar and Dock for our borderless window, so the playfield rendered behind them.

On BGFX pass NULL instead for borderless fullscreen desktop (cover the current display, no video mode change), which hides the menu bar and Dock and matches the intent of 'fullscreen without a video mode change'. The target display is the one the window was created on, so multi-monitor setups keep each window on its own display. Legacy (DX9/GL) backends keep exclusive fullscreen with the selected mode.

Wayland was unaffected because the compositor drives fullscreen regardless of the requested mode.

fixes #3451 (part 2)
fixes #3560 (wayland hidpi)

- [x] Validate on mac
- [x] Validate on Wayland
- [ ] Validate on Windows (somebody else)

### DIFF
--- a/src/renderer/Window.cpp
+++ b/src/renderer/Window.cpp
@@ -277,10 +277,20 @@ Window::Window(const string& title, const Settings& settings, VPXWindowId window
    m_sdrWhitePoint = SDL_GetFloatProperty(props, SDL_PROP_WINDOW_SDR_WHITE_LEVEL_FLOAT, 1.0f);
    m_hdrHeadRoom = SDL_GetFloatProperty(props, SDL_PROP_WINDOW_HDR_HEADROOM_FLOAT, 1.0f);
 
-   // Switch to request fullscreen display mode (must be done after window creation)
-   if (fullscreenDisplayMode)
+   // Activate fullscreen after window creation.
+   if (m_fullscreen)
    {
+#ifdef ENABLE_BGFX
+      // Borderless fullscreen desktop: cover the current display with no video mode change. A non-NULL
+      // mode requests exclusive fullscreen instead, and on macOS that path does not take over the menu
+      // bar and Dock for our borderless window, leaving it rendered behind them. NULL hides them as
+      // expected and matches the intent of "fullscreen without a video mode change" (see #3451). The
+      // target display is the one the window was created on (its center), so this stays per-display.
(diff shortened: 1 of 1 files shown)


vbousquet:
Not needed since the fullscreen cleanup

francisdb:
Re-implemented by 8446b31595b62803db35af6a38a4de99e8256228

PR #3567, opened 2026-06-24
renderer: fix EXC_BAD_ACCESS on shutdown

francisdb:
Crash bug found on macOS when shutting down vpinball with 2 screens / windows

Player::~Player freed render resources (render targets via RenderProbe:: RenderRelease / IRenderable::RenderRelease) before deleting the render device, which is what stopped the render thread. An in-flight frame on the render thread could then dereference a freed RenderTarget and segfault on exit (RenderDevice::NextView via RenderTarget::Activate, EXC_BAD_ACCESS).

#3556 added the render-thread stop wait, but only inside ~RenderDevice, which runs after these resources are already freed. Extract that wait into an idempotent RenderDevice::StopRenderLoop() and call it at the start of the renderer teardown in Player::~Player, before any render resource is released. The destructor still calls it as a backstop.

Most easily reproduced with two fullscreen windows (playfield + backglass on separate displays), where the render thread is busy at every instant of shutdown; single-window vsync play rarely parked mid-frame so it seldom hit.

### DIFF
--- a/src/core/player.cpp
+++ b/src/core/player.cpp
@@ -986,6 +986,12 @@ Player::~Player()
    delete m_physics;
    m_physics = nullptr;
 
+   // Stop the render thread before releasing any render resource below (render targets are freed by the
+   // RenderRelease() calls). Otherwise an in-flight frame on the render thread can use a freed render
+   // target and crash on exit (most easily reproduced with a second fullscreen window). The render device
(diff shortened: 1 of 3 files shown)


vbousquet:
Just pushed a commit that should fix this by using the existing mutex. This should not be needed anymore

PR #3572, opened 2026-06-24
liveui: stop in-game UI spacing drifting

### LINKED ISSUE #3565: Changing playfield display size +/- messes up the gui
Linux
master
Wayland

1. `F12`
2. display settings
3. Playfield
4. Click on the +/- buttons

You end up wit ha compressed ui after a few clicks and width/height look strange (wayland has no concept of window placement)

Also on macOS there are funny things going on in that screen where suddenly the mouse starts to move up

### Before

Image

### After clicking a few times

Image

francisdb:
fixes #3565


vbousquet:
This should be properly fixed now so I will close

PR #3573, opened 2026-06-24
liveui: hide window position controls when positioning is unsupported

francisdb:
On backends that do not honor app-driven window positioning (e.g. Wayland), the X/Y position sliders were just stuck at 0 and did nothing. Detect the capability through SDL_SetWindowPosition's return value (cached on the window) and skip the position controls and drag hint when it is unsupported, without checking the video driver name.


vbousquet:
This was implemented inside Window class so closing

PR #3574, opened 2026-06-24
renderer: fix assertion failing on startup for fractional display scaling

francisdb:
This fixes an assertion error on displays with fractional scaling. Vpinball would just not start after saving the window size (development).

Under fractional display scaling, the Vulkan swapchain extent can be a pixel off from SDL_GetWindowSizeInPixels (e.g. a 1920px window on a 1.75x display giving a 3361px backbuffer vs 3360px reported), tripping the SetBackBuffer assertion in debug builds. The swapchain is what gets presented, so treat its size as authoritative and sync the window pixel size to it; 

A different fix could be storing physical sizes?

### DIFF
--- a/src/renderer/Window.cpp
+++ b/src/renderer/Window.cpp
@@ -500,7 +500,15 @@ Window::DisplayConfig Window::GetDisplayConfig(const string& display)
 
 void Window::SetBackBuffer(RenderTarget* rt, const bool wcgBackbuffer)
 {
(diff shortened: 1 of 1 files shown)


vbousquet:
I would expect this to be fixed by now. If not we will need to solve the core problem (have a matching RT size) and not hide the bug behind a hack so I will close this PR

PR #3579, opened 2026-06-26
fix: BGFX fullscreen leaves menu bar and Dock visible on macOS

Techbert08:
When running fullscreen in BGFX on MacOS, I observed the menu bar and dock sitting on top of the game window.

For BGFX builds, fullscreenDisplayMode is set to the current display mode (to read screen dimensions) and was then passed to SDL_SetWindowFullscreenMode. This sets window->fullscreen_exclusive=true in SDL3, which causes the macOS Cocoa backend to bypass Spaces fullscreen and fall back to a plain borderless window — one that doesn't hide the menu bar or Dock.

BGFX owns the fullscreen/mode decision internally and does not need SDL to request an exclusive mode change. Guard the SDL_SetWindowFullscreenMode / SDL_SetWindowFullscreen / SDL_SyncWindow block with #ifndef ENABLE_BGFX, consistent with the existing #ifndef ENABLE_BGFX guards in the same function that already express this same design boundary.

I tested by building locally on my Mac, and the menubar+dock are now properly hidden.  I don't believe this is harmful for other platforms using BGFX.

Please let me know if this isn't the right way to send in external contributions and happy to change it.

### DIFF
--- a/src/renderer/Window.cpp
+++ b/src/renderer/Window.cpp
@@ -278,12 +278,17 @@ Window::Window(const string& title, const Settings& settings, VPXWindowId window
    m_hdrHeadRoom = SDL_GetFloatProperty(props, SDL_PROP_WINDOW_HDR_HEADROOM_FLOAT, 1.0f);
 
    // Switch to request fullscreen display mode (must be done after window creation)
+   // BGFX owns the fullscreen/mode decision internally; SDL must not request an exclusive
+   // mode change here (SDL_SetWindowFullscreenMode sets fullscreen_exclusive=true which
+   // overrides platform fullscreen behavior, e.g. bypassing macOS Spaces fullscreen).
+#ifndef ENABLE_BGFX
    if (fullscreenDisplayMode)
    {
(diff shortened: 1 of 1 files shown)


vbousquet:
The fullscreen implementation existing in the codebase is currently wrong as this is a legacy exclusive fullscreen that most other platforms do not support at all (and is mostly deprecated since Windows 10). This is a known issue that needs a proper refactor between windowed / borderless fullscreen / exclusive fullscreen (Windows only, no BGFX) which is just pending some time to be implemented cleanly

francisdb:
Duplicate of #3566 which also fixes the hidpi render

Techbert08:
Sounds good, closing as duplicate of that fix.

Techbert08:
Is there any chunk I could potentially take on?  From what I can tell there are a few different efforts under way (GL -> BGFX, COM -> plugins, maybe updates to the supporting PinMAME/Serum libraries, as well as the rendering refactor you're talking about).  Up to you of course, just enjoyed hacking on the project to develop some ROM modifications and wanted to try giving back.

PR #3580, opened 2026-06-26
cmake: Add THIRD_PARTY_CHECK option.

raiford:
Add an option for skipping the third-party build directory check.  When building for nix external.sh build scripts are not used.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,8 +82,8 @@ set(VERSION_REV "${CMAKE_MATCH_1}")
 project(vpinball VERSION "${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_REV}")
 
 # Fail early with a clear message if the third-party deps aren't built yet (every backend needs SDL3)
-
-if(NOT EXISTS "${CMAKE_CURRENT_LIST_DIR}/third-party/include/SDL3/SDL.h")
+option(THIRD_PARTY_CHECK "Verify that third-party deps are built." ON)
+if(THIRD_PARTY_CHECK AND NOT EXISTS "${CMAKE_CURRENT_LIST_DIR}/third-party/include/SDL3/SDL.h")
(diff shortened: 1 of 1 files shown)


francisdb:
@raiford where is that nix build config? I wonder what it does instead.

jsm174:
Other build platforms like batocera expect everything in their normal install folders.  `/usr/include`, `/usr/lib`, etc. 

I have some thoughts and will implement something over the weekend.

jsm174:
Other build platforms like batocera expect everything in their normal install folders.  `/usr/include`, `/usr/lib`, etc. 

I have some thoughts and will implement something over the weekend.

raiford:
> @raiford where is that nix build config? I wonder what it does instead.

My flake.nix builds for vpinball are here: https://github.com/raiford/vpinball-flake

Nix has a rather unique build system where packages are all kept in `/nix/store` in special paths that are both versioned and hashed.  This means you can have any number of versions of a package installed dependencies are linked by complex RPATH, or PATH,  trickery that is unique to each build.

I have basically ported all of the standalone build scripts to the nix build system: https://github.com/raiford/vpinball-flake/tree/main/pkgs

I then have a script that reads hashes from the various `config.sh` scripts and pin those in to the `flake.lock`: https://github.com/raiford/vpinball-flake/blob/69ba9cc5d68e173f7eb4df97ab15f3b58160d812/flake.nix#L324

I've got my builds running off of this fork so I'm unblocked for now and happy to close this PR if that is the suggested path for now.  I can wait to see what @jsm174 is working on.  :-)

jsm174:
Can you see if https://github.com/vpinball/vpinball/pull/3586 helps out:

now you should be able to use something like:

```
-DUSE_SYSTEM_LIBS=ON
-DSYSTEM_LIBS_INCLUDE_DIR=$(STAGING_DIR)/usr/include
-DSYSTEM_LIBS_LIB_DIR=$(STAGING_DIR)/usr/lib
```

raiford:
@jsm174 #3586 fixes my nix builds.  Thanks!

I'll close this PR.

PR #3588, opened 2026-06-28
Fix occasionnal crash on exit (race between RenderThread using render ressources and Player freeing them)

vbousquet:
See https://github.com/vpinball/vpinball/pull/3567

### DIFF
--- a/src/core/player.cpp
+++ b/src/core/player.cpp
@@ -366,7 +366,7 @@ Player::Player(PinTable *const table, const PlayMode playMode)
 
    try
    {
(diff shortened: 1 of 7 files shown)


jsm174:
@vbousquet - just checked this and it makes MacOS completely lock up on table exit.

vbousquet:
@jsm174 This should be solved now (and cleaner). Could you check on MacOS ?

jsm174:
Exit is now working again on MacOS! Thanks!

PR #3589, opened 2026-06-28
accept events from PPUC

mkalkbrenner:


plugins/plugins/MsgPluginManager.cpp
```diff
@@ -197,9 +212,21 @@ void MsgPluginManager::SendMsg(const uint32_t endpointId, const unsigned int msg
    assert(pm.m_msgs[msgId].refCount > 0);
    assert(1 <= endpointId && endpointId <= pm.m_plugins.size());
 
-   for (const CallbackEntry& entry : pm.m_msgs[msgId].callbacks)
+   const std::list<CallbackEntry> callbacks = pm.m_msgs[msgId].callbacks;
+   for (const CallbackEntry& entry : callbacks)
       if (entry.endpointId == targetEndpointId)
       {
+         if (entry.callback == nullptr || reinterpret_cast<uintptr_t>(entry.callback) < 4096)
```
vbousquet: See above

plugins/plugins/MsgPluginManager.cpp
```diff
@@ -185,8 +186,22 @@ void MsgPluginManager::BroadcastMsg(const uint32_t endpointId, const unsigned in
    assert(pm.m_msgs[msgId].refCount > 0);
    assert(1 <= endpointId && endpointId <= pm.m_plugins.size());
 
-   for (const CallbackEntry& entry : pm.m_msgs[msgId].callbacks)
+   const std::list<CallbackEntry> callbacks = pm.m_msgs[msgId].callbacks;
```
vbousquet: I'm not sure to fully understand this code block. It looks like a defensive protection against an invalid callback list, using a validation scheme (callback not null, with a somewhat valid address) and a somewhat valid id). If so, why is this needed ? And this is really the intent, the validation scheme looks really hacky to me. Couldn't we identify and solve the thing that makes this code needed ?

plugins/pup/PUPPlugin.cpp
```diff
@@ -60,6 +60,13 @@ LPI_IMPLEMENT_CPP // Implement shared log support
 
 MSGPI_STRING_VAL_SETTING(pupPathProp, "PUPFolder", "PinUp Player Folder", "", true, "", 1024);
 
+typedef struct PUPQueueEventMsg
```
vbousquet: It would be cleaner to have a very short header that can be shared between the plugin and components that want to use the exposed message (so simply moving PUPQueueEventMsg and the defines of the namespace and the message to a header)

vbousquet (commented):
Allowing to run plugins on a variety of host is really great. I have added a few comments, but I will need @jsm174 for the makefile changes as this is not my skill area.

make/CMakeLists_plugins.txt
```diff
@@ -35,3 +37,17 @@ include("${CMAKE_SOURCE_DIR}/make/CMakeLists_plugin_Serum.txt")
 include("${CMAKE_SOURCE_DIR}/make/CMakeLists_plugin_VNI.txt")
 include("${CMAKE_SOURCE_DIR}/make/CMakeLists_plugin_UpscaleDMD.txt")
 include("${CMAKE_SOURCE_DIR}/make/CMakeLists_plugin_WMP.txt")
+
+set(PPUC_MEDIA_PLUGIN_TARGETS)
```
vbousquet: Is this really needed ? It would be better to not add anything specific from PPUC inside this build script. For example, the iOS builds do perform static linking without needing any custom commands inside the plugin makefile which stays generic.

plugins/pup/PUPPlugin.cpp
mkalkbrenner: done

make/CMakeLists_plugins.txt
mkalkbrenner: removed

mkalkbrenner:
@vbousquet @jsm174 now there're conflicts. Before I take care of them, is there a chance to get this merged?

PR #3621, opened 2026-07-03
Wayland: keep keyboard focus on the playfield (priming frame + focus handler)

Le-Syl21:
## Problem

On native Wayland the playfield loses keyboard focus a few seconds after startup, which pauses the table (`willPlay` gates on `IsFocused()`): DMD and backglass appear frozen while the ROM keeps playing sound. Cabinet setup: GNOME/mutter 46, NVIDIA, 3 displays, SDL video driver forced to `wayland`.

Root cause, established with a compositor-side trace of the map/focus sequence:

- On Wayland a toplevel is only **mapped when its first buffer is committed** — that is the first Vulkan present, *not* `SDL_ShowWindow()`. So the mapping order cannot be controlled by the order of the `Show()` calls.
- Ancillary windows are revealed lazily, when their content first shows up (typically 1–2 s after the playfield, once PinMAME/B2S produce frames). Their deferred first present maps them *after* the playfield…
- …and mutter gives keyboard focus to the **last mapped toplevel of the active application**. `SDL_WINDOW_NOT_FOCUSABLE` cannot help here: the Wayland protocol has no way for a client to refuse keyboard focus (it works on X11/XWayland, which is why the bug only shows on native Wayland).
- Handing focus back afterwards via `xdg_activation` is unreliable: at that instant SDL usually holds no valid input event serial, so the compositor is entitled to reject the activation (same anti-focus-stealing rule that affects launcher tokens).

## Fix (both parts Linux-gated)

1. **Priming frame** (`Player::Player`, right before the playfield's first present): render a black clear on every windowed ancillary output and submit that frame. Only the touched swapchains get presented, so the ancillary toplevels are mapped (black) first, and the playfield — committed last through the same ordered client socket — naturally ends up with the keyboard focus. Deterministic, no token needed.
2. **Focus handler** (`Player::OnFocusChanged`): if the playfield still loses focus to another window of the application (never a legitimate user target, ancillary windows are output-only), raise it back. With the priming frame in place this only catches a rare transient steal during startup and resolves it in ~100 ms.

Validated on the cabinet through full game sessions (launcher → table start → gameplay): the playfield gets and keeps the keyboard focus, the table never pauses, DMD/backglass animate.

## Context: previous attempt #3608

I closed #3608 because the problem suddenly appeared to be fixed. That was a false positive: SDL had silently fallen back to **XWayland** (SDL prefers x11 when the compositor lacks the `fifo-v1` protocol, i.e. GNOME < 48), and on X11 `SDL_WINDOW_NOT_FOCUSABLE` works, masking the whole issue. Under a real native Wayland session (`SDL_VIDEO_DRIVER=wayland`) the bug is always there; this PR addresses the actual mapping/focus mechanics instead of the show ordering.

https://claude.ai/code/session_01JSBKQPTy9VSD53FZLjoS4y

### DIFF
--- a/src/core/player.cpp
+++ b/src/core/player.cpp
@@ -798,6 +798,36 @@ Player::Player(PinTable *const table, const PlayMode playMode)
 
    m_progressDialog.SetProgress("Starting..."s, 100);
 
+#if defined(__linux__) && !defined(__ANDROID__) && defined(ENABLE_BGFX)
+   // Priming frame for Wayland: a toplevel is only mapped when its first buffer is
+   // committed (the first Vulkan present, not SDL_ShowWindow), and the compositor
+   // gives keyboard focus to the last mapped toplevel of the active application
+   // (SDL_WINDOW_NOT_FOCUSABLE cannot be honored by the protocol). Ancillary windows
+   // normally present their first frame seconds after the playfield, when their
+   // content shows up, which steals the playfield focus and pauses the table.
+   // So present a black frame on every ancillary window NOW, before the playfield's
+   // first present: they get mapped first and the playfield, committed last through
+   // the same ordered client socket, naturally ends up with the keyboard focus.
+   {
+      bool primed = false;
+      for (VPX::RenderOutput* out : { &m_backglassOutput, &m_scoreViewOutput, &m_topperOutput })
+      {
+         if (out->GetMode() == VPX::RenderOutput::OM_WINDOW && out->GetWindow() != nullptr)
+         {
+            out->GetWindow()->Show();
+            m_renderer->m_renderDevice->SetRenderTarget("Prime Ancillary"s, out->GetWindow()->GetBackBuffer(), false, true);
+            m_renderer->m_renderDevice->Clear(clearType::TARGET, 0x00000000);
+            primed = true;
+         }
+      }
+      if (primed)
+      {
+         SubmitFrame();
+         LockFrameMutex();
+      }
+   }
+#endif
+
    if (m_renderer->IsUsingStaticPrepass())
    {
       // Perform a quick render to avoid displaying a blank screen while the static prerendering will be performed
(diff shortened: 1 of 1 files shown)


vbousquet:
Converted to draft as per our contribution policy. Will have a look when I get some time off.

Le-Syl21:
Marking this ready for review — six weeks of daily use on a cabinet since I opened it.

It has been running on my Wayland cabinet (GNOME/Mutter, playfield + backglass + score view) since early July without a regression, and it also ships in the fork builds people have been downloading from my Actions page to test the sensor scale work, so it has had eyes other than mine on it.

Nothing changed in the patch itself; both changes are still Linux-gated and no-ops elsewhere. Happy to rebase, split the priming frame from the focus handler, or adjust anything you'd rather see done differently.

vbousquet:
Just had a deeper look and this PR looks wrong to me for the following reasons:
- this is a local setup hack with no guarantee this will broadly work on other setups (other window managers)
- this address a setup issue by importing a hack inside VPX codebase while I guess it could be handled through local tools (I'm not a Linux, so I may be wrong but still I would be surprised VPX is the only app needing window placement/focus management if not allowed on some specific setups)
- our design path is to remove per platform hacks and use clean standard code, based on robust cross platform library. Here for window management, we use SDL. If something is broken in SDL, then it must be solved there, not imported as a hack in VPX.
- this breaks the cross platform behavior by modifying it for Linux users (ancilary window ends up being always shown on Linux while they should only be displayed if actually rendered). This is something we do not want (as it makes debugging, issue reporting, tutorial, community, ... more difficult if things are different on each platform)
- this adds a focus manager hacks (if focus lost then requets it back) which is what we have been searching to remove in 10.8.1 as this is against normal UX policy (since loosing focus is a normal UX operation we are supposed to honor).

Le-Syl21:
Fair enough — closing it.

I don't think you're wrong about the direction. The focus handler in particular is the
weakest part: re-requesting focus after losing it does fight normal UX policy, and if
10.8.1 is moving away from that, adding one here works against you.

On where it belongs: I'll take the SDL route and see what I can do there. If the priming
frame turns out to paper over something SDL gets wrong on Wayland, that's the place to fix
it, and everyone gets it rather than just VPX on Linux.

Thanks for taking the time to write out the reasoning rather than just closing it.

Le-Syl21:
Following up as promised: libsdl-org/SDL#16200.

It turned out to be a real SDL bug rather than a setup issue. `SDL_WINDOW_NOT_FOCUSABLE`
is only implemented for popup menus on Wayland — on a regular toplevel it is accepted at
creation, stays visible in `SDL_GetWindowFlags()`, and does nothing, while
`SDL_SetWindowFocusable()` on the same window fails with an explicit error. X11 has
honored it forever through `WM_HINTS`. So VPX was asking for the right thing and SDL was
silently not delivering it.

Measured on GNOME/Mutter with a six-step focus scenario: the X11 backend passes all six,
Wayland passes four, and the patch brings it to six. Nothing application-side.

If it lands, both changes from this PR become unnecessary — the focus handler obviously,
and the priming frame too, since mapping order stops mattering for focus.

PR #3630, opened 2026-07-06
Fix use-after-free crash when the static prepass is re-run (e.g. PUP pack start)

dynajoe:
### Fix
`RenderStaticPrepass` deletes `m_staticPrepassRT` synchronously on the game thread. With BGFX, already-submitted frames can still hold copy commands **sourcing** that RT (e.g. the static prepass flush blit at `Renderer::PrepareFrame`). If the prepass is invalidated and re-run after the first pass, the render thread executes a blit from the freed RT → crash in `RenderTarget::CopyTo`.

Defer the delete to the render thread's end-of-frame — the same idiom the function already uses for `accumulationSurface` a few lines below.

### Repro
Linux (Ubuntu 24.04) + BGFX, `mm` + the Medieval Madness PUP pack. The PUP plugin showing its ancillary windows at game start invalidates the prepass → re-entry → 100% reproducible SIGSEGV:

```
Thread 25 "RenderThread" received signal SIGSEGV
#0 RenderTarget::CopyTo(...)
#1 RenderCommand::Execute(int, bool)
#2 RenderPass::Execute(bool)
#3 RenderFrame::Execute(bool)
#4 RenderDevice::SubmitRenderFrame()
```

Faulting instruction reads `this->m_rd` (offset 0x30) from the freed source RT and gets reused-heap garbage (`0x1e`), then derefs it.

### Verification
Same tree, only this change: crash gone, table + PUP pack (backglass/topper/scoreview media, triggers) run normally.

### DIFF
--- a/src/renderer/Renderer.cpp
+++ b/src/renderer/Renderer.cpp
@@ -1705,7 +1705,11 @@ void Renderer::RenderStaticPrepass()
    const bool isNoBackdrop = m_noBackdrop || ((m_render_mask & Renderer::REFLECTION_PASS) != 0);
 
    // The code will fail if the static render target is MSAA (the copy operation we are performing is not allowed)
-   delete m_staticPrepassRT;
+   // Defer deletion to the render thread end-of-frame: already submitted frames may still hold copy
+   // commands sourcing this RT (e.g. the static prepass flush blit), and a synchronous delete here
+   // races their execution (dangling source -> crash in RenderTarget::CopyTo).
+   if (m_staticPrepassRT)
+      m_renderDevice->AddEndOfFrameCmd([rt = m_staticPrepassRT]() { delete rt; });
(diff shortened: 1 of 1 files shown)


vbousquet:
Could you be more precise on the initial issue this PR fixes. The PR looks fine but I don't get why a pup pack would ever invalidate the static prerendering (and this looks like another bug if it is happening). Using the issue template could be a good thing when proposing a fix as the issue you are solving was never reported

dynajoe:
Good call — instrumented it, and you're right: the pack doesn't invalidate the prepass. The re-run is the by-design re-enable at the end of `Player::Player()`'s quick-render bracket; the quick frame already submitted to the render thread references the RT the re-run then deletes. A PuP pack's startup load just delays the render thread so the race fires every time. Full analysis + repro filed as #3637.

PR #3653, opened 2026-07-12
B2S: fix null-pointer dereference on .directb2s files without optional image elements

### LINKED ISSUE #3652: B2S plugin: null-pointer dereference loading any .directb2s without BackglassOnImage/BackglassOffImage/DMDImage
**Describe the bug**

The B2S plugin segfaults (null pointer dereference) while loading any `.directb2s` that does not contain **all five** of the image elements it reads. VPX dies instantly — no error, no log line, just a crash — for every table with such a backglass.

`plugins/b2s/B2SDataModel.cpp` reads five images:

https://github.com/vpinball/vpinball/blob/master/plugins/b2s/B2SDataModel.cpp#L359-L363

```cpp
   , m_thumbnailImage(GetImageAttribute(root, "Images/ThumbnailImage"s))
   , m_backglassImage(GetImageAttribute(root, "Images/BackglassImage"s))
   , m_backglassOnImage(GetImageAttribute(root, "Images/BackglassOnImage"s))
   , m_backglassOffImage(GetImageAttribute(root, "Images/BackglassOffImage"s))
   , m_dmdImage(GetImageAttribute(root, "Images/DMDImage"s))
```

but `GetImageAttribute` is the only `Get*` helper in that file that does **not** null-check the node it just looked up:

```cpp
static B2SImage GetImageAttribute(const tinyxml2::XMLNode& doc, const std::string& nodePath) noexcept
{
   const tinyxml2::XMLElement* node = GetNode(doc, nodePath);
   //return node ? B2SImage(*node) : B2SImage();   // <-- correct handling, commented out
   return B2SImage(*node);                          // <-- derefs nullptr when the element is absent
}
```

Every sibling helper (`GetStringAttribute`, `GetIntAttribute`, `GetColorAttribute`, `GetTextureAttribute`, `GetSoundAttribute`, `GetList`, `GetPtrList`, `GetFilteredList`) returns a default when `GetNode` returns `nullptr`. Only this one dereferences it.

A backglass with a single `<BackglassImage>` and no `<BackglassOnImage>` / `<BackglassOffImage>` / `<DMDImage>` is perfectly normal and common (it is what the B2S editor produces for non-illuminated backglasses), and such files load fine under the Windows B2S Server. On the Linux plugin they are fatal.

Note the deref happens through a *reference* parameter (`const XMLNode& doc`), so the compiler elides the null test inside `GetNode` (`const XMLNode* node = &doc;` is assumed non-null) and the crash surfaces one frame deeper, inside `B2S::GetNode`, which makes it look like an XML-parsing fault rather than a missing null check.

**To Reproduce**

1. Take any table whose `.directb2s` contains `<Images>` with `ThumbnailImage` and `BackglassImage` but no `BackglassOnImage`, `BackglassOffImage` or `DMDImage` (e.g. the widely distributed *Attack from Mars (Bally 1995)* backglass).
2. Enable the B2S plugin (`[Plugin.B2S] Enable = 1`).
3. `VPinballX_BGFX -play "<table>.vpx"`
4. VPX segfaults during table load.

Adding empty `<BackglassOnImage/><BackglassOffImage/><DMDImage/>` elements under `<Images>` in the same file makes the crash disappear and the table loads normally — confirming the missing elements are the trigger and that `B2SImage`'s attribute reads are already default-safe.

**Expected behavior**

Absent optional image elements should yield an empty `B2SImage` (as the commented-out line intends), not a crash.

**Crash dump**

Linux build, so no `crash.dmp`. Kernel log:

```
Main[13890]: segfault at 0 ip 00007f96043de021 sp 00007f954cf215c0 error 4 in plugin-b2s.so[21021,7f96043cd000+27000]
```

Disassembly at the faulting address (`plugin-b2s.so+0x21021`), inside `B2S::GetNode(tinyxml2::XMLNode const&, std::string const&)`:

```
   21015:  cmpq   $0x0,0x38(%rsp)
   2101b:  jne    21120 <_ZN3B2SL7GetNodeERKN8tinyxml27XMLNodeERKNSt7__cxx1112basic_string...>
   21021:  mov    (%r12),%rax          <-- r12 == 0  (segfault at 0)
   21025:  mov    %r12,%rdi
   21028:  call   *0x30(%rax)          <-- virtual call through null vtable
```

**Versions & Settings**

- VPX: 10.8.1 Beta, Rev. 5317 (408e32c27), linux BGFX 64-bit — also reproduces on Rev. 3788 (2151290)
- OS: Linux (Batocera 43 / kernel 6.18), x86-64
- GPU: NVIDIA
- Reproduces with default settings other than `[Plugin.B2S] Enable = 1`.

**Additional context**

I have a fix and will open a PR referencing this issue: restore the author's own null check and give `B2SImage` a default constructor (it currently has none, which is likely why the check was commented out in the first place).

Optional hardening: `GetNode`'s signature takes `const XMLNode&`, so any caller that passes a dereferenced null pointer gets its null check optimised away. Taking a `const XMLNode*` would make the whole file null-safe by construction.

dynajoe:
Fixes #3652

`GetImageAttribute()` in `plugins/b2s/B2SDataModel.cpp` is the only `Get*` helper in that file that does not null-check the element returned by `GetNode()`:

```cpp
static B2SImage GetImageAttribute(const tinyxml2::XMLNode& doc, const std::string& nodePath) noexcept
{
   const tinyxml2::XMLElement* node = GetNode(doc, nodePath);
   //return node ? B2SImage(*node) : B2SImage();
   return B2SImage(*node);
}
```

It is used for five image elements, three of which are optional (`Images/BackglassOnImage`, `Images/BackglassOffImage`, `Images/DMDImage`). A `.directb2s` with only a plain `<BackglassImage>` — which is what the B2S editor produces for non-illuminated backglasses, and which loads fine under the Windows B2S Server — therefore segfaults VPX during table load.

Because the deref is passed through a reference parameter, the compiler elides the null test inside `GetNode`, so the crash surfaces one frame deeper and looks like an XML-parsing fault:

```
segfault at 0 ... in plugin-b2s.so[21021,...]
   21021:  mov    (%r12),%rax     ; r12 == 0
   21028:  call   *0x30(%rax)     ; virtual call through a null vtable
```

**The fix** simply restores the null check that is already sitting there commented out, and adds the default constructor `B2SImage` was missing — its `const` members are almost certainly why the check could not be uncommented as-is.

Verified on Linux (BGFX, rev 5317 and rev 3788): the six tables in my library whose backglasses lack these elements crashed 100% of the time before, and all load and play correctly after. Tables whose backglasses do contain them are unaffected.

Also worth considering (not included here, to keep the fix minimal): `GetNode()` takes `const XMLNode&`, so any caller that dereferences a null pointer gets its internal null check optimised away. Changing the signature to `const XMLNode*` would make the file null-safe by construction.


vbousquet:
This is now fixed.

PR #3671, opened 2026-07-15
plugins: ignore game start with empty gameId (fixes DOF; defensive guard for altsound/serum/vni)

### LINKED ISSUE #3670: Plugins (altsound, serum, vni, dof) accept empty-gameId game start and lock out the real ROM
**Describe the bug**

An empty `gameId` reaches the game-start handlers of the controller plugins. A B2S-backed ROM table broadcasts `CTLPI_EVT_ON_GAME_START` **before the ROM name is committed**, so the first event carries an empty `gameId`.

**DOF is actively broken by this**: it sets `isRunning` unconditionally on the empty event, so the real ROM-named start ~18 ms later is dropped as "already running", and DOF holds the output controller with no per-ROM mappings — zero solenoids fire for the whole session.

PUP already guards against exactly this (`Ignoring game start with empty gameId`); DOF and the other game-start consumers (altsound, serum, vni) do not.

**To Reproduce**

1. Standalone build, a WPC/PinMAME table with a `.directb2s` (observed: Medieval Madness, `mm_109c`).
2. Launch the table.
3. DOF opens the controller and its updater thread connects, but no solenoids fire.

**Log evidence (identical on every launch):**
```
[PUP:PUP::OnControllerGameStart] Ignoring game start with empty gameId
[DOF:DOFPlugin::OnControllerGameStart] OnControllerGameStart: gameId=
[Serum:Serum::OnControllerGameStart] No colorization file found for mm_109c
[DOF:DOFPlugin::OnControllerGameStart] Ignoring game start, already running
...
[DOF] Cant load config from directoutput.ini or ledcontrol.ini file(s) since no RomName was supplied. No ledcontrol config will be loaded.
[DOF] Table config loading finished: romname=, tablename=1455577933_MedievalMadness_Upgrade(Real_Final)
[DOF] LedWizEquivalent 11: Resolved 0 of 63 outputs
```
Note `PUP` ignores the empty id, `DOF` accepts `gameId=` (empty) and then discards the real `mm_109c`. `Serum` logs the real `mm_109c` — it is *not* poisoned (see below).

**Expected behavior**

The empty-`gameId` event is ignored (as PUP already does), and the real ROM-named start initializes DOF — solenoids fire.

**Screenshots** — N/A (log-level).

**Crash dump** — N/A (not a crash).

**Versions & Settings**
 - VPX: standalone build (BGFX), Linux x64, current `master`; the affected code is cross-platform plugin code.
 - OS: Linux (Batocera).
 - GPU/video settings: not applicable.

**Additional context**

Scope of the affected plugins differs, and the fix PR is explicit about it:
- **DOF** — confirmed bug above (unconditional `isRunning`).
- **altsound / serum / vni** — *not* observably broken: they commit `isRunning` only after matching an asset file, so the empty event returns early and the real start still processes (the `Serum ... mm_109c` line above confirms this). They still build filesystem paths from the empty id, and altsound's empty path can resolve to the `altsound/` directory itself — so guarding them is defensive hardening, not a bug fix.
- **dmdutil** is intentionally out of scope: its `isRunning` is driven by DMD-source detection, not the `gameId` broadcast.

Related wart (not addressed here): because of the same `isRunning` guard, a second start with a *different, non-empty* gameId is also discarded rather than restarting the plugin.

dynajoe:
> **⚠️ AI-generated fix — draft per [CONTRIBUTING.md](https://github.com/vpinball/vpinball/blob/master/CONTRIBUTING.md#ai-generated-fixes-for-identified-issues).**
> The issue was identified and reproduced on real hardware; the code change was AI-generated and is offered as a **starting point / reference**, not a finished, human-validated contribution.

Fixes #3670.

### Summary

An empty `gameId` is never valid input to a game-start handler, but a B2S-backed ROM table broadcasts `CTLPI_EVT_ON_GAME_START` before the ROM name is committed, so the first event carries one. This adds the guard **PUP already has** to the other four game-start consumers.

### Two tiers — stated honestly

- **DOF (bug fix).** DOF sets `isRunning` unconditionally on the empty event, so the real ROM-named start ~18 ms later is dropped as "already running" and the session runs with no ROM: `0 of 63 outputs resolved`, no solenoids. Reproduced on Medieval Madness (`mm_109c`) — see #3670 for the full log.
- **altsound / serum / vni (defensive, not a bug fix).** These commit `isRunning` only after matching an asset file, so today the empty event returns early and the real start still processes (the `Serum … mm_109c` line in #3670 confirms Serum self-recovers). They are guarded anyway because an empty `gameId` should never reach path-building at all — it produces pointless filesystem lookups, and altsound's empty path can resolve to the `altsound/` directory itself. This is hardening for consistency with PUP, **not** a claim that they are currently broken.

`dmdutil` is intentionally excluded: its `isRunning` is DMD-source-driven, not `gameId`-driven.

### Technical overview

Immediately after each handler's existing `assert(msg->gameId != nullptr)`, add PUP's exact early-out:

```cpp
// The gameId can be an empty string on an early game-start broadcast.
if (msg->gameId[0] == '\0')
{
   LOGW("Ignoring game start with empty gameId"s);
   return;
}
```

Same wording and log line as PUP, for consistency. No behavioral change on the normal (non-empty) path. No files added/removed/renamed → no CMake/`.vcxproj` changes.

### Testing

- **DOF**: this exact guard has run on a real Linux (Batocera) standalone BGFX build and fixed the dead-solenoid regression on `mm_109c` — the real ROM start is accepted and DOF loads the table config.
- **altsound / serum / vni**: **not independently exercised.** Applied as defensive guards; their correctness is argued from the code (early-return on empty is a no-op relative to today's behavior), not from a reproduced scenario. Flagged honestly — part of why this is a draft.

### Discussion

The alternative long-term fix is to stop broadcasting the empty-`gameId` start at the source so no consumer needs the guard (PUP, and now these, could then drop it). This PR takes the consistent-with-PUP consumer-side path.

### DIFF
--- a/plugins/altsound/AltSoundPlugin.cpp
+++ b/plugins/altsound/AltSoundPlugin.cpp
@@ -189,6 +189,13 @@ static void OnControllerGameStart(const unsigned int eventId, void* userData, vo
     const CtlOnGameStartMsg* msg = static_cast<const CtlOnGameStartMsg*>(msgData);
     assert(msg != nullptr && msg->gameId != nullptr);
 
+    // The gameId can be an empty string on an early game-start broadcast.
+    if (msg->gameId[0] == '\0')
+    {
+       LOGW("Ignoring game start with empty gameId"s);
+       return;
+    }
+
     VPXTableInfo tableInfo;
     vpxApi->GetTableInfo(&tableInfo);
     std::filesystem::path tablePath = tableInfo.path;
--- a/plugins/dof/DOFPlugin.cpp
+++ b/plugins/dof/DOFPlugin.cpp
@@ -211,6 +211,13 @@ static void OnControllerGameStart(const unsigned int eventId, void* userData, vo
    const CtlOnGameStartMsg* msg = static_cast<const CtlOnGameStartMsg*>(msgData);
    assert(msg != nullptr && msg->gameId != nullptr);
 
+   // The gameId can be an empty string on an early game-start broadcast.
+   if (msg->gameId[0] == '\0')
+   {
+      LOGW("Ignoring game start with empty gameId"s);
+      return;
+   }
+
    // FIXME: Temp fix for issues 3298, 3309, and maybe 3322?
    if (isRunning)
    {
--- a/plugins/serum/serum.cpp
+++ b/plugins/serum/serum.cpp
@@ -335,6 +335,13 @@ static void OnControllerGameStart(const unsigned int eventId, void* userData, vo
    const CtlOnGameStartMsg* msg = static_cast<const CtlOnGameStartMsg*>(msgData);
    assert(msg != nullptr && msg->gameId != nullptr);
 
+   // The gameId can be an empty string on an early game-start broadcast.
+   if (msg->gameId[0] == '\0')
+   {
+      LOGW("Ignoring game start with empty gameId"s);
+      return;
+   }
+
    VPXTableInfo tableInfo;
    vpxApi->GetTableInfo(&tableInfo);
    std::filesystem::path tablePath = tableInfo.path;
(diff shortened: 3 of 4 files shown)


vbousquet:
The way running game was broadcasted was just a quick preliminary impelmentation. With the latest commits, it's somewhat better (still needs some more API design)

PR #3674, opened 2026-07-15
Fix #3666 - Plumb Damping setting displaying half value

### LINKED ISSUE #3666: Plumb Damping setting displays half value in ingameui
**Describe the bug**
Plumb Damping setting in ingameui displays a different value than what's listed in the INI file.

**To Reproduce**
1. Set `PlumbDamping = 1.0` (or any other value) in `VPinballX.ini`
2. Go to Nudge Settings ingameui
3. See that `Plumb Damping` displays 0.5, half of the real value
4. Save settings
5. See that `PlumbDamping = 0.5` in `VPinballX.ini`

The net effect is that every time you save the Plumb Damping with ingameui the value gets cut in half when written back to the INI file.

**Expected behavior**
INI value and ingameui value should match.

**Screenshots**

Image

**Versions & Settings**
 - VPX: Latest `master` [ca80d16](https://github.com/vpinball/vpinball/commit/ca80d1604e6a525db2a3f6d1ec704fc651d573fa)

[VPinballX.zip](https://github.com/user-attachments/files/29998253/VPinballX.zip)

cwick:
Fixes #3666

Getters and setters are now logically consistent with the constructor so the Plumb Damping settings UI works properly:
* Fix `GetPlumbDamping` to divide by the same coefficient the constructor uses
* Fix `SetPlumbDamping` to multiply by the same coefficient the constructor uses

### DIFF
--- a/src/physics/cabinet/PlumbHandler.cpp
+++ b/src/physics/cabinet/PlumbHandler.cpp
@@ -28,14 +28,13 @@ PlumbHandler::~PlumbHandler()
 
 float PlumbHandler::GetPlumbDamping() const
 {
-   return m_plumbAngularDamping0 / 2.5f;
-   //return m_plumbAngularDamping1 / 1.5f;
+   return m_plumbAngularDamping0 / m_dampingCoef0;
 }
 
 void PlumbHandler::SetPlumbDamping(float v)
 {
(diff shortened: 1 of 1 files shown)


vbousquet:
Thanks!

Issue #3675, opened 2026-07-15
dmdutil: per-score-event texture uploads stutter the playfield on high-refresh cabs (GPU queue contention)


dynajoe:
**Describe the bug**

On a 120 Hz cabinet, the playfield stutters in a way that tracks *gameplay* — worst during flipper volleys. Root cause: PinMAME emits a DMD frame on **every score event**, so scoring bursts produce DMD texture-upload rates far above any real DMD's refresh. Each upload contends with the playfield render on the same GPU queue.

This is a performance/quality issue, not a crash. Filing because the fix is a **behavioral change** (frame coalescing) that deserves maintainer input on the right approach — see the linked draft PR.

**To Reproduce**

1. Standalone build, a PinMAME/DMD table, DMDUtil enabled, on a high-refresh playfield (120 Hz).
2. Enable the VPX FPS overlay.
3. Play — keep a ball in play and fire the flippers to generate score events.
4. Watch the playfield fps dip on scoring activity; the DMD's `BGFX->GPU` submission time spikes on event frames.

**Measurements** (RTX 3080, 4K @ 119.88 Hz, SyncMode 0, cap 119.88; one plugin isolated at a time, simulated gameplay):

| tracker | PUP | B2S | DMDUtil | playfield |
|---|---|---|---|---|
| off | on | on | **on** | dips to 111.8 fps |
| on | off | on | **on** | dips to 111.7 fps |
| on | off | off | **on** | dips to 103.8 fps |
| off | off | off | **off** | clean 119.9 ×3 |
| on | on | on | **on + rate limit** | **clean 119.9 ×3** |

Dip signature: `BGFX->GPU` submission time inflates 0.1 ms → 0.9–1.8 ms on event frames — queue contention, not render cost.

**Expected behavior**

DMD updates should not drive the playfield below its refresh target. A DMD does not need faster-than-hardware upload rates; bursts can be coalesced to the newest frame.

**Screenshots** — N/A (fps overlay numbers above).

**Crash dump** — N/A (not a crash).

**Versions & Settings**
 - VPX: standalone build (BGFX/Vulkan), Linux x64, current `master`.
 - OS: Linux (Batocera). GPU: NVIDIA RTX 3080. 3-screen cabinet, 4K playfield @ 119.88 Hz.

**Additional context**

[...]

Issue #3672, opened 2026-07-15
dmdutil: UpdateThread dereferences null frame data — segfault takes down the player


dynajoe:
**Describe the bug**

`DMDUtilPlugin::UpdateThread` dereferences `frame.frame` without a null check, taking the whole player down. `GetRenderFrame()` can return a frame whose data pointer is null — no frame produced yet, or the source torn down between the guards above and the call — and all three format branches dereference it immediately.

**To Reproduce**

Runs on its own on a cabinet: `VPinballX_BGFX` segfaults several times a day during normal play, always a null read at the same instruction. It is a teardown/startup race on the DMD source, not tied to a specific table or input.

**Crash evidence** (six crashes, identical offset in `plugin-dmdutil.so`):
```
VPinballX_BGFX[13073]: segfault at 0 ip ... in plugin-dmdutil.so[7750,...]
VPinballX_BGFX[16263]: segfault at 0 ip ... in plugin-dmdutil.so[7750,...]
VPinballX_BGFX[24724]: segfault at 0 ip ... in plugin-dmdutil.so[7750,...]
```
Disassembling `UpdateThread` at that offset:
```asm
   7747:  cvtsi2ss %edx,%xmm2          ; tintB
   774b:  mov    %r14,%rdx
   7750:  movss  (%rdx),%xmm0          ; <-- %rdx = 0, the LUM32F luminanceData[i] load
```

**Cause**
```cpp
const DisplayFrame frame = selectedDmdId.GetRenderFrame(selectedDmdId.id);
if (lastFrameID == frame.frameId)
   continue;
lastFrameID = frame.frameId;

switch (selectedDmdId.frameFormat) {
   case CTLPI_DISPLAY_FORMAT_LUM32F: {
      const float* const __restrict luminanceData = static_cast<const float*>(frame.frame);
      ...
      const float lum = luminanceData[i];    // <-- null deref
```
`LUM32F` reads `frame.frame` directly; `SRGB888`/`SRGB565` hand it straight to `UpdateRGB24Data`/`UpdateRGB16Data`. The code immediately above **already guards** `pDmd`, `selectedDmdId.id.id` and `GetRenderFrame` against precisely this teardown race (with a comment noting the source "can be reset ... concurrently with this thread"); the frame **data** pointer was missed.

**Expected behavior**

A null frame is skipped rather than dereferenced; the player keeps running.

**Screenshots** — N/A.

[...]

PR #3676, opened 2026-07-15
dmdutil: rate-limit DMD texture uploads to stop playfield stutter (proposed approach, needs review)

dynajoe:
> **⚠️ AI-generated change — draft per [CONTRIBUTING.md](https://github.com/vpinball/vpinball/blob/master/CONTRIBUTING.md#ai-generated-fixes-for-identified-issues).**
> This is a **take-it-or-leave-it proposed approach**, not a finished contribution. The goal was to reduce a real, measured playfield stutter (see #3675). The maintainers may well have a cleaner way to get the same result — that is genuinely expected here, and this PR is meant as a reference/starting point.

Addresses #3675.

### What this is trying to fix

On a 120 Hz cabinet the playfield stutters in a way that tracks gameplay. PinMAME emits a DMD frame on every score event, so flipper volleys burst DMD texture uploads far above any real DMD's refresh, and each upload contends with the playfield on the GPU queue. Measured: the 119.9 fps playfield dips to ~104–112 fps during scoring; disabling DMDUtil removes the dips entirely. Full table in #3675.

### The approach in this PR (one option)

Throttle uploads to 40 Hz — faster than any physical DMD — placed **after** the `frameId` dedup but **before** recording the id, so a rate-limited frame is skipped **without** being marked seen. The newest frame therefore lands on the next tick: bursts are **coalesced, never lost**, and the DMD stays current.

```cpp
{
   static std::chrono::steady_clock::time_point lastUpload{};
   const auto now = std::chrono::steady_clock::now();
   if (now - lastUpload < std::chrono::milliseconds(25))
      continue;
   lastUpload = now;
}
```

### Why it needs your scrutiny — open questions

- **Right layer?** This throttles on the consumer (plugin) side. Coalescing at the source (PinMAME emit) or a latest-wins single-slot buffer might be architecturally cleaner and avoid a fixed poll cadence.
- **Right cadence?** 40 Hz / 25 ms is a chosen constant. Should it be configurable, or derived from something?
- **`static` in `UpdateThread`.** Fine for the single update thread today, but it bakes in that assumption; a member on the plugin state would be more honest if that ever changes.
- **Interaction with real DMD content** (mode transitions, animations): 40 Hz should be imperceptible, but a maintainer who knows the DMD pipeline may see a case I don't.

### Testing

Ran on a real Linux (Batocera) standalone BGFX/Vulkan build, RTX 3080, 4K playfield @ 119.88 Hz: with the limiter the playfield holds a clean 119.9 fps through flipper volleys that previously dipped to ~104 fps (#3675 table, last row). The DMD looked unchanged in normal play. This is empirical on one rig, by eye for the DMD content — not a rigorous frame-accuracy check, which is part of why it is a draft.

### Note

Independent of the null-frame crash fix (#3672 / its PR), though both touch `UpdateThread` and one will need a trivial rebase after the other merges.

### DIFF
--- a/plugins/dmdutil/DMDUtilPlugin.cpp
+++ b/plugins/dmdutil/DMDUtilPlugin.cpp
@@ -120,6 +120,23 @@ static void UpdateThread()
       const DisplayFrame frame = selectedDmdId.GetRenderFrame(selectedDmdId.id);
       if (lastFrameID == frame.frameId)
          continue;
+
+      // Rate-limit uploads. PinMAME emits DMD frames on every score event — a flipper
+      // volley produces bursts well above any real DMD's refresh — and each one is a
+      // texture upload contending with the playfield on the same GPU queue. Profiled on
+      // a cab: with this plugin enabled the 119.9fps playfield dips to ~104-112 during
+      // flipper events (BGFX->GPU submission 0.1ms -> 0.9-1.8ms); with it disabled the
+      // dips vanish entirely. 40Hz is faster than the hardware ever was and caps the
+      // burst. The frameId is NOT recorded on a skipped frame, so the newest frame
+      // still lands on the next tick — nothing is lost, only coalesced.
+      {
+         static std::chrono::steady_clock::time_point lastUpload{};
+         const auto now = std::chrono::steady_clock::now();
+         if (now - lastUpload < std::chrono::milliseconds(25))
+            continue;
+         lastUpload = now;
+      }
+
       lastFrameID = frame.frameId;
 
       switch(selectedDmdId.frameFormat) {


francisdb:
#3469 avoids dmd render on the game render thread, probably a better fix

toxieainc:
i also added some checks to only upload/update display data in the plugins if something changed.

toxieainc:
@dynajoe Did you test again with recent master? And/or #3469?

vbousquet:
There is an issue with the approach of the proposed fix: it limits all DMDs to 1/25ms that is to say 40FPS. It may be good for some displays but also wrong for others (creating stutters as the FPS will not match the device refresh rate). Ideally, DMDUtil should know and feed at the connected device refresh rate. If this not known by DMDUtil then this should be a user option.

PR #3699, opened 2026-07-22
VR: fix DMD not appearing in backglass

### LINKED ISSUE #3698: DMD not visible in VR backglass when PlayerVR.AddBackglass=1
**Describe the bug**
DMD fails to display inside of the VR backglass when `PlayerVR.AddBackglass=1`

**To Reproduce**
1. Enable `Plugin.B2S` and set `Plugin.B2S.BackglassDMDOverlay=1`
1. Set `Plugin.B2S.BackglassDMDAutoPos=1` or manually position DMD 
1. Pick any table that doesn't have a built-in VR backglass (e.g. `Attack from Mars (Bally 1995) g5k 1.3.11`)
1. Add a well-formed `.directb2s` file next to the table that contains a backglass image
1. Start the table in 2D mode and see that the DMD renders properly inside of the backglass
1. Set `PlayerVR.AddBackglass=1`
1. Start the table in VR mode and see that the DMD does not render in the backglass

**Expected behavior**
The DMD should render inside of the backglass, matching the 2D case.

**Screenshots**

Missing DMD display

Image

It should look like this

Image

**Versions & Settings**
BGFX, latest `master`

**Additional context**
See attached PR for root cause and fix.

cwick:
Disclaimer: Based on my own basic knowledge of 3D transforms supplemented by Claude / Codex

Fixes #3698

`Renderer::DrawMatrixDisplay` passed local normalized vertex coordinates to `DrawTexturedQuad`. This worked in the 2d case because the shader was set up with a matrix that maps normalized coordinates directly to window coordinates.

It also worked in VR in the pre-BGFX era because the legacy DMD shader (`src\shaders\hlsl_glsl\DMDShader.hlsl`) applied the backglass model transform itself: `mul(vPosition, matWorldViewProj);`

But in the era of BGFX the DMD shader (`src\shaders\bgfx\vs_dmd.sc`) does not apply any model transform. It expects vertices in world coordinates. So when `Renderer::DrawMatrixDisplay` passes local normalized coordinates to the new BGFX shader and we're running in VR mode, no backglass model transform is applied and the DMD incorrectly renders somewhere near the world (0,0,0) origin instead of on the backglass.

Appears to be a regression since it appears to have worked before BGFX. Note: limited to tables that don't have their own built in backglass/dmd and the user has selected `PlayerVR.AddBackglass`

The fix is to pre-transform vertices into world space inside of `DrawMatrixDisplay`, gated on VR mode and `ENABLE_BGFX`. `DrawSegmentDisplay` follows the same pattern so I fixed it there too.

### DIFF
--- a/src/renderer/Renderer.cpp
+++ b/src/renderer/Renderer.cpp
@@ -3155,11 +3155,17 @@ void Renderer::DrawMatrixDisplay(VPXRenderContext2D* ctx, VPXDisplayRenderStyle 
    const float vy1 = 1.f - srcY / ctx->srcHeight;
    const float vx2 = (srcX + srcW) / ctx->srcWidth;
    const float vy2 = 1.f - (srcY + srcH) / ctx->srcHeight;
-   const Vertex3D_NoTex2 vertices[4] = { //
+   Vertex3D_NoTex2 vertices[4] = { //
       { vx2, vy1, 0.f, 0.f, 0.f, 1.f, 1.f, 1.f }, // 
       { vx1, vy1, 0.f, 0.f, 0.f, 1.f, 0.f, 1.f }, //
       { vx2, vy2, 0.f, 0.f, 0.f, 1.f, 1.f, 0.f },  //
       { vx1, vy2, 0.f, 0.f, 0.f, 1.f, 0.f, 0.f } };
+#if defined(ENABLE_BGFX)
+   if (!ctx->is2D)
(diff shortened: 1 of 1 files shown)


vbousquet:
The right fix will be to update the shader. Now that everything is merged, we are trying to remove all the hacky #ifdef we have (and here, we need to apply the transform per eye). I will close for now, and will update the shader when I get some time off

PR #3700, opened 2026-07-24
Input: Fix broken OpenPinDev

### LINKED ISSUE #3701: OpenPinDev is non-functional
**To Reproduce**
1. Plug in an OpenPinball device (e.g. Pinscape Pico)
2. Start any VPX table
3. See that nothing happens

**Expected behavior**
Device should be detected, it should offer to apply default mappings, and those default mappings should be correct.

**Additional context**
Fixed in #3700

cwick:
Fixes #3701

Plugged in a Pinscape Pico only to discover that OpenPinDev has somehow been broken since https://github.com/vpinball/vpinball/commit/03e457fcedfba8c1d4cdf5b423d84de0e8bd0dbd (August 2025)

This PR restores OpenPinDev to a functional state:
* Fix incorrect string comparison causing OpenPinball devices to not be recognized at all (should be prefix matching, not strict equality)
* Fix incorrect OpenPinDev buttons in the default mapping
* Fix incorrect UI labels
* Fix off-by-one error in the named button handler


mjrgh:
That all looks good to me.

PR #3714, opened 2026-07-29
VR: Center table relative to player

### LINKED ISSUE #3713: VRCenter action doesn't really center the table
**Describe the bug**
Pressing the `VRCenter` button should translate and rotate the scene such that the table is in front of you, but it doesn't do that.

**To Reproduce**
1. Load any table in VR mode
2. Press the `VRCenter` button
3. Instead of centering the table, it translates and rotates the scene in a seemingly random way, and it's very disorienting.

**Expected behavior**
Table should end up in front of you and rotated so the playfield is facing you.

**Versions & Settings**
 - Latest master

**Additional context**
See attached PR for proposed fix.

cwick:
Fixes #3713 

VR centering behavior changed in 8c6ff6cb9. I believe it is a regression.

The VR centering code didn't take player position and rotation into account so you end up facing a random direction after centering table. Claude fixed the math, I confirmed it works: After centering, table now ends up in front of you, offset by ScreenPlayerY.

### DIFF
--- a/src/renderer/VRDevice.cpp
+++ b/src/renderer/VRDevice.cpp
@@ -1103,9 +1103,19 @@ void VRDevice::RenderFrame(RenderDevice* rd, const std::function<void(RenderTarg
             XrPosef_ToMatrix3D(&medianView, &medianPoseInVPU);
 
             m_headsetViewCentering = false;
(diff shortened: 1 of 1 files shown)


vbousquet:
Thanks!

PR #3715, opened 2026-07-29
Timer: fix Defer() scheduling the timer in the past instead of the future

cwick:
`Defer` was scheduling the timer one second in the past instead of in the far future.

This code may have been mis-copied from 10.8.0:
https://github.com/vpinball/vpinball/blob/7cdbb55bf49c73774d152def3719bb0b95d7c49f/ieditable.cpp#L77
https://github.com/vpinball/vpinball/blob/7cdbb55bf49c73774d152def3719bb0b95d7c49f/src/parts/timer.cpp#L192

I don't think this breaks anything today but it's a bug just waiting to happen.

### DIFF
--- a/src/physics/hittimer.cpp
+++ b/src/physics/hittimer.cpp
@@ -20,8 +20,8 @@ void HitTimer::SetInterval(int intervalMs)
 
 void HitTimer::Defer()
 {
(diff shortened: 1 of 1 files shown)


vbousquet:
Thanks

PR #3731, opened 2026-07-30
fix: don't apply parallax in 2D mode

### LINKED ISSUE #3729: Ancillary DMD/segment displays render nothing after eb32f30c8 (parallax unit mismatch)
Since eb32f30c8 ("Enable display parallax effect for ancillary renderers"), DMD and segment displays on ancillary windows (backglass / score view / topper) don't render at all. Reproduced on Linux/KMSDRM with RENDERER=BGFX on a score view window; the cause looks platform-independent.

That commit replaced the explicit nullptr, // No parallax arguments in DrawMatrixDisplay and DrawSegmentDisplay with the quad's vertices, which enables this block in SetupDisplayRenderer:

if (!isBackdrop && (vertices != nullptr))
{
   const Matrix3D& mv = GetMVP().GetModelView(0);
   ...
   constexpr float depth = CMTOVPU(0.5f);   // 9.27 VP units
   parallaxU = (depth / tN) * tangent.Dot(eye) / tN;
}
m_DMDShader->SetVector(SHADER_glassPad, emitterPad.x - parallaxU, ...);
The math is world-space — it projects through the model-view of a 3D camera and offsets by a depth in VP units. But an ancillary quad is built in normalised space:

const float vx1 = srcX / ctx->srcWidth;   // 0..1, z = 0
so tangent/bitangent come out ≈ 1 while depth is ≈ 9.3. The resulting glassPad offset is several times the size of the display and pushes the emitter area off the glass, so nothing is drawn.

The !isBackdrop guard doesn't catch it: all three ancillary call sites pass isBackdrop = false, so the nullptr was the only thing disabling parallax there.

Suggested fix — the render context already distinguishes the two cases (flasher.cpp sets is2D from m_desktopBackdrop), so passing ctx->is2D ? nullptr : vertices at the three call sites keeps parallax for genuine 3D displays and disables it where it's undefined. A flat window viewed head-on has no view direction, so zero seems more correct than a rescaled offset.

cwick:
@vbousquet 
eb32f30c80b20fd9d340445931a52aac388575bf broke DMD ScoreView display in 2d windows.

It seems parallax should not be applied in 2D contexts. Should we pass `nullptr` for the `vertices` param when context is 2D or did you have something else in mind?

Fixes #3729

### DIFF
--- a/src/renderer/Renderer.cpp
+++ b/src/renderer/Renderer.cpp
@@ -3147,11 +3147,12 @@ void Renderer::DrawMatrixDisplay(VPXRenderContext2D* ctx, VPXDisplayRenderStyle 
       { vx2, vy2, 0.f, 0.f, 0.f, 1.f, 1.f, 0.f }, //
       { vx1, vy2, 0.f, 0.f, 0.f, 1.f, 0.f, 0.f }
    };
+   const Vertex3D_NoTex2* const parallaxVertices = ctx->is2D ? nullptr : vertices;
    if (style == VPXDMDStyle_Pixelated || style == VPXDMDStyle_Smoothed || style == VPXDMDStyle_CRT)
    {
       g_pplayer->m_renderer->SetupCRTRender(style - VPXDMDStyle_Pixelated, false, vec3(dispTintR, dispTintG, dispTintB), brightness, dTex, alpha, //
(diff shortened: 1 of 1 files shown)


vbousquet:
This is properly fixed in latest (with working parallax effect)

PR #3735, opened 2026-07-31
Fix deprecated VR backglass flasher stealing a part name

### LINKED ISSUE #3657: Time Machine (Data East 1988) script error due to pintable GetUniqueName
**Describe the bug**

Following up from [discord](https://discord.com/channels/652274650524418078/1076655472846831667/1521071757807587369), now that [this](https://gitlab.winehq.org/wine/wine/-/merge_requests/11278) wine fix has been [merged](https://github.com/vpinball/vpinball/commit/40123f20c9e012d048dc4a99f9d41db2d095042a) to vpinball, running [Time Machine (Data East 1988)](https://virtualpinballspreadsheet.github.io/games?game=pQsqGpS1) table [uzvM4j_Dru](https://vpuniverse.com/files/file/24756-time-machine-data-east-1988-v21/) logs error:
```
2026-07-13 13:59:55.984 ERROR [Main] [Player::OnScriptError@1238] Runtime error on line 4875, col 40: Object doesn't support this property or method
```
which now points correctly to the following script source:
```vbscript
Sub SetBackglassRefl(Opt)
	Select Case Opt
		Case 0:
			For each xx in BackglassReflections: xx.ReflectionEnabled = False : Next
			If VRMode = False Then
				For Each VR_Obj in BackglassReflections : VR_Obj.Visible = 0 : Next
			End If
		Case 1:
			For each xx in BackglassReflections: xx.ReflectionEnabled = True : Next ' <---- fails on this line
			For Each VR_Obj in BackglassReflections : VR_Obj.Visible = 1 : Next
	End Select
End Sub
```

Analysis of root cause from @francisdb:
> VPX's deprecated-VR-backglass hack (pintable.cpp:1738) names its auto-created flasher with GetUniqueName(primitive->GetWName()) mid-load, before all parts are registered - so on PinCab_Backglass it generates PinCab_Backglass001, which isn't actually free (the real primitive of that name isn't loaded yet), stealing the name.                                                                                                                                                                                                               
The table's BackglassReflections collection then resolves PinCab_Backglass001 to that flasher, and xx.ReflectionEnabled = True throws err 438 because flashers have no ReflectionEnabled.

**Versions & Settings**

 - VPX: 10.8.1-5319-40123f20c

gitfool:
> [!IMPORTANT]
> **This is an AI-generated fix, opened as a draft per [CONTRIBUTING.md → AI-Generated Fixes for Identified Issues](https://github.com/vpinball/vpinball/blob/master/CONTRIBUTING.md#ai-generated-fixes-for-identified-issues).**
>
> I identified the issue and reproduced it, but the code change itself was generated with AI assistance (Claude via Kiro). It is offered as a starting point / possible direction to help design the final solution, not as a finished, human-validated contribution. Please treat it as a reference rather than a review-ready patch. I am happy to iterate on it, or for a maintainer to take it over or reimplement it.

Fixes #3657

## Describe the bug

Loading *Time Machine (Data East 1988)* logs a script runtime error:

```
ERROR [Main] [Player::OnScriptError@1236] Runtime error on line 4870, col 40: Object doesn't support this property or method
```

The failing script line is in the table's own code:

```vb
Sub SetBackglassRefl(Opt)
    Select Case Opt
        Case 0:
            For each xx in BackglassReflections: xx.ReflectionEnabled = False : Next
            ...
        Case 1:
            For each xx in BackglassReflections: xx.ReflectionEnabled = True : Next   ' <-- fails here
            ...
```

### Root cause

`PinTable::LoadGameFromFilename` does two things in the same pass over the loaded parts:

1. Registers each part name via `AddPart` (renaming duplicates), and
2. Runs the deprecated-VR-backglass conversion: a primitive using the `backglassimage` texture is hidden and replaced by an external-render flasher, named with `GetUniqueName(primitive->GetWName())`.

`GetUniqueName` can only see names that are *already registered*. This table has both `PinCab_Backglass` and `PinCab_Backglass001`, and the former is processed first, so `PinCab_Backglass001` still looks free and the generated flasher takes it. When the real primitive is registered moments later it is treated as a duplicate and bumped. Both steps are visible in the log:

```
ERROR [Main] [PinTable::LoadGameFromFilename@1786] Primitive 'PinCab_Backglass' used as a deprecated VR backglass was hidden and an external renderer flasher named 'PinCab_Backglass001' was added. This may cause script issues.
ERROR [Main] [PinTable::LoadGameFromFilename@1686] Duplicate part name found: PinCab_Backglass001 renamed it to PinCab_Backglass001001
```

`Collection::InitPostLoad` resolves collection members by name *after* loading, so the table's `BackglassReflections` collection binds `PinCab_Backglass001` to the flasher instead of to the primitive. Flashers have no `ReflectionEnabled` property, hence error 438.

Credit to @francisdb, who diagnosed this in #3657; this PR implements that analysis.

## To Reproduce

1. Use a table that contains a primitive textured with `backglassimage` (the deprecated VPVR backglass hack) **and** another part whose name is the auto-generated candidate for it, i.e. `<primitiveName>001`. *Time Machine (Data East 1988)* has `PinCab_Backglass` + `PinCab_Backglass001`.
2. Launch the table.
3. Observe in the log: the generated flasher takes `PinCab_Backglass001`, the real primitive is renamed to `PinCab_Backglass001001`, and the script raises error 438 when touching `ReflectionEnabled` on a collection containing `PinCab_Backglass001`.

## Expected behavior

The auto-created flasher should never take a name belonging to a part in the table file. Existing part names must keep resolving to their original objects so table scripts and collections continue to work.

## Technical overview

Single change in `src/parts/pintable.cpp`: collect the candidate primitives during the registration pass, then perform the flasher conversion in a second pass once every part name is registered, so `GetUniqueName` sees the complete registry.

```cpp
vector<Primitive *> deprecatedVRBackglasses;
...
      // in the registration loop
      if (part->GetItemType() == eItemPrimitive && StrCompareNoCase(((Primitive *)part)->m_d.m_szImage, "backglassimage"s))
         deprecatedVRBackglasses.push_back((Primitive *)part);
...
// after the loop
for (Primitive *const primitive : deprecatedVRBackglasses)
{ /* conversion body, unchanged */ }
```

The conversion body itself is unchanged logic. It is only re-indented because it moved out of the loop, which inflates the raw line count; **`git diff -w` shows the real change as 12 added / 6 removed lines**. The parts remain valid in the second pass since `AddPart` takes a reference.

## Testing

* Built `Release_BGFX-x64` on Windows (MSVC), no new warnings or errors.
* Ran *Time Machine (Data East 1988)* before and after the change on the same machine and settings.

Before:

```
ERROR ... flasher named 'PinCab_Backglass001' was added
ERROR ... Duplicate part name found: PinCab_Backglass001 renamed it to PinCab_Backglass001001
ERROR [Main] [Player::OnScriptError@1236] Runtime error on line 4870, col 40: Object doesn't support this property or method
```

After:

```
ERROR ... flasher named 'PinCab_Backglass009' was added
INFO  [Main] [Player::Player@833] Startup done
```

No duplicate-rename, no script error, table starts and plays. The flasher now lands on a genuinely free name and `PinCab_Backglass001` resolves to the real primitive again. Remaining log entries for this table (duplicate sound `WireRamp_Stop`, duplicate image `flasher2`, deprecated PinMAME calls, plugin refcount warnings at shutdown) are pre-existing and unrelated.

Tested only on Windows / BGFX (DX12). Not tested on OpenGL ES or standalone, though the change is in platform-independent table loading code.

## Discussion / risks

* Behaviour is unchanged for tables where the generated name was already free; only the collision case differs.
* The `hasBackglassFlasher` pre-check still iterates the loaded `parts` list, so a table with two `backglassimage` primitives would still produce two flashers, exactly as before. I left that as-is to keep the change to one concern, but a maintainer may prefer checking the live part list instead.
* The generated name is now less predictable (`PinCab_Backglass009` rather than `...001`), since it depends on which names are already taken. That was already true in general, and the name is only reachable from script if a table happens to reference it.
* Broader question for maintainers: the conversion intentionally keeps the original primitive, so a table script tweaking the *primitive* has no effect on the *flasher*. This PR does not address that; it only stops the flasher from hijacking an existing name.

## Additional work (not included)

Two pre-existing bugs I noticed in the same block while reading it, deliberately left out to keep this PR to one change. Happy to open a separate PR:

* `maxx = max(miny, px);` looks like a copy/paste slip and should be `max(maxx, px)`, so the computed backglass width is wrong.
* `maxx` / `maxy` are seeded with `FLT_MIN`, the smallest *positive* float, where `-FLT_MAX` is intended.

Both affect the generated flasher's size/position rather than the name binding, so neither is involved in this issue.

## Versions & Settings

* VPX: `v10.8.1 Beta (Rev. 9999 (unknown), windows BGFX 64bits)`, local `Release_BGFX-x64` build of `master` plus this commit
* Windows 11, BGFX / DX12 backend

### DIFF
--- a/src/parts/pintable.cpp
+++ b/src/parts/pintable.cpp
@@ -1661,6 +1661,7 @@ HRESULT PinTable::LoadGameFromFilename(const std::filesystem::path &filename, VP
             };
 
             // Handle failed loading & duplicates
+            vector<Primitive *> deprecatedVRBackglasses;
             if (!parts.empty())
             {
                // Process unnamed parts after named parts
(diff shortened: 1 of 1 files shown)


vbousquet:
Thanks for the fix hint. I did a slightly cleaner version and merged

gitfool:
@vbousquet what did you think about the other pre-existing bugs it noticed?

vbousquet:
They are indeed bugs that I will fix in a separate commit.

At some point it would be interesting to have a systematic review of the codebase but this can lead to an overwhelming result and we already have the Sonar review which is not fully processed...

PR #3749, opened 2026-08-02
Support SAM palette changes in VNI colorization

djrobx:
## What changed

- Forward PinMAME console-data callbacks through the plugin message API.
- Parse SAM `Pxx` palette requests in the VNI plugin and select the corresponding PAL palette directly through `libvni`.
- Fall back to VPX's native rendered DMD source when no indexed identify source is available, preserving desktop textbox rendering.

## Why

SAM colorized ROMs send palette changes out-of-band over the serial console stream; the DMD frame itself does not contain the palette marker. The VNI plugin previously had no access to that stream, so it stayed grayscale or used the wrong palette.

## Dependency

This PR requires the companion `libvni` change adding `Vni_SetPalette`: https://github.com/PPUC/libvni/pull/9

## Validation

Validated with Metallica Premium Monsters (Stern 2013) on VPX 10.8.1: internal desktop DMD and the external DMDext display matched frame-for-frame, including SAM palette changes.


vbousquet:
This looks nice. I have updated libPinMAME to define PMPI_EVT_ON_CONSOLE_DATA in PinMAMEPlugin.h
Perhaps we should wait for libvni to be updated before merging this one.

As a side note, while improving libvni to put it in par with DMDExt (as its purpose is more or less the same in a modular & cross platform way), perhaps we should also and add encryption support ?

djrobx:
Agreed, don't merge this until the downstream VNI library takes the change.  

And yes, adding encryption/PAC support to that vni library would be the next thing I'd be interested in making work.   I was researching that situation last night.    Just testing the waters with a simpler change first, since I knew exactly what was missing in this case.   :)

vbousquet:
Just updated PinMAME again as it feels more clean that the event are broadcasted from libPinMAME like Audio events.
So in this PR, the PinMAMEPlugin change can be removed as I think they are useless now:
- Controller events were already broadcasted by libpinmame
- Console events are now broadcasted by libpinmame

Only the VNI stuff is still needed

plugins/vni/vni.cpp
```diff
@@ -89,15 +139,46 @@ static void ColorizeThread()
...
          break;
 
       if (frame.frameId != lastFrameId)
       {
          lastFrameId = frame.frameId;
-         const uint8_t bitlen = dmdId.identifyFormat == CTLPI_DISPLAY_ID_FORMAT_BITPLANE4 ? 4 : 2;
-         const uint32_t result = Vni_Colorize(pVni, static_cast<const uint8_t*>(frame.frame), dmdId.width, dmdId.height, bitlen);          
+         uint8_t bitlen = hasIdentifyFrame ? (dmdId.identifyFormat == CTLPI_DISPLAY_ID_FORMAT_BITPLANE4 ? 4 : 2) : 4;
+         const uint8_t* indexed = static_cast<const uint8_t*>(frame.frame);
+         if (!hasIdentifyFrame)
```
vbousquet: I would really avoid to do that. By definition, non identify frame are unstable and won't be bacwkard compatible. So this is creating something meant to break at a later point.

plugins/vni/vni.cpp
```diff
@@ -172,12 +253,20 @@ static void OnDmdSrcChanged(const unsigned int, void*, void*)
       return;
    getSrcMsg = { getSrcMsg.count, 0, new DisplaySrcId[getSrcMsg.count] };
    msgApi->BroadcastMsg(endpointId, getDmdSrcId, &getSrcMsg);
-   for (unsigned int i = 0; i < getSrcMsg.count; i++)
+   // Prefer a true indexed identify source. If none is available, fall back to
```
vbousquet: See my comment above. If the aim is to gather VPX's source as identify frame, then it's VPX that should be modified to advertise and provide it. Not here to hack it in

vbousquet (changes requested):
Building blindly an identify frame is not clean. If this frame is missing, it must be provided by the provider that is responsible for not here as it would create bugs later on.

djrobx:
Nice.  Updated per comments, verified working.

djrobx:
Note that I've also added a working PAC implementation now 

https://github.com/PPUC/libvni/pull/10

If that's accepted, I'll raise the corresponding PR to implement it here, but I have it working locally.

djrobx:
Per discussion with Markus we will not include direct PAC support.  Users can use an external utility to decrypt the PAC files and use them that way.   The SAM change was merged so this should be safe to include now. 

One other small suggestion - we might rename this plugin to something like "pal-vni".   Pinball browser SAM colorizations and basic single-palette PAL file support have nothing to do with VNI at all, but the plugin works well for both.

jsm174:
Can you bump the sha in platforms/config.sh?

djrobx:
> Can you bump the sha in platforms/config.sh?

Have opened libdmdutil PR: https://github.com/vpinball/libdmdutil/pull/121
Once that's merged, I will update dmdutil's sha.

jsm174:
Done!

PR #3761, opened 2026-08-05
BGFX: reduce per-draw CPU cost in command submission (snapshot-direct uniforms + unchanged-value skip)

dynajoe:
**Disclosure: this change was authored by an AI assistant (Claude), per CONTRIBUTING's AI policy it is opened as a draft.** It was designed, reviewed and measured on real hardware as described below; please review with that in mind.

Two commits reducing render-thread CPU in the BGFX submit path:

1. **Apply uniforms from the command snapshot instead of copying it per draw.** `RenderCommand::Execute` copied the command's ShaderState over `m_state` (full uniform block + a shared_ptr copy per bound sampler) on every draw, only so `Begin()` could read it back. `Begin(commandState)` now applies straight from the snapshot (BGFX only — GL/DX9 keep the copy since their inline execution may rely on `m_state` reflecting the last draw). `GetCore()` resolves the technique from the active state.

2. **Skip re-recording uniforms whose value has not changed.** The memcmp skip in `ApplyUniform` is compiled out for BGFX, so every draw re-recorded every uniform of its technique. A global cache keyed by bgfx handle idx skips identical bytes. Correctness rests on: (a) every VPX view uses `ViewMode::Sequential` (RenderDevice::NextView), so replay order == submission order and delta-encoding is sound; (b) bgfx retains uniform values in the backend store across submits/frames; (c) handles are interned by name so the cache must be global, not per-shader (e.g. `balls` is shared by 3 shaders); (d) samplers are never skipped (per-draw bindings); (e) cache resets on Shader create/destroy (handle idx recycling); (f) the one direct `bgfx::setUniform` on a shader-driven handle (MSAA depth resolve, SHADER_layer) is routed through the cache.

**Measured** on a Linux x64 BGFX/Vulkan cabinet (RTX 3080, playfield 3840x2160@119.88, Medieval Madness attract, instrumented 1 Hz counters): ~960 draws/frame carried **25,741 uniform records/frame before → 11,250 after (-56%)**; value-uniform records specifically dropped ~80% (remainder ≈ the 8-samplers-per-draw floor). VPX→BGFX submit time 1.86 → 1.74 ms avg. No visual difference (A/B screenshots), fps lock unchanged, and a 67-check behavior suite (two tables: PinMAME+DOF and Stern SPIKE paths) passed.

**Not tested:** Windows, OpenGL/DX9 backends (behavior deliberately unchanged there), VR/stereo (nEyes=2 arrays are compared as full byte ranges, but no stereo hardware was available).

https://claude.ai/code/session_015aZLT3JJJxnFXDzTLWuxCg

### DIFF
--- a/src/renderer/RenderCommand.cpp
+++ b/src/renderer/RenderCommand.cpp
@@ -117,8 +117,15 @@ void RenderCommand::Execute(const int nInstances, const bool log)
    {
       m_renderState.Apply(m_rd);
       m_shaderState->SetInt(SHADER_layer, RenderTarget::GetCurrentRenderLayer() < 0 ? 0 : RenderTarget::GetCurrentRenderLayer());
+      #if defined(ENABLE_BGFX)
+      // Apply uniforms straight from this command's snapshot instead of first
+      // copying it over m_shader->m_state: that copy ran for every single draw
+      // call (full uniform block + a shared_ptr copy per bound sampler).
+      m_shader->Begin(m_shaderState);
+      #else
       m_shader->m_state->CopyTo(false, m_shaderState);
       m_shader->Begin();
+      #endif
       m_rd->m_curDrawCalls++;
       switch (m_command)
       {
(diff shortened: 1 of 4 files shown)


dynajoe:
I had Claude searching for rendering optimizations and it preemptively created the PR. Closed since it was on autopilot.

PR #3762, opened 2026-08-05
Skip DMD texture re-upload when the source frame is unchanged

dynajoe:
**Disclosure: this change was authored by an AI assistant (Claude), per CONTRIBUTING's AI policy it is opened as a draft.**

Display sources only advance `DisplayFrame.frameId` when content actually changed, but the pull-side consumers (flasher DMD/DISPLAY, textbox DMD, scoreview, both B2S DMD overlays) called `UpdateTexture` once per rendered frame unconditionally — so a DMD changing ~60×/s was re-uploaded at full render rate, and a fully static DMD kept uploading forever.

Adds `ResURIResolver::DisplayUploadGate`: per consumer and per target texture, remember the (source id, frameId, geometry, texture object) last uploaded and skip when unchanged. Design notes: per-consumer (never shared — one puller's upload must not suppress another's), inequality comparison (frameId sequences restart/wrap), target texture pointer recorded AFTER upload (Update may reallocate), and textbox resets its gate when the text rasterizer writes the shared texture. Push-model paths (script DMDPixels), segment displays, and paths already gated on frameId are deliberately untouched; PUP video has no change token and is untouched.

**Measured** on a 120 Hz Linux BGFX/Vulkan cabinet (Medieval Madness, instrumented build): DMD-texture uploads previously ran at ~2.15× rendered-frame rate (4.3× the content change rate); with the gate they track the content change rate, and a static DMD drops to ~0 uploads/s. A 67-check behavior suite over a PinMAME+DOF table and a Stern SPIKE table (different DMD path) passed; DMD rendering verified on both plus B2S overlays.

**Not tested:** Windows, OpenGL/DX9 (code is backend-agnostic pull-side logic), FlexDMD-composited edge cases beyond the two suite tables.

https://claude.ai/code/session_015aZLT3JJJxnFXDzTLWuxCg

### DIFF
--- a/plugins/b2s/B2SDMDOverlay.cpp
+++ b/plugins/b2s/B2SDMDOverlay.cpp
@@ -120,12 +120,20 @@ void B2SDMDOverlay::Render(VPXRenderContext2D* ctx)
    if (m_frame.z == 0 || m_frame.w == 0)
       return;
 
-   switch (dmd.source->frameFormat)
+   // m_dmdTex is shared with the other overlay (scoreview/backglass), but the
+   // gate is per-overlay: each records what IT last uploaded, so neither can
+   // suppress the other's first upload after a change. Worst case is one
+   // redundant upload per content change while both windows are active.
+   if (m_displayUploadGate.NeedsUpload(dmd, m_dmdTex))
    {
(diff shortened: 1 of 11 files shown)


dynajoe:
Preemptive PR while I was searching for optimizations.

toxieainc:
But i think the basic idea is not the worst. Don't know if @vbousquet already has more involved plans though (so tackling this at a higher level).

PR #3766, opened 2026-08-06
Player: enumerate existing plugin audio sources after subscribing

pgboyd:
## Summary

This fixes a race where plugin audio sources can be registered before `Player`
subscribes to `CTLPI_AUDIO_ON_SRC_CHG_MSG`.

If the initial source announcement is missed, the plugin's audio source is
never enumerated, so subsequent `AudioUpdate` messages are discarded because
no matching audio lane exists.

## Fix

After subscribing to `OnAudioSrcChanged`, perform one initial enumeration:

```cpp
OnAudioSrcChanged(m_onAudioSrcChangedMsgId, this, nullptr);

This discovers plugin audio sources that were already registered before the
subscription while preserving the existing event-driven behavior for later
changes.

Testing
Linux BGFX

Wuthering Waves PuP-Pack
Before: MP3 decoded and PCM broadcast, but no music was audible.
After: audio lane is created and PuP music plays correctly.

This is a minimal four-line change with no impact on the normal
OnAudioSrcChanged event flow.

### DIFF
--- a/src/core/player.cpp
+++ b/src/core/player.cpp
@@ -778,6 +778,10 @@ Player::Player(PinTable *const table, const PlayMode playMode)
    msgApi->SubscribeMsg(m_pluginAPI.GetVPXEndPointId(), m_onAudioUpdatedMsgId, OnAudioUpdated, this);
    msgApi->SubscribeMsg(m_pluginAPI.GetVPXEndPointId(), m_onAudioSrcChangedMsgId, OnAudioSrcChanged, this);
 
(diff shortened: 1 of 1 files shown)


gitfool:
You're behind master. This is already [fixed](https://github.com/vpinball/vpinball/commit/69795f154b41e21ee8533eb47599a843f53d8e8d).

pgboyd:
Thanks. I confirmed that commit 69795f154 already implements the same initial audio-source discovery fix. My branch was behind current master, so I’ll close this PR as already resolved.

PR #3767, opened 2026-08-08
perf: slurp vpx into memory for network-attached storage

gitfool:
Note: generated with ai assistance and shows massive perf improvements in manual testing.

On standalone builds (macOS, Linux, Windows MinGW) the POLE structured-storage shim re-opens and re-parses the container metadata once per stream, then reads each 4 KiB block via an individual seekg+read with a per-call heap allocation. For a table with 1,858 streams (Fish Tales, 428 MiB) served from a NAS over WiFi, the multi-threaded stream load scatters reads across the file and hits a ~3 ms/block random-access penalty, resulting in a ~40 s load time.

Fix:

Read the entire .vpx sequentially into a shared vector<uint8_t> at open time (~3.3 s at 130 MiB/s over WiFi), then serve all POLE block reads via memcpy from that buffer. The buffer is shared across all PoleStorage clones via shared_ptr, so the 1,858 stream opens share one copy.

Changes:
- pole.h/cpp: Add Storage(const unsigned char*, uint64) constructor and matching StorageIO memory-backed path. loadBigBlocks() uses memcpy when backed by memory.
- standalone/PoleStorage.{h,cpp}: Create(filename) slurps the file then delegates to Create(shared_ptr<vector<uint8_t>>, path). Clone and OpenStorage share the buffer.

Measured on Fish Tales (428 MiB) over WiFi SMB, cold cache:
- ThreadPool phase (stream extraction): 40 s -> 0.1 s
- Total table load: 47 s -> 10 s
- Local SSD (no network): 7.3 s -> 5.3 s (also benefits from eliminating per-block syscall overhead)
- Peak memory: +file size during load (freed when load completes); checking 669 tables the largest is 493 MiB

PTAL @vbousquet

### DIFF
--- a/standalone/PoleStorage.cpp
+++ b/standalone/PoleStorage.cpp
@@ -2,13 +2,36 @@
 #include "PoleStorage.h"
 #include "PoleStream.h"
 
+#include <fstream>
+
 HRESULT PoleStorage::Create(const string& szFilename, const string& szName, IStorage** ppstg)
 {
-   POLE::Storage* pPOLEStorage = new POLE::Storage(szFilename.c_str());
+   // Slurp the entire file into memory for network-friendly access
+   std::ifstream ifs(szFilename, std::ios::binary | std::ios::ate);
+   if (!ifs.good())
+      return STG_E_FILENOTFOUND;
+
+   const auto size = ifs.tellg();
+   if (size <= 0)
+      return STG_E_FILENOTFOUND;
+
+   auto fileData = std::make_shared<std::vector<uint8_t>>(static_cast<size_t>(size));
+   ifs.seekg(0);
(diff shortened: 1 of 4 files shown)


toxieainc:
In principle i like the idea, but the problem is that this will harm mobile, etc, users that are already facing memory issues as-is. So if it could be detected if the load is going over net, or something similar, and then only spend the memory, that would be better IMHO.

vbousquet:
I also like the idea but as you said this need to handle all builds (linux but also mobile and Windows)

My preference would be to:
- drop Windows OLE storage and have a single codepath through POLE library (allows broader testing and simplified codebase)
- handle switching between in-memory vs storage loading based on available memory (can be tricky as the available memory varies a lot while loading)

vbousquet:

PR #3773, opened 2026-08-09
Throttle standalone progress log to at most once per second

gitfool:
The __STANDALONE__ path in ProgressDialog::SetProgress logged on every single texture load (~1500 lines for Blood Machines). Throttle to one log line per second, with immediate logging on phase changes.

Also fix m_progress being unconditionally set to -1 on text-only calls (value=-1), which caused subsequent GetProgress() to return -1 and produced bogus percentage values (e.g. 'Starting Game Scripts... 4%' instead of the correct 90%).

* [vpinball-before-fix.log](https://github.com/user-attachments/files/30866453/vpinball-before-fix.log)
* [vpinball-after-fix.log](https://github.com/user-attachments/files/30866454/vpinball-after-fix.log)

### DIFF
--- a/src/ui/win/ProgressDialog.cpp
+++ b/src/ui/win/ProgressDialog.cpp
@@ -50,8 +50,15 @@ void ProgressDialog::SetProgress(const string &text, const float value)
 #else
    if (value >= 0.f && m_progress != value)
    {
-      PLOGI.printf("%s %d%%", text.c_str(), (int)value);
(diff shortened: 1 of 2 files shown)


toxieainc:
Idea is not bad, but could you please move the 2 statics to the class instead, so that these are not outdated on a 2nd/3rd/.. run?

gitfool:
@toxieainc done

Issue #3632, opened 2026-07-07
10.8.1 BGFX Win64 - New plunger velocity system highly inconsistent across tables


DJOMaverick:
**Describe the bug**
From the latest plunger velocity rework builds, on a Pinscape with a linear slide-pot plunger, setting up the plunger velocity axis and using the native Pinscape plunger velocity unit leads to extraordinarily inconsistent results across different tables.  Some spot checks:

Nudge Test Table - works acceptably
Time Lord - Registers the slightest love-tap on the ball, not hard enough at full pull to even go 1/4 way up the chute
Futurama - works acceptably, if perhaps coming out somewhat slow
High Speed - Ball does not even react to the plunger

**To Reproduce**
Enable plunger position and velocity axes, set to Pinscape plunger velocity unit.  Load any of the above tables and try to launch a ball.

**Additional context**
This is actually seemingly identical behavior to behavior I observed and tried to tell MJR about when plunger velocity axis was first added as a feature to Pinscape and 10.8, right down to the same sample tables.  It rendered the velocity axis unusable then, and led to me just disabling it when nobody else appeared to notice it at the time.  Disabling the velocity axis now, however, does not appear to solve the problem.

[VPinballX.zip](https://github.com/user-attachments/files/29727719/VPinballX.zip)

[vpinball.log](https://github.com/user-attachments/files/29727721/vpinball.log)

dekay:
I am running Standalone on Linux and see similar issues with my Pinscape Pico equipped with the ADS1115 ADC connected to a slide potentiometer for my plunger. Previous to the refactor, position + velocity was problematic but position alone worked fine. Now even position only doesn't work either.

My testing was done on Addams Family. I can try to launch the ball once and it won't move, except maybe a tiny bit side to side in the trough. Then sometimes I can try again and it does launch. It seems to be very sensitive to where exactly the ball sits in the trough whether it launches or not. I tried it with both the new Custom Units setting for the Pinscape and with the gain set as high as 300%. I am also using the "plunger.z0" output from the Pico, which is calibrated position only without firing events. This is as recommended by mjr in the Pico docs for programs like VPX that can now process position and velocity at the same time.

The LiveUI plots show reasonable looking output as the plunger is pulled and released.

dekay:
@vbousquet I tried your most latest attempt to resolve this issue with commit 1f97cc880
[...]

PR #3780, opened 2026-08-11
Input: restore unit scales on default sensor mappings

Le-Syl21:
A sensor mapping's `scale` is a unit conversion: the acquired (normalized) axis value times the scale gives m/s² for accelerations, m/s for velocities, and per-unit/s for the plunger.

The default device mappings were left at the default scale of 1.0, so the raw value reaches the physics engine unconverted:

- nudge acceleration tops out at 1 m/s² full scale instead of 1g;
- the plunger impulse comes out two orders of magnitude too low — the `/100` in `HitPlunger::HitTest` is meant to be absorbed by the unit scale.

The plunger one is easy to see on a cabinet: pulling the plunger all the way back and releasing gets the ball about 3 cm up the shooter lane. The nudge one is quieter, the ball just barely moves.

This uses the very scales the sensor setup page already proposes as unit presets, and moves them next to `SensorMapping` so the defaults and the UI presets cannot drift apart. Devices configured automatically then behave like ones set up by hand.

Tested on a Pinscape KL25Z (plunger + accelerometer) on Linux/SDL3.

---

While tracking this down I ran into something else, which may or may not be intentional.

`NudgeIntentHandler::EvaluateImpulse` compares the segment strength against a hardcoded 1 m/s² threshold, below which no impulse is sent at all. It is applied after `Strength`, so it interacts with it, but it is not exposed anywhere and not mentioned in the sensor setup page.

On a 1g board with `Strength` at its neutral 1.0, that threshold sits at about 10% of the axis. That is a sensible floor and I am not suggesting it is wrong. But combined with the 25 ms impulse replacing the measured knock, reaching the 1° default plumb angle takes roughly 0.7g at the sensor, which is a firm shove. Someone tuning `PlumbThresholdAngle` upwards on a 1g board runs out of sensor range before running out of slider: past ~1.4° nothing can tilt any more.

Would it make sense to expose this threshold, or at least document it next to the sensor type? Happy to open a separate issue if you would rather keep this PR to the scales.

### DIFF
--- a/src/input/OpenPinDevHandler.cpp
+++ b/src/input/OpenPinDevHandler.cpp
@@ -391,11 +391,14 @@ OpenPinDevHandler::OpenPinDevHandler(InputManager &pininput)
                               map.MapAction(ButtonMapping::Create(deviceId, 26), m_inputManager.GetVolumeDownActionId()); // Audio volume down
                               std::unique_ptr<PlungerSensor> plunger = std::make_unique<PlungerSensor>(&m_inputManager);
                               plunger->GetPositionSensor()->SetMapping(SensorMapping::Create(deviceId, 0x200, SensorMapping::Type::Position));
-                              plunger->GetVelocitySensor()->SetMapping(SensorMapping::Create(deviceId, 0x201, SensorMapping::Type::Velocity));
+                              // Velocities are expected in m/s (per-unit/s for the plunger): without the unit scale, the reported
+                              // values are used as is and the plunger impulse ends up two orders of magnitude too low
+                              plunger->GetVelocitySensor()->SetMapping(
+                                 SensorMapping::Create(deviceId, 0x201, SensorMapping::Type::Velocity).WithScale(SensorUnitScale::pinscapePlungerVelocity));
                               map.MapPlunger(std::move(plunger));
                               std::unique_ptr<VPX::Physics::GamepadNudge> nudge = std::make_unique<VPX::Physics::GamepadNudge>(&m_inputManager);
-                              nudge->GetXSensor().SetMapping(SensorMapping::Create(deviceId, 0x204, SensorMapping::Type::Velocity));
-                              nudge->GetYSensor().SetMapping(SensorMapping::Create(deviceId, 0x205, SensorMapping::Type::Velocity));
+                              nudge->GetXSensor().SetMapping(SensorMapping::Create(deviceId, 0x204, SensorMapping::Type::Velocity).WithScale(SensorUnitScale::pinscapeNudgeVelocity));
+                              nudge->GetYSensor().SetMapping(SensorMapping::Create(deviceId, 0x205, SensorMapping::Type::Velocity).WithScale(SensorUnitScale::pinscapeNudgeVelocity));
                               map.MapNudge(std::move(nudge));
                            };
                            m_inputManager.SetDeviceDefaultMapping(deviceId, defaultMapping);
--- a/src/input/SDLInputHandler.h
+++ b/src/input/SDLInputHandler.h
@@ -356,8 +356,9 @@ private:
                plunger->GetPositionSensor()->SetMapping(SensorMapping::Create(deviceId, 0x0202, SensorMapping::Type::Position));
                map.MapPlunger(std::move(plunger));
                std::unique_ptr<VPX::Physics::GamepadNudge> nudge = std::make_unique<VPX::Physics::GamepadNudge>(&m_pininput);
-               nudge->GetXSensor().SetMapping(SensorMapping::Create(deviceId, 0x0200, SensorMapping::Type::Acceleration));
-               nudge->GetYSensor().SetMapping(SensorMapping::Create(deviceId, 0x0201, SensorMapping::Type::Acceleration));
+               // Accelerations are expected in m/s^2: without the unit scale, a full scale reading would be 1m/s^2 instead of 1g
+               nudge->GetXSensor().SetMapping(SensorMapping::Create(deviceId, 0x0200, SensorMapping::Type::Acceleration).WithScale(SensorUnitScale::gravity));
+               nudge->GetYSensor().SetMapping(SensorMapping::Create(deviceId, 0x0201, SensorMapping::Type::Acceleration).WithScale(SensorUnitScale::gravity));
                map.MapNudge(std::move(nudge));
             });
          break;
--- a/src/input/SensorMapping.h
+++ b/src/input/SensorMapping.h
@@ -7,6 +7,22 @@
 #include <complex>
 
 
+// Unit scales for the physics sensors.
+//
+// A sensor mapping's scale converts the acquired (normalized) axis value to the
+// unit the physics engine expects: m/s^2 for accelerations, m/s for velocities,
+// and per-unit/s for the plunger velocity (the 'unit' being the plunger travel
+// length). These are the scales used both by the default device mappings and by
+// the unit presets of the sensor setup page, so that a device set up
+// automatically behaves like one set up by hand.
+namespace SensorUnitScale
+{
+   constexpr float gravity = 9.80665f; // 1g accelerometer range, the usual default of Pinscape boards
+   constexpr float pinscapeNudgeVelocity = 4096.f / (20.f * 1000.f); // Pinscape measures in mm/s, multiplies by 20 and reports over a -4096/4096 range
+   constexpr float pinscapePlungerVelocity = 12.5f; // Per-unit velocity regarding the full plunger frame length (so no direct length unit)
+}
+
+
 class SensorMapping final
 {
 public:
--- a/src/ui/live/ingameui/SensorSetupPage.cpp
+++ b/src/ui/live/ingameui/SensorSetupPage.cpp
@@ -161,7 +161,7 @@ void SensorSetupPageSection::AppendSection(InGameUIPage* page, PhysicsSensor* se
    {
       // Accelerations must be provided to the engine in m/s^2 (acquired value x scale => m/s^2)
       // We propose some default scales, that is to say 1/2/4/8g (which is what Pinscape boards propose)
-      constexpr float g = 9.80665f;
+      constexpr float g = SensorUnitScale::gravity;
       int accUnit = abs(liveScale - g) < 0.01f ? 1 //
          : abs(liveScale - 2.f * g) < 0.01f    ? 2 //
          : abs(liveScale - 4.f * g) < 0.01f    ? 3 //
@@ -193,8 +193,8 @@ void SensorSetupPageSection::AppendSection(InGameUIPage* page, PhysicsSensor* se
    {
       // Velocities must be provided to the engine in m/s (acquired value x scale => m/s)
       // We propose some default scales (derived from Pinscape firmware)
-      constexpr float pinscapeDefaultNudge = 4096.f / (20.f * 1000.f); // Pinscape measure in mm/s then multiply by 20 and send a -4096/4096 range (this scale can be overriden)
-      constexpr float pinscapeDefaultPlunger = 12.5f;
+      constexpr float pinscapeDefaultNudge = SensorUnitScale::pinscapeNudgeVelocity; // Pinscape measure in mm/s then multiply by 20 and send a -4096/4096 range (this scale can be overriden)
+      constexpr float pinscapeDefaultPlunger = SensorUnitScale::pinscapePlungerVelocity;
       int velUnit = fabs(liveScale/pinscapeDefaultNudge - 1.f) < 0.005f   ? 1 //
                   : fabs(liveScale/pinscapeDefaultPlunger - 1.f) < 0.005f ? 2 //
                                                                           : 0;


DJOMaverick:
So I'm really excited about this PR, have been waiting for someone to address plunger issues, and want to believe, but- from your workflow artifact 5430, I could tell no appreciable change in plunger behavior tested, including in all the problem tables listed in issue #3632 .  It does feel like nudge is more responsive at defaults, however.

Le-Syl21:
Thanks for actually testing it — and your two observations together are the answer, I think.

This only changes the **default** mappings, the ones `ApplyDefaultDeviceMapping` installs when VPX auto-detects a device and you accept the layout prompt. If your `VPinballX.ini` already holds mappings for that device — or has `Device.<id>.NoAutoLayout = 1` — those defaults are never applied, so the old `scale` of 1.0 stays and the plunger stays weak. That's exactly the split you saw: nudge changed "at defaults", plunger didn't, because the plunger mapping was already written.

To try it on an existing setup, the scale is the **5th field** of the mapping line, and nothing else needs touching:

```
Mapping.Plunger0.Velocity = <dev>;517;V;<deadzone>;12.500000;1.000000
Mapping.Nudge0.AccX       = <dev>;512;A;<deadzone>;9.806650;1.000000
```

12.5 is per-unit/s for a Pinscape plunger; 9.80665 m/s² for a 1G accelerometer (19.6133 for 2G, 39.2266 for 4G). Same values the sensor setup page offers as unit presets.

If you'd rather not hand-edit, PinReady writes those lines for you: it asks the board for its accelerometer range over HID and rewrites the plunger and nudge mappings from what it gets back, so an existing ini is corrected instead of left alone. It won't touch anything else in your setup. https://github.com/Le-Syl21/PinReady — disclosure, I wrote it.

@vbousquet this raises a question I don't have an answer to: should existing `scale = 1.0` values be migrated? As it stands, this fix only reaches new setups, and everyone already running keeps a plunger two orders of magnitude short with nothing to indicate why. A one-time migration would be intrusive; leaving it means the bug outlives the fix for most users.

On #3632 — I don't think this PR addresses it, and I don't want to claim otherwise. That one is about the velocity system being inconsistent on a linear-pot plunger, which is a separate problem from the unit scale never being applied. Worth keeping the two apart.

DJOMaverick:
My word choice was poor; to clarify, I meant at the correctly selected stock values.  So my plunger test was with the velocity axis set to 12.5 (Pinscape plunger velocity) and the nudge axes were also set to appropriate Pinscape units (I have my Pinscape set for 2g range, so using that setting in VPX).

Le-Syl21:
Thanks for clarifying — that changes my read, and my previous answer doesn't apply to your setup. You already had the right scales in, by hand, so there was nothing left for this PR to fix on your side. Sorry for the detour.

Which makes your result the useful one: with correct units on both axes, the plunger still behaves the same. That points at the velocity handling itself rather than at the unit conversion, so #3632 stays open on its own merits and this PR doesn't touch it.

It does confirm one thing worth having, though: you arrived at 12.5 and your 2G range independently, which is exactly what the defaults now write. So the values aren't in question — what this fixes is that nobody should have to find them by hand in the first place.

One thing I'd be curious about, if you feel like testing: does the plunger behave differently with `Mapping.Plunger0.Velocity` removed entirely, leaving only `Position`? On a Pinscape reporting both, that tells whether the velocity path is the problem or whether position alone is already off — it's the split I can't make from here, since my own plunger works.

Le-Syl21:
One more thing that may help you narrow it down: the scale is a pure ratio, so whatever mismatch exists between what your board reports and what VPX is told shows up as a constant factor — and the direction of the error is readable from the symptom.

For the accelerometer, VPX offers four unit presets (`SensorSetupPage.cpp`), and declaring one your firmware doesn't use multiplies everything by `VPX ÷ firmware`:

| firmware ↓ / VPX → | 1G | 2G | 4G | 8G |
|---|---|---|---|---|
| **1G** | correct | ×2 | ×4 | ×8 |
| **2G** | ÷2 | correct | ×2 | ×4 |
| **4G** | ÷4 | ÷2 | correct | ×2 |
| **8G** | ÷8 | ÷4 | ÷2 | correct |

Above the diagonal VPX over-reads: nudge is violent and the table tilts on nothing. Below it under-reads: the ball barely moves and tilting gets hard. You're on 2G both sides, so that one is clean.

For the plunger, `scale` converts to per-unit/s over the plunger travel, and there are only two presets — `12.5` (Pinscape plunger) and `4096/(20×1000)` = `0.2048` (Pinscape nudge velocity, which is a different quantity). The ratio between those two is ~61×, so picking the wrong one is not subtle.

Which gives a way to read your symptom directly:

| What you see | What it means |
|---|---|
| ball crawls a few cm at full pull | scale far too low — the `/100` in `HitPlunger::HitTest` isn't being absorbed |
| ball launches at the same speed regardless of pull depth | the velocity axis isn't tracking; position is driving it alone |
| ball fires much too hard, or erratically | scale too high, or the velocity axis is reporting in a different unit than assumed |
| behaviour varies table to table with identical settings | not the scale — tables differ in how they read plunger state |

If yours is the second or fourth line, the unit is not your problem and #3632 is the right place for it. If it's the first, it's worth checking that the velocity axis is the one your board actually sends velocity on — on a Pico that's whatever you assigned to `nudge.vx`/`plunger.v`, not a fixed number.

DJOMaverick:
So to clarify any other assumptions here, I'm using traditional Pinscape (KL on Rigmaster), and Win64.  The Pinscape velocity axis is set up correctly, verifiable by response on graphs in vpx).  Scale is, indeed, set to 12.5 unit.

From your four categories, problem is type 1, weaker than expected response across the board, but mostly usable.  Also, on all the tables mentioned in that issue, the response with velocity axis enabled does not appear much (or at all) different than disabling velocity axis.

dekay:
> It does confirm one thing worth having, though: you arrived at 12.5 and your 2G range independently, which is exactly what the defaults now write. So the values aren't in question — what this fixes is that nobody should have to find them by hand in the first place.

I don't understand what this PR is trying to accomplish. Nobody is "finding them by hand". VPX is properly setting the values when going through the existing nudge configuration screens as selected from dropdown lists. Below is what I wrote on Discord earlier today. Perhaps talking to a real person instead of an AI would help resolve this issue quicker.

---

I just tried with an ordinary stock build of VPX from weeks ago and VPX is already populating the ini file correctly. In Plunger settings, there is Velocity sensor -> Sensor unit which has a dropdown list from which "12.5 p.u/s" can be selected. In Nudge settings, there is Side / Front Acceleration sensor -> Sensor unit -> "2 Gravity". That results in the following entries in the VPinxballX.ini file that match the values in your message in the 5th value.

```
Mapping.Plunger0.Position = SDLJoy_0300252cfafa0000ebea000011010000_1;514;P;0.000000;1.007634;1.000000
Mapping.Plunger0.Velocity = SDLJoy_0300252cfafa0000ebea000011010000_1;517;V;0.000000;12.500000;1.000000
<snip>
Mapping.Nudge0.AccX = SDLJoy_0300252cfafa0000ebea000011010000_1;512;A;0.000000;19.613300;1.000000
Mapping.Nudge0.AccY = SDLJoy_0300252cfafa0000ebea000011010000_1;513;A;0.000000;19.613300;1.000000
Mapping.Nudge0.VelX = SDLJoy_0300252cfafa0000ebea000011010000_1;515;V;0.000000;20.000000;1.000000
Mapping.Nudge0.VelY = SDLJoy_0300252cfafa0000ebea000011010000_1;516;V;0.000000;20.000000;1.000000
```

Le-Syl21:
Fair enough, I wasn't clear, so let me be precise about what this touches.

The settings pages are correct and this PR doesn't change them. The dropdown you used writes the right scale, and always has.

It changes the other path: `ApplyDefaultDeviceMapping` (`InputManager.cpp:230`), the mapping VPX installs when it detects a device and you accept the "configure this automatically?" prompt. That path built its sensor mappings with no unit scale, so accelerations and velocities were left at 1.0, and it isn't live-only: `AddSensor` calls `sensor->Save()` then `settings.Save()`, so those 1.0 values get written into your ini and stay there. Scale is a plain multiplier (`m_value = clamp(value, -limit, limit) * m_scale`), so a full-scale accelerometer read lands as 1 m/s² instead of 9.80665.

For a traditional Pinscape over SDL that's the nudge acceleration axes, since that default deliberately leaves the velocities unmapped. On the OpenPinDev path it's the plunger velocity and both nudge velocities. In both cases the PR just makes the defaults write the same numbers the dropdowns write.

One thing I got wrong earlier in this thread, worth correcting: I said the fix only reaches new setups. It doesn't. `SensorMapping::IsSame` compares the scale, so anyone still carrying the old 1.0 defaults no longer matches the proposed layout, the prompt comes back, and accepting it rewrites the correct values. Existing auto-configured setups are reached after all.

None of which changes anything on your machine, since you configured yours by hand, and none of which is a fix for #3632 either. I said earlier the two should stay separate and I still think so.

— Claude, reviewed by Sylvain

dekay:
While the [nudge docs in the Pinscape Build Guide V2](http://mjrnet.org/pinscape/BuildGuideV2/BuildGuide.php?sid=nudge) state that 1G  is the default range, that appears to be an error.

- The MMA8451Q used on the KL25Z Freedom board is +/- 2G minimum. [Reference](https://www.nxp.com/docs/en/data-sheet/MMA8451Q.pdf)
- The Pinscape Pico guide that references that same accelerometer contains a json snippet confirming 2G is the most sensitive setting. [Reference](https://github.com/mjrgh/PinscapePico/blob/main/Guides/MMA8451Q-Setup.md)

```
mma8451q: {
   i2c: 0,        // which I2C unit you're using - 0 if you connected the device to I2C0 pins, 1 for I2C1 pins
   addr: 0x1D,    // the chip's I2C address - 0x1C if you connected "A" (SA0) to GND, 0x1D if you connected it to 3V3 OUT or left it unconnected
   interrupt: 22, // the GPIO port number where you connected the INTERRUPT (INT or INT1) pin - OMIT IF NOT CONNECTED
   gRange: 2,     // the "g" range to select - 2 selects +/- 2g, which is the most sensitive setting, and usually the best for virtual pinball
},
```

Le-Syl21:
Careful there. The chip's range and the range the axis reports are not the same thing. Pinscape runs the MMA8451Q in 2G hardware mode and rescales the reading. What it rescales to is a firmware setting. Its default is 1G.

Config variable 4 holds it. Byte 4 is the dynamic range: 0 is ±1G rescaled from the 2G hardware mode, then 1, 2, 3 for 2G, 4G, 8G. So the guide is not wrong. It describes what leaves the board. The datasheet describes what the part can do. Your board reports 2G because it is set to 2G.

That is the case for having the board state it rather than the user. The KL25Z answers that query today. The others do not, so I asked:

- Pinscape Pico: mjrgh/PinscapePico#117. The range sits in the config file behind libusb, not in the feedback protocol. mjr has taken it on.
- DudesCab: asked on arnoz's Discord, https://discord.com/channels/729406528683245604/1219343910799802518/1537086240787013774
- PinOne: philipellisis/csd-docs#3

When a board reports its range, VPX can read it at detection time and set the scale itself. The dropdown stays for the boards that keep quiet. That is a separate PR, once there is something to read.

This one does not go that far. It only makes the auto-configuration path write the same scales the dropdowns already write.

dekay:
Can you point to where the **Pico** firmware rescales the readings?

For point 1, what mjr has taken on is is adding an interface to report back the range setting. He made no mention of adding a range for the Pinscape Pico of 1G, nor was he asked to.

[From the Pinscape Pico json docs](http://mjrnet.org/pinscape/PinscapePico/Help/JSONConfigRef.htm#mma8451q)

> mma8451q.gRange numberOptional
> 
> Sets the dynamic range for the accelerometer readings, in units of standard Earth gravity. This device allows settings of 2, 4, or 8. The default is 2, which is usually the best setting for a virtual pin cab, since this provides the best sensitivity for mild nudges.

Le-Syl21:
You're right, and I conflated two firmwares.

The rescaling I described is the KL25Z one: `Pinscape_Controller/main.cpp`, `rawToReport()`, divides the native 14-bit reading by 4096 in the 1G mode and by 8192 otherwise, so a ±4096 axis means ±1 g even though the chip runs in 2 g hardware mode. Config variable 4 carries that setting (`cfgVarMsgMap.h`, case 4: `accel.range`), and my own board reads range=0 — ±1 g, hardware 2 g.

The Pico does no such thing. `Firmware/Devices/Accel/MMA8451Q.cpp` shifts the 14-bit sample up into an `int16_t`, and `Firmware/Nudge.h` says why: the firmware picks `int16_t` as its canonical type because that is what VP ingests, and requires each chip driver to rescale to it. That is a width conversion, not a range one — full scale stays at `gRange`, 2, 4 or 8 as the docs say. No 1G mode, and I wasn't suggesting mjr add one.

Which leaves the point I care about unchanged, and smaller than I made it sound: the value VPX needs is a firmware setting rather than a property of the chip, and it already differs between the two Pinscape generations for the same MMA8451Q. That is the case for having the board report it rather than having the user pick it.

dekay:
So this PR would pick the correct recommended value for the Pinscape KL25Z but an invalid value for the Pinscape Pico. Whether or not the board reports the value is inconsequential to this PR, and no available board currently supports this function.

Arguing further with an AI is not worth the frustration. I'm out.

Le-Syl21:
You're right.

Issue #3647, opened 2026-07-11
Nudge: 4G and 8G unit scales show as 100%


dekay:
Custom unit scales for 1G and 2G show as 980% and 1961% as expected. 4G and 8G scales show as 100%. I'd expect them to be multiples of 980%?

<img width="693" height="463" alt="Image" src="https://github.com/user-attachments/assets/b2e96675-b019-4d12-8071-f1cff4a306d1" />
  
  
<img width="706" height="499" alt="Image" src="https://github.com/user-attachments/assets/a4dce8de-9d59-4966-8b0e-0398974143ae" />

PR #3782, opened 2026-08-12
docs: add a 10.8.1 configuration guide

Le-Syl21:
10.8.1 rewrote several subsystems rather than extending them. The settings that replaced the old ones aren't documented anywhere, so cabinet owners are currently reverse-engineering their own `VPinballX.ini` — reading source, or guessing.

The most visible symptom is the plunger: a sensor mapping's `scale` is a unit conversion that defaults to 1.0, so the launch impulse comes out two orders of magnitude short and nothing on screen says why. That's what sent me digging, and #3780 fixes the defaults — but the docs gap remains for anyone with an existing ini.

This adds `docs/10.8.1/`, eight chapters in English and French:

- **input** — the sensor mapping schema and its six fields, the unit scales, the three nudge modes and the hardcoded 1 m/s² intent threshold, the plumb
- **view** — the Window projection, the two autofit modes, the glass heights the frame is built from
- **window** — the four outputs and their shared schema, and the physical screen dimensions autofit needs
- **rendering** — frame pacing now being the default, the three AA stages, reflections, stereo
- **audio** — devices, the six speaker modes, and the per-source gain added in July
- **plugins** — one section per plugin, and why a disabled plugin's keys appear nowhere
- **vr** — OpenXR, cabinet placement, the preview window
- **removed** — what 10.8 had and no longer exists, with what replaced it, or the fact that nothing did

Everything was verified against the code rather than inferred from behaviour, and every chapter lists the files and functions it describes so a reader — or a future maintainer — can check whether it still holds. Where a change alters a way of reasoning rather than just a key, the mechanics are explained: the nudge chain and the tilt thresholds get the most space, since that's where the questions come from.

Two things worth flagging before review.

**Scope**: this covers changes since the 10.8.0 release, not since the CI's `VERSION_START_SHA`, which is the origin of the revision counter and predates 10.8.0 by three years.

**It's a first pass.** It's deliberately short, and it doesn't cover LiveUI, or the internal settings of plugins other than PinMAME. I'm happy to extend, restructure, drop the French, or fold it into the existing `docs/` layout — whatever fits how you'd like the docs to grow. If a section reads wrong, tell me which and I'll fix it against the code.

**AI disclosure**, per CONTRIBUTING.md — I should have put this in from the start, sorry. I wrote this with Claude Code: it read the source to establish each behaviour, and drafted the prose in both languages. I set the scope and the chapter structure, and I checked the content against the code myself. That's also why every chapter ends with the files and functions it describes: the claims are meant to be verifiable rather than taken on trust, by you now and by whoever maintains the docs later. The French is my own language, so I can vouch for that half directly.

### DIFF
--- a/docs/10.8.1/README.md
+++ b/docs/10.8.1/README.md
@@ -1,0 +1,41 @@
+# VPinballX 10.8.1 — configuration
+
+🇬🇧 [English](#-english) | 🇫🇷 [Français](#-français)
+
+---
+
+## 🇬🇧 English
+
+- **[Input, nudge and tilt](input_eng.md)** — sensor mappings, unit scales, the
+  three nudge modes, the plumb.
+- **[View and cabinet fitting](view_eng.md)** — the Window projection, autofit,
+  glass heights.
+- **[Windows and displays](window_eng.md)** — the four outputs, their shared
+  schema, physical screen dimensions.
+- **[Rendering](rendering_eng.md)** — frame pacing, anti-aliasing, reflections,
+  stereo.
+- **[Audio](audio_eng.md)** — devices, speaker modes, the per-source gain.
+- **[Plugins](plugins_eng.md)** — one section per plugin, and where their keys
+  really come from.
+- **[VR](vr_eng.md)** — OpenXR, cabinet placement, the preview window.
+- **[Removed settings](removed_eng.md)** — what 10.8 had, what replaced it.
+
+---
+
+## 🇫🇷 Français
+
+- **[Entrées, nudge et tilt](input_fra.md)** — mappings de capteurs, échelles
+  d'unité, les trois modes de nudge, le pendule de tilt.
+- **[Vue et ajustement cabinet](view_fra.md)** — la projection Window,
+  l'ajustement automatique, les hauteurs de vitre.
+- **[Fenêtres et écrans](window_fra.md)** — les quatre sorties, leur schéma
+  commun, les dimensions physiques de l'écran.
+- **[Rendu](rendering_fra.md)** — frame pacing, anticrénelage, réflexions,
+  stéréo.
+- **[Audio](audio_fra.md)** — périphériques, modes d'enceintes, le gain par
+  source.
+- **[Plugins](plugins_fra.md)** — une section par plugin, et d'où viennent
+  vraiment leurs clés.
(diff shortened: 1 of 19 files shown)


Smitty2k1:
WOW! Thank you!

Le-Syl21:
I've edited the description to add the AI disclosure CONTRIBUTING.md asks for — flagging the edit here rather than making it silently. It should have been there when I opened this; the omission was an oversight, not a position.

vbousquet:
I'm really not easy with merging large AI content like this in the main repository. I think this can be useful but should be kept and maintained outside of the main repo. @toxieainc What do you think ?

Regarding the content, one thing to be aware is that the Live UI is what is exposed to the user and therefore for which tutorials would be needed. The INI layout and definition is not part of the public API (nor the code or commits referenced by the AI generated docs). In normal use, users should never have to search, open & tweak it directly. There is absolutely no guarantee that this will stay stable (layout, definition will change with each release). So I would avoid to document this part as it is misleading toward manual edition which used to be a common causes of crashes and misunderstanding (it is better now as VPX validates the INI file when loaded, but with the hundreds of settings this is still something very touchy and highly discouraged).

dekay:
Case in point: the input doc says "Driving sensitivity through scale lies to the engine about what the sensor is, and makes a 4 g or 8 g board impossible to describe." This isn't user documentation. This was [a bug fixed almost a couple months ago](https://github.com/vpinball/vpinball/issues/3647).

An end user doesn't need to know how devices are declared in the `[Input]` section of the ini, or that "The preference folder is SDL_GetPrefPath(nullptr, "VPinballX")". This type of content is more like intro developer docs for some moment in time rather than higher level user docs that should more or less hold up over the period of a release.

Having said that, there are some useful bits here offering insights to someone interested in how things work under the hood. Maybe they could live as gists in le-Syl21's repo or something like that.

Le-Syl21:
Fair on both counts — closing this.

dekay's example is the one that settles it: the nudge-scale line he quoted describes behaviour that was fixed in #3647, which is exactly what goes wrong with a document written against one moment in the code. User documentation has to hold up across a release, and this was written from the inside out.

The parts that explain how things work under the hood will live in my own repo instead, as suggested. Thanks to you both for actually reading it.

PR #3814, opened 2026-08-23
Reduce .vpx container read cost without buffering the file

gitfool:
> **Draft, and AI-assisted.** Opened as a draft per the project's policy. I have measured
> and verified every claim below myself, and I have tried to be equally clear about what is
> *not* measured. Happy to split, reorder or drop any part of this.

Supersedes the approach in #3767. That PR buffered the whole `.vpx` in memory, and
@toxieainc and @vbousquet both raised the memory cost. Measuring it properly showed the
buffer was treating a symptom: the real problem is *how* the container is read, not where it
is read from. This PR fixes the read pattern and **does not change peak memory**.

I would like to close #3767 in favour of this, if you agree with the direction.

## Summary

Loading a table opens every stream through the POLE shim. Three things made that far more
expensive than it needed to be:

1. Every stream read issued one seek and one block-sized request **per block**, and
   `StreamIO::read` did that on every call regardless of how little was asked for. `BiffReader`
   reads structure fields 1, 2 and 4 bytes at a time, so reading a 4-byte field cost a full
   4 KiB read.
2. A stream under the 4096-byte threshold is stored as 64-byte mini-blocks packed inside
   ordinary blocks, so 64 mini-blocks share one block. Reading such a stream fetched that same
   block once per mini-block, up to 64 times over.
3. `PoleStorage::OpenStream` cloned the storage before opening a stream, and each clone
   re-parsed the entire container: header, FAT, mini FAT and directory tree. Fish Tales has
   1860 streams, so it parsed the container 1860 times.

Over a network filesystem each of those is a round trip rather than a syscall, which is why
this shows up so badly on NAS-hosted tables.

I originally framed this as a fix for network-attached storage. That was wrong, or at least far
too narrow, and measuring local disk afterwards showed it. The cost this removes is per-block
syscall overhead, not data transfer, so it pays off on any medium. The clearest evidence is that
master's stream-extraction phase barely notices whether the page cache is cold: 2.537s cold
against 2.339s warm on a local SSD. Almost none of master's cost is the actual reading.

Real application, both phases measured from the log, medians of 3 interleaved repetitions,
`purge` before every cold load, AC power.

| | | master | this PR |
| --- | --- | --- | --- |
| **local SSD, cold** | stream extraction | 2.537s | **0.461s** (5.5x) |
| | total `LoadGameFromFilename` | 7.51s | **5.41s** (1.39x) |
| **local SSD, cold**, Dark Chaos | stream extraction | 4.527s | **0.375s** (12.1x) |
| | total `LoadGameFromFilename` | 10.45s | **6.27s** (1.67x) |
| **local SSD, warm** | stream extraction | 2.339s | **0.207s** (11.3x) |
| **SMB over WiFi, cold** | stream extraction | 39.21s | **18.90s** (2.07x) |
| | total `LoadGameFromFilename` | 45.91s | **25.60s** (1.79x) |
| **SMB over WiFi, cold**, Dark Chaos | stream extraction | 33.06s | **10.86s** (3.04x) |
| | total `LoadGameFromFilename` | 40.43s | **18.27s** (2.21x) |
| peak resident memory | | 8570 MB | 8155-8452 MB |
| peak phys_footprint | | 12992 MB | 12627-13021 MB |

Fish Tales is 428 MB and 1860 streams, Dark Chaos 312 MB and 3555 streams of which 2262 are
mini streams. Unless noted, rows are Fish Tales.

Two things I would draw out of that. The gain is *larger* on the stream-heavy table, 3.04x
against 2.07x cold over SMB, because the mini-stream fix has more to work with. And a user with
tables on an ordinary SSD sees 1.39x to 1.67x off total load time, which is the case I had no
numbers for when I first opened this.

Memory is unchanged within run-to-run variance, by design: the readahead window is allocated
lazily and never exceeds the stream's own length.

## The commits

Each is independently verifiable and they can be taken separately.

**`POLE: coalesce block reads and add a per-stream readahead window`**
`loadBigBlocks` now extends each run of consecutive blocks and issues one request per run.
Block chains in a `.vpx` are contiguous in practice: across three 450 MB tables every big
stream occupied exactly one run, averaging 352 KB. `StreamIO::read` serves big-stream reads
from a 256 KiB window, sized from the measured average run length above.

This commit also stops `loadBigBlocks` when a chain points past the end of the file. The
clamp below it computes `filesize - pos` in unsigned arithmetic and that result was used
directly as a read length, so a truncated or malformed chain could produce an enormous read.
That is pre-existing, and worth having regardless of the rest.

**`POLE: stop re-reading the same block for every mini-block`**
Needs two halves, and either alone does nothing. `StreamIO::read` asked for one mini-block per
call via `loadSmallBlock`, which wraps a single block into a vector, so a memo inside
`loadSmallBlocks` would never see a second block to skip. The small-file branch now fills the
same window the big-file branch uses; a mini stream is by definition under the threshold, so
its whole chain fits in one window. `loadSmallBlocks` then remembers the last block it loaded.

**`standalone: parse the .vpx container once instead of once per stream`**
The clone was not gratuitous: POLE reads through a single `std::fstream`, and table loading
opens streams from a thread pool, so per-stream storage also gave per-stream file handle.
Sharing the parse means sharing the handle, so `PoleStream::Read` serialises on a mutex held
beside the parsed container in `PoleSharedStorage`.

Serialising is not a compromise here. Reading the container from twelve threads at once
measured **3.4x slower** than from one at the same traversal order, because concurrent readers
at scattered offsets defeat readahead in the filesystem. Contention on this mutex is doing
useful work.

Nothing else on that path needs the lock: `DirTree::entry` only mutates under
`create && writeable` and this opens read-only, `AllocTable::follow` is `const`, and the
readahead window lives on `StreamIO`, one per open stream. The reference counts on
`PoleStorage` and `PoleStream` become atomic, since streams are opened and released from pool
threads.

**`PinTable: fix data race on the load progress counters`**
Unrelated to the above and separable. The three counters driving the progress bar are
incremented by pool workers and read by the main thread with no synchronisation.
ThreadSanitizer flags it at `pintable.cpp:1660`. Effect was minor, since a torn read only
misdraws a progress bar, but it is undefined behaviour and a two-line fix. Note the sum of the
three is still not a consistent snapshot, which is all a progress indicator needs.

## Which platforms this touches

- `third-party/include/pole/pole.cpp` is compiled by standalone builds **and by the FlexDMD
  plugin on Windows**, so the POLE changes are not standalone-only. They are behaviour
  preserving and byte-verified, but reviewers should know Windows is in scope here.
- `standalone/PoleStorage.*` and `standalone/PoleStream.*` are standalone only.
- `src/parts/pintable.cpp` is all platforms.

## How it was verified

**Byte-identical output.** A benchmark harness reads every stream in a container and folds an
FNV hash over all bytes in a fixed stream order. Before and after produce
`fnv=77d1733d72ba7e6f` across every combination of read strategy, traversal order, thread
count (1 to 12) and read size (4 B to 1 MiB). A short read from a stale mount would change the
hash, so this also guards against an arm looking fast because it read less.

**Table structure unchanged.** A fingerprint of the content-derived log output (duplicate-name
renames, layer renames, image downsizing) is identical across every load, before and after.

**ThreadSanitizer.** Instrumented build, full table load. Zero races in any file this PR
touches. I checked that this means something: `__tsan_` reference counts confirm each object
was instrumented, and the run completed a full load with 1826 streams read through the shared
handle on 12 pool threads, which is exactly the scenario the shared-parse commit creates.
TSan does report 77 pre-existing races elsewhere, listed under follow-ups.

**Methodology.** Cold numbers are medians of 3 interleaved repetitions with `purge` before
every arm, on AC power. That is deliberate: this link drifts, two runs of an identical
configuration came out 62% apart, and single-shot measurements led me to two wrong conclusions
before I started repeating them. Any difference smaller than the largest observed per-arm
spread is treated as noise.

## What is *not* verified

- **Mobile.** Not measured at all. Peak memory is unchanged so I do not expect a regression,
  but I cannot demonstrate it.
- **Windows.** Not measured. The POLE changes apply there via FlexDMD, as noted above.
- **Table breadth.** Two tables, picked to differ as much as possible in stream count and
  physical layout. The gain is clearly table-dependent, 2.07x against 3.04x cold over SMB, so
  please read any single figure as one data point rather than a general one.
- **Only this machine.** One macos arm64 laptop, one internal SSD, one WiFi link to one NAS. The
  direction has been consistent across four storage conditions now, but the magnitudes are mine.

## Follow-ups, as stacked PRs

1. **A single ordered reader.** The remaining and larger win. Streams are physically laid out
   almost in reverse of the order the loader reads them: on Fish Tales 93.9% of reads move
   backwards through the file. Reading in ascending file offset instead measured 4.9x faster
   again, on top of this PR. That needs a position per stream, which only POLE can supply, so
   it will come with a capability query that returns false on Windows OLE and lets callers fall
   back to declaration order. I have this working and held it back so this PR contains no
   unused API.
2. **The 77 pre-existing races.** Concentrated in `player.cpp`, `RenderDevice.cpp`,
   `Renderer.cpp`, `PerfUI::RenderStats`, `FrameProfiler::Reset` and `HitBall::UpdateVelocities`.
   I have not triaged them and some may be benign; happy to open an issue with the report.
3. **Two unrelated build fixes**, kept out of here on purpose: `.gitignore` for the clangd
   index cache, and `BX_CONFIG_DEBUG` only being set for `Debug` and `Release` generator
   expressions, which makes `-DCMAKE_BUILD_TYPE=RelWithDebInfo` fail outright on macos, linux,
   ios and android. windows-mingw already used the correct form.

## One thing that may interest @vbousquet

You mentioned wanting to drop Windows OLE storage in favour of a single POLE codepath. While
tracking down the layout problem I found that VPX already creates streams in ascending order:
the `GameItem`, `Sound` and `Image` loops in `SaveToStorage` all write ascending `i`. The
reversal comes from the OLE transacted commit relocating streams on `Commit(STGC_DEFAULT)`,
not from VPX.

So moving writes onto POLE would likely fix the layout at source, since POLE appends blocks as
streams are created, and would make the follow-up ordering work unnecessary on Windows. I
could not test that here because `PoleStorage::CreateStream` is `E_NOTIMPL`, so treat it as a
prediction rather than a result.

### DIFF
--- a/src/parts/pintable.cpp
+++ b/src/parts/pintable.cpp
@@ -4,6 +4,7 @@
 #include "pintable.h"
 
 #include <algorithm>
+#include <atomic>
 #include <fstream>
 #include <sstream>
 
@@ -1569,7 +1570,7 @@ HRESULT PinTable::LoadGameFromFilename(const std::filesystem::path &filename, VP
             ThreadPool pool(g_app->GetLogicalNumberOfProcessors());
             vector<IEditable *> parts;
             parts.resize(csubobj);
-            int nLoadedParts = 0;
+            std::atomic<int> nLoadedParts { 0 };
             for (int i = 0; i < csubobj; i++)
             {
                pool.enqueue(
@@ -1606,7 +1607,7 @@ HRESULT PinTable::LoadGameFromFilename(const std::filesystem::path &filename, VP
 
             assert(m_vsound.empty());
             m_vsound.resize(csounds);
-            int nLoadedSounds = 0;
+            std::atomic<int> nLoadedSounds { 0 };
             for (int i = 0; i < csounds; i++)
             {
                pool.enqueue(
@@ -1630,7 +1631,7 @@ HRESULT PinTable::LoadGameFromFilename(const std::filesystem::path &filename, VP
 
             assert(m_vimage.empty());
             m_vimage.resize(ctextures);
-            int nLoadedImages = 0;
+            std::atomic<int> nLoadedImages { 0 };
             for (int i = 0; i < ctextures; i++)
             {
                pool.enqueue(
--- a/standalone/PoleStorage.cpp
+++ b/standalone/PoleStorage.cpp
@@ -7,14 +7,9 @@ HRESULT PoleStorage::Create(const string& szFilename, const string& szName, ISto
    POLE::Storage* pPOLEStorage = new POLE::Storage(szFilename.c_str());
 
    if (pPOLEStorage->open() && pPOLEStorage->result() == POLE::Storage::Ok) {
-      PoleStorage* pStg = new PoleStorage();
-      pStg->m_szFilename = szFilename;
-      pStg->m_szPath = szName;
-      pStg->m_pPOLEStorage = pPOLEStorage;
+      auto shared = std::make_shared<PoleSharedStorage>(pPOLEStorage);
 
-      pStg->AddRef();
-
-      *ppstg = pStg;
+      *ppstg = Attach(shared, szName);
 
       return S_OK;
    }
@@ -24,19 +19,25 @@ HRESULT PoleStorage::Create(const string& szFilename, const string& szName, ISto
    return STG_E_FILENOTFOUND;
 }
 
-HRESULT PoleStorage::Clone(PoleStorage* pPoleStorage, IStorage** ppstg)
+PoleStorage* PoleStorage::Attach(const std::shared_ptr<PoleSharedStorage>& shared, const string& szName)
 {
-   return PoleStorage::Create(pPoleStorage->m_szFilename, pPoleStorage->m_szPath, ppstg);
+   PoleStorage* pStg = new PoleStorage();
+   pStg->m_shared = shared;
+   pStg->m_szPath = szName;
+
+   pStg->AddRef();
+
+   return pStg;
 }
 
 HRESULT PoleStorage::StreamExists(const string& szName)
 {
-   return m_pPOLEStorage->exists(m_szPath + '/' + szName) ? S_OK : STG_E_INVALIDNAME;
+   return m_shared->storage->exists(m_szPath + '/' + szName) ? S_OK : STG_E_INVALIDNAME;
 }
 
 POLE::Storage* PoleStorage::getPOLEStorage()
 {
-   return m_pPOLEStorage;
+   return m_shared->storage;
 }
 
 string PoleStorage::getPath()
@@ -58,8 +59,7 @@ STDMETHODIMP_(ULONG) PoleStorage::Release()
    ULONG dwRef = --m_dwRef;
 
    if (dwRef == 0) {
-      delete m_pPOLEStorage;
-
+      // The parsed container outlives this handle if any stream still holds it.
       delete this;
    }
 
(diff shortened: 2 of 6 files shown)


toxieainc:
Looks good to me now, @vbousquet, what do you think?

vbousquet:

PR #3815, opened 2026-08-23
Fix BX_CONFIG_DEBUG for build types other than Debug and Release

gitfool:
`BX_CONFIG_DEBUG` was set by a pair of generator expressions, one for `Debug` and one for
`Release`. Any other configuration matches neither, so the define is absent and
`third-party/include/bx/config.h` stops the build with:

```
error "BX_CONFIG_DEBUG must be defined in build script!"
```

That makes `-DCMAKE_BUILD_TYPE=RelWithDebInfo` and `MinSizeRel` fail outright on macos, linux,
ios and android. The windows-mingw target already used the single-expression form and was
unaffected, so this just makes the other four match it.

Checked on macos by reading the define out of `flags.make` for each build type:

| build type | before | after |
| --- | --- | --- |
| Debug | `BX_CONFIG_DEBUG=1` | `BX_CONFIG_DEBUG=1` |
| Release | `BX_CONFIG_DEBUG=0` | `BX_CONFIG_DEBUG=0` |
| RelWithDebInfo | absent, build fails | `BX_CONFIG_DEBUG=0` |
| MinSizeRel | absent, build fails | `BX_CONFIG_DEBUG=0` |

`Debug` and `Release` are unchanged.

Found while trying to build `RelWithDebInfo` to run a sanitizer against optimised code, which
is also the case it is most useful for.

Written with AI assistance; I verified the behaviour above myself.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -226,8 +226,7 @@ if(PLATFORM STREQUAL "windows")
       )
       target_compile_definitions(vpinball PRIVATE
          ENABLE_BGFX
-         $<$<CONFIG:Debug>:BX_CONFIG_DEBUG=1>
-         $<$<CONFIG:Release>:BX_CONFIG_DEBUG=0>
+         $<IF:$<CONFIG:Debug>,BX_CONFIG_DEBUG=1,BX_CONFIG_DEBUG=0>
       )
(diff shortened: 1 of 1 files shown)


toxieainc:
Thanks!

PR #3817, opened 2026-08-23
Read table streams in physical layout order on a single reader

gitfool:
&gt; **Draft, and AI-assisted.** Opened as a draft per the project&#39;s policy. As before I have
&gt; measured every claim myself and tried to be as clear about what is *not* measured.
&gt;
&gt; **This is stacked on #3814 and should be read and merged after it.** GitHub needs the base
&gt; branch to live in this repo, so I cannot point the base at #3814 while it is open. That means
&gt; the diff below currently shows seven commits, of which the **first four are #3814** and only
&gt; the **last four are new here**:
&gt;
&gt; - `POLE: add Storage::streamOffset`
&gt; - `Add GetStreamOffsets, a capability query for stream layout order`
&gt; - `PinTable: read table streams in layout order on a single thread`
&gt; - `PinTable: bound live buffer memory before allocation`
&gt;
&gt; Once #3814 merges this will collapse to those four on its own. Happy to rebase, retarget or
&gt; squash it differently if you would rather review it another way.

This is the follow-up promised in #3814, and it is the larger half of the win.

Streams in a `.vpx` are stored close to the reverse of the order the loader reads them. On
Fish Tales 93.9% of stream reads move backwards through the file. This reads them in ascending
file offset instead, on one thread, and parses them on the pool as before.

## Summary

Two changes, and which one pays depends on the table:

- **Order by physical offset.** The loader walked `GameItem0..n`, `Sound0..n`, `Image0..n`,
  which is close to reverse layout order, so every read seeked backwards and defeated readahead.
- **Read on one thread.** #3814 already established that reading the container from twelve
  threads at once is 3.4x slower than from one, because concurrent readers at scattered offsets
  defeat readahead. Parsing stays on the pool, so the CPU work is still parallel.

Cold, over SMB on WiFi. Medians of 3 interleaved repetitions, `purge` before every load, AC
power. `extract` is the stream read and parse phase, `startup` is all of
`LoadGameFromFilename`.

| | master | #3814 | this PR |
| --- | --- | --- | --- |
| Fish Tales, extract (428 MB, 1860 streams) | 39.21s | 18.90s | **3.70s** |
| Fish Tales, startup | 45.91s | 25.60s | **10.41s** |
| Dark Chaos, extract (312 MB, 3555 streams) | 33.06s | 10.86s | **3.98s** |
| Dark Chaos, startup | 40.43s | 18.27s | **11.44s** |

On top of #3814 that is 5.10x on Fish Tales and 2.73x on Dark Chaos for the extract phase,
2.46x and 1.60x for the whole startup phase. Against master, 10.6x and 8.3x on extract, 4.4x
and 3.5x on startup.

Two notes on reading that table. The 39.21s here and the 53.4s in #3814 are not the same
measurement: #3814 quoted a standalone harness reading container streams only, this is the real
application and a different phase boundary. And #3814 turns out to help Dark Chaos *more* than
Fish Tales, 3.04x against 2.07x, so its own body understates it for stream-heavy tables.

## Why both changes are needed

The two effects are close to independent, and neither is sufficient on its own. Isolating them
in the harness, on the same two tables:

| | Fish Tales | Dark Chaos |
| --- | --- | --- |
| serialising alone | 1.24x, noise | **4.22x** |
| ordering alone, on top | **4.89x** | 1.00x, nothing |
| both | 6.20x | 4.11x |

Dark Chaos is already 94.6% forward-laid-out, so ordering has nothing to fix there and the win
is all serialisation. Fish Tales is the opposite. I picked Dark Chaos deliberately as the table
where ordering should do nothing, and it did nothing, which is the result I wanted from it.

Worth being explicit about a limit: **the application numbers in the first table cannot
attribute the gain between the two changes**, because both are in one commit. The split above
comes from the harness. If you would rather see it split into two commits so the attribution is
in the history, say so and I will.

Scanning 688 tables, 686 need the ordering and 2 are already forward. Where it is not needed it
costs the offset lookups and nothing else, measured at -2.7% on Dark Chaos, within noise.

## The commits

**`POLE: add Storage::streamOffset`**
Returns the absolute byte position of a named stream. The subtlety is mini streams: a stream
under 4096 bytes lives in 64-byte mini-blocks packed inside the mini-stream container, so its
block index is in a different coordinate space from a big stream&#39;s. Mixing the two silently
produced a sort order that interleaved the two groups wrongly. Resolving mini-block indices
through the mini-stream container&#39;s own chain before converting to an absolute offset was worth
19.5% on its own. Offsets are therefore comparable across every stream in the container.

**`Add GetStreamOffsets, a capability query for stream layout order`**
Returns false when the storage cannot report positions, which is the case for Windows OLE:
`IStorage` exposes no equivalent. Callers keep declaration order in that case, so this is a
hint and never a requirement.

It is a capability query rather than a platform check on purpose, so there is one code path,
and so that moving Windows onto POLE later would start reporting positions with no caller
change. The `#ifdef __STANDALONE__` inside it is unavoidable, since `PoleStorage` only exists
in standalone builds, but it is confined to this one function and the loader above it has no
conditional.

It uses `dynamic_cast` rather than `QueryInterface`. `FastIStorage` also implements `IStorage`,
is used for undo and the editor clipboard, and has no notion of a physical position, so the
cast has to distinguish them. `QueryInterface` cannot: `PoleStorage` returns `E_NOTIMPL` and
`FastIStorage` returns `S_OK` for every IID it is asked about, so it would claim the capability
falsely.

**`PinTable: read table streams in layout order on a single thread`**
The loader builds a list of every stream it wants, sorts by offset with a stable sort so
declaration order survives as a tiebreak, then reads each one on the main thread and hands the
bytes to the pool. Workers parse from a new `MemoryIStream`, a read-only `IStream` over a
caller-owned buffer, stack allocated for the duration of the parse, with only `Read`, `Seek` and
`Stat` implemented because that is all `BiffReader` and `VPX::Sound::CreateFromStream` use.

I added that rather than reuse `FastIStream`, which is write-then-read, owns its buffer, and has
a stub `Stat`.

A short read now logs and leaves the entry null, which the existing duplicate and failure
handling already copes with, exactly as it did when a stream failed to open.

**`PinTable: bound live buffer memory before allocation`**
Moves the `bytesInFlight` cap check before the buffer allocation and accounts for the incoming
stream size in the comparison. Previously the producer read the next stream unconditionally and
only checked the cap afterward, so two large buffers could coexist (the one being parsed plus
the one just read). Now the producer waits until `bytesInFlight + streamSize <= cap` (or the
queue is fully drained, to avoid deadlock on streams larger than the cap). This guarantees that
a stream exceeding the cap runs alone: nothing else is in flight while it is being parsed, and
nothing new is allocated until it finishes.

Measured: Fish Tales peak RSS during extraction drops from 810 MB to 773 MB (-37 MB) with no
change to extraction time (0.21s vs 0.21s).

## Memory

This holds per-stream buffers, which master does not, so it is the first arm in this series with
a real memory question. Fish Tales, warm, AC, like-for-like `ps -o rss` peak during extraction,
medians of 3 repetitions:

| build | peak RSS during extraction | delta vs master |
| --- | --- | --- |
| master | 660 MB | — |
| #3814 | 635 MB | **-25 MB** |
| this PR | 773 MB | **+113 MB** |

**What that 113 MB actually is.** Profiled with `vmmap` and `heap` snapshots at the extraction
phase boundary:

- **Live heap delta:** +14.8 MB. Only 15 MB of the RSS difference is live allocations. Nearly
  all of it is the single stream buffer currently being parsed by a worker.
- **Allocator page retention:** +139 MB. macOS libmalloc retains freed large-allocation pages
  in `MALLOC_LARGE (empty)` rather than returning them to the kernel immediately. The 1800+
  stream buffers churn through the allocator, and each freed buffer leaves its pages mapped.
  This inflates RSS but is reclaimable under memory pressure. The kernel reclaims these pages
  before killing the process.

**The cap now provides a real bound.** `bytesInFlight` tracks total live buffer memory from
the moment of allocation to the end of parsing. The maximum live buffer memory is
`max(maxBytesInFlight, largestStream)`. For 89.5% of tables in a 686-table survey the largest
stream is under 32 MB, so the cap is effective. For the remaining 10.5%, the largest stream
(up to 210 MB in the worst case, Flash Gordon) runs alone with nothing else in flight.

**Process peak is unchanged.** The load phase peaks at 773 MB against a process peak driven by
texture decode (8+ GB on macOS unified memory). The load phase cannot move the process peak.

**#3814 reduces load-phase memory by 25 MB**, worth noting in its favour.

## How it was verified

**Correctness on every timed run.** All 18 cold loads in the table above were checked, not a
sample. Fish Tales produced content fingerprint `cad268adfcc5` with 0 errors on every build.
Dark Chaos produced `33babd9c056a` with 1 error on every build **including unmodified master**,
so that error is the table&#39;s own VBScript, a `Scripting.Dictionary` duplicate key at line 55248,
and not a regression. The fingerprint covers content-derived load output such as duplicate-name
renames, layer renames and image downsizing, so a load that produced different table structure
would show up here.

**ThreadSanitizer.** Instrumented build, full table load, clean in everything this PR touches:
`MemoryIStream`, `GetStreamOffsets`, `streamOffset`, the in-flight accounting,
`LoadGameFromFilename` and `fileio.cpp`. The only remaining reports in `pintable.cpp` are six
hits on `PinTable::PlaySound`, which is script audio and untouched here. The pre-existing races
from #3814 are still there, listed below.

**Methodology.** As #3814: medians of 3 interleaved repetitions, `purge` per arm, AC power, and
any difference smaller than the largest per-arm spread treated as noise. Per-arm spread here was
1.6% to 8.3% against ratios of 3.5x and up. The interleaving is not ceremony, this link drifts
by tens of percent and single-shot measurements led me to two wrong conclusions earlier in this
work.

## What is *not* verified

- **Mobile.** Not measured. The live buffer cost is bounded at `max(32 MB, largestStream)` and
  the allocator-retained pages are reclaimable under pressure, so this should not cause OOM on
  constrained devices unless the table itself contains an unreasonably large stream.
- **Windows.** Not run on Windows, but no longer just an assumption, see the section below.
  `GetStreamOffsets` returns false on OLE, so the sort is skipped there. The single reader is
  not conditional though, so that half does apply: master calls `IStorage::OpenStream` from pool
  threads, and this moves every storage call onto one thread while parsing stays on the pool.
  That should also be safer, since storage returned by `StgOpenStorage` is not free-threaded.
- **Local disk.** Now measured, warm and cold, and the numbers are in the Windows section above.
  Cold local SSD gives 12.35x and 23.58x on the extract phase and 1.47x and 1.73x on total
  startup. Warm gives nothing beyond #3814, correctly: with the payload already in page cache
  there is no I/O left for a seek-order fix to improve.
- **The 32 MB cap is chosen, not tuned.** I measured 1 MB and 32 MB. Anything between 32 MB and
  the whole payload is unexplored.
- **Attribution in the application.** As above, the harness splits serialising from ordering,
  the application measurement does not.

## What Windows gets from this, measured

I could not test on Windows, so I built the Windows configuration instead: the same branch with
`GetStreamOffsets` forced onto the OLE path, which gives serialised reads and no reordering,
exactly what a Windows build receives. It differs from the real branch by one preprocessor
branch, which also makes it the cleanest isolation of reordering I have managed.

Cold, local SSD, medians of 3 interleaved repetitions, `purge` before each of 24 loads.

| | Fish Tales | Dark Chaos |
| --- | --- | --- |
| master | 2.544s | 4.574s |
| this PR&#39;s parent, #3814 | 0.439s | 0.374s |
| **this PR as Windows receives it** | **0.379s** | **0.171s** |
| this PR with POLE offsets | 0.206s | 0.194s |

**Serialising the reader without reordering does not regress; it is 1.16x and 2.19x faster than
#3814.** So the half of this PR that reaches Windows is a win on its own, and the concern I had
about losing the twelve pool readers there was unfounded. Dark Chaos&#39;s #3814 arm had a 55.6%
spread from one slow run, but the conclusion survives comparing against its fastest run too,
0.365s against 0.171s.

Isolating the reordering itself, it is worth 1.84x on Fish Tales and costs 12% on Dark Chaos.
That is the expected shape: Fish Tales is 93.9% backward so there is a lot to fix, Dark Chaos is
already 94.6% forward so there is nothing to fix and only the offset lookups to pay for. In
absolute terms the cost is 23 ms. I had previously written that ordering never hurts, based on a
harness measurement that put it within noise. Isolated properly it does hurt slightly on an
already-forward table, and I would rather correct that than leave it standing.

That also suggests a cheaper way to get Windows the remaining 1.84x than the storage migration
discussed at the end of #3814. `pole.cpp` already compiles on Windows as part of the FlexDMD
plugin, it is simply not linked into the main binary, since it sits in `VPX_STANDALONE_SOURCES`.
`GetStreamOffsets` could open the `.vpx` read-only with POLE purely to read the directory and
return positions, while every actual read stays on OLE. No write path, no `CreateStream`, no
`STGM_TRANSACTED` crash-safety to reimplement, and it already falls back to declaration order if
the parse fails. I have not built that, and it belongs in its own PR, but it looks like most of
the benefit for very little of the risk.

## Follow-ups

1. **Coalescing reads across adjacent streams.** This still issues one request per stream, 1860
   of them, where #3767 did a single sequential pass. That is why #3767 is still faster on the
   raw container read, 1.39x on Fish Tales and 2.59x on Dark Chaos, at the cost of holding the
   whole file. Merging adjacent stream reads would close some of that gap without the memory.
   Block-level coalescing is already in #3814 and measured 1.15x on its own, so this is the
   stream-level case.
2. **The buffering trade, if anyone wants it.** `maxBytesInFlight` is one constant. Raising it
   buys less producer stalling for more memory, and could be made conditional on platform or
   free memory. Worth saying that this makes #3767&#39;s approach a tuning decision rather than an
   architectural fork, so it can be argued separately from any of this.
3. **The pre-existing races.** Still untriaged, concentrated in `player.cpp`,
   `RenderDevice.cpp`, `Renderer.cpp`, `PerfUI::RenderStats`, `FrameProfiler::Reset` and
   `HitBall::UpdateVelocities`. Happy to open an issue with the report rather than let it sit in
   a PR description.
4. **The write side**, which is where the reverse layout comes from in the first place. Covered
   at the end of #3814: VPX already writes streams in ascending order and the OLE transacted
   commit relocates them, so moving writes onto POLE would likely fix this at source and make
   the ordering here unnecessary on Windows. Untestable while `PoleStorage::CreateStream` is
   `E_NOTIMPL`.

### DIFF
--- a/src/parts/pintable.cpp
+++ b/src/parts/pintable.cpp
@@ -4,6 +4,7 @@
 #include "pintable.h"
 
 #include <algorithm>
+#include <atomic>
 #include <fstream>
 #include <sstream>
 
@@ -1569,7 +1570,7 @@ HRESULT PinTable::LoadGameFromFilename(const std::filesystem::path &filename, VP
             ThreadPool pool(g_app->GetLogicalNumberOfProcessors());
             vector<IEditable *> parts;
             parts.resize(csubobj);
-            int nLoadedParts = 0;
+            std::atomic<int> nLoadedParts { 0 };
             for (int i = 0; i < csubobj; i++)
             {
                pool.enqueue(
@@ -1606,7 +1607,7 @@ HRESULT PinTable::LoadGameFromFilename(const std::filesystem::path &filename, VP
 
             assert(m_vsound.empty());
             m_vsound.resize(csounds);
-            int nLoadedSounds = 0;
+            std::atomic<int> nLoadedSounds { 0 };
             for (int i = 0; i < csounds; i++)
             {
                pool.enqueue(
@@ -1630,7 +1631,7 @@ HRESULT PinTable::LoadGameFromFilename(const std::filesystem::path &filename, VP
 
             assert(m_vimage.empty());
             m_vimage.resize(ctextures);
-            int nLoadedImages = 0;
+            std::atomic<int> nLoadedImages { 0 };
             for (int i = 0; i < ctextures; i++)
             {
                pool.enqueue(
--- a/src/parts/pintable.cpp
+++ b/src/parts/pintable.cpp
@@ -1567,94 +1567,141 @@ HRESULT PinTable::LoadGameFromFilename(const std::filesystem::path &filename, VP
                }
             }
 
+            // Read the container on this thread only, visiting streams in the order they are
+            // physically laid out, and parse them on the pool. Reading from several threads at
+            // once measured 3.4x slower, and declaration order costs a further 4.9x because
+            // streams are stored close to the reverse of the order they are read in.
             ThreadPool pool(g_app->GetLogicalNumberOfProcessors());
             vector<IEditable *> parts;
             parts.resize(csubobj);
             std::atomic<int> nLoadedParts { 0 };
-            for (int i = 0; i < csubobj; i++)
-            {
-               pool.enqueue(
-                  [i, &feedback, &parts, loadfileversion, pstgData, hch, hkey, this, &nLoadedParts, csubobj]
-                  {
-                     const wstring wStmName = L"GameItem" + std::to_wstring(i);
-
-                     IStream *pstmItem;
-                     HRESULT hr;
-                     if (FAILED(hr = pstgData->OpenStream(wStmName.c_str(), nullptr, STGM_DIRECT | STGM_READ | STGM_SHARE_EXCLUSIVE, 0, &pstmItem)))
-                        return hr;
-
-                     ULONG read;
-                     ItemTypeEnum type;
-                     pstmItem->Read(&type, sizeof(int), &read);
-
-                     IEditable *const piedit = EditableRegistry::Create(type);
-                     if (piedit == nullptr)
-                        return E_FAIL;
-
-                     piedit->m_onLoadExpectedPartGroup.clear();
-                     BiffReader reader(pstmItem, loadfileversion, (loadfileversion < 1000) ? hch : NULL, (loadfileversion < 1000) ? hkey : NULL); // 1000 (VP10 beta) removed the encryption //!! NO_ENCRYPTION_FORMAT_VERSION?
-                     piedit->Load(reader);
-                     pstmItem->Release();
-                     pstmItem = nullptr;
-                     if (reader.HasError())
-                        return E_FAIL;
-
-                     parts[i] = piedit;
-                     nLoadedParts++;
-                     return S_OK;
-                  });
-            }
-
             assert(m_vsound.empty());
             m_vsound.resize(csounds);
             std::atomic<int> nLoadedSounds { 0 };
-            for (int i = 0; i < csounds; i++)
-            {
-               pool.enqueue(
-                  [i, &feedback, loadfileversion, pstgData, this, &nLoadedSounds, csounds]
-                  {
-                     const wstring wStmName = L"Sound" + std::to_wstring(i);
-
-                     IStream *pstmItem;
-                     HRESULT hr;
-                     if (FAILED(hr = pstgData->OpenStream(wStmName.c_str(), nullptr, STGM_DIRECT | STGM_READ | STGM_SHARE_EXCLUSIVE, 0, &pstmItem)))
-                        return hr;
-
-                     VPX::Sound *pps = VPX::Sound::CreateFromStream(pstmItem, loadfileversion);
-                     pstmItem->Release();
-                     pstmItem = nullptr;
-                     m_vsound[i] = pps;
-                     nLoadedSounds++;
-                     return hr;
-                  });
-            }
-
             assert(m_vimage.empty());
             m_vimage.resize(ctextures);
             std::atomic<int> nLoadedImages { 0 };
+
+            enum class LoadKind { GameItem, Sound, Image };
+            struct LoadItem { LoadKind kind; int index; wstring name; uint64_t offset; };
+
+            vector<LoadItem> loadItems;
+            loadItems.reserve((size_t)csubobj + csounds + ctextures);
+            for (int i = 0; i < csubobj; i++)
+               loadItems.push_back({ LoadKind::GameItem, i, L"GameItem" + std::to_wstring(i), 0 });
+            for (int i = 0; i < csounds; i++)
+               loadItems.push_back({ LoadKind::Sound, i, L"Sound" + std::to_wstring(i), 0 });
             for (int i = 0; i < ctextures; i++)
+               loadItems.push_back({ LoadKind::Image, i, L"Image" + std::to_wstring(i), 0 });
+
+            // Layout order is a hint. Windows OLE structured storage cannot report where it put
+            // a stream, in which case declaration order is kept.
             {
+               vector<wstring> names;
+               names.reserve(loadItems.size());
+               for (const LoadItem &item : loadItems)
+                  names.push_back(item.name);
+               vector<uint64_t> offsets;
+               if (GetStreamOffsets(pstgData, names, offsets) && offsets.size() == loadItems.size())
+               {
+                  for (size_t k = 0; k < loadItems.size(); k++)
+                     loadItems[k].offset = offsets[k];
+                  std::ranges::stable_sort(loadItems, [](const LoadItem &a, const LoadItem &b) { return a.offset < b.offset; });
+               }
+            }
+
+            // Cap the bytes handed to the pool but not yet parsed. ThreadPool's own queue limit
+            // is 100000 entries, far more than any table has streams, so without this the queue
+            // could hold the whole payload. Measured, this bounds throughput more than memory:
+            // dropping it to 1 MB cost 1.2s of startup and saved 10 MB, because what is resident
+            // is mostly the buffers being parsed right now, one per pool thread.
+            constexpr size_t maxBytesInFlight = 32 * 1024 * 1024;
+            std::atomic<size_t> bytesInFlight { 0 };
+            const int totalToLoad = csubobj + csounds + ctextures;
+
+            for (const LoadItem &item : loadItems)
+            {
+               IStream *pstmItem;
+               if (FAILED(pstgData->OpenStream(item.name.c_str(), nullptr, STGM_DIRECT | STGM_READ | STGM_SHARE_EXCLUSIVE, 0, &pstmItem)))
+                  continue;
+
+               STATSTG ss;
+               ss.cbSize.QuadPart = 0;
+               pstmItem->Stat(&ss, STATFLAG_NONAME);
+               const size_t streamSize = (size_t)ss.cbSize.QuadPart;
+
+               auto buffer = std::make_shared<vector<uint8_t>>(streamSize);
+               ULONG read = 0;
+               if (streamSize != 0)
+                  pstmItem->Read(buffer->data(), (ULONG)streamSize, &read);
+               pstmItem->Release();
+               pstmItem = nullptr;
+
+               // A short read leaves the entry null, which the duplicate and failure handling
+               // below already copes with, as it did when a stream failed to open.
+               if (read != streamSize)
+               {
+                  PLOGE << "Short read on " << MakeString(item.name);
+                  continue;
+               }
+
+               while (bytesInFlight.load(std::memory_order_relaxed) > maxBytesInFlight)
+               {
+                  SDL_Delay(1);
+                  feedback.LoadingProgressUpdated(nLoadedParts + nLoadedSounds + nLoadedImages, totalToLoad);
+               }
+               bytesInFlight.fetch_add(streamSize, std::memory_order_relaxed);
+
                pool.enqueue(
-                  [i, loadfileversion, pstgData, this, &nLoadedImages, ctextures]
+                  [item, buffer, streamSize, &bytesInFlight, &parts, &nLoadedParts, &nLoadedSounds, &nLoadedImages, loadfileversion, hch, hkey, this]
                   {
-                     const wstring wStmName = L"Image" + std::to_wstring(i);
+                     MemoryIStream stream(buffer->data(), buffer->size());
+                     HRESULT hr = S_OK;
 
-                     IStream *pstmItem;
-                     HRESULT hr;
-                     if (FAILED(hr = pstgData->OpenStream(wStmName.c_str(), nullptr, STGM_DIRECT | STGM_READ | STGM_SHARE_EXCLUSIVE, 0, &pstmItem)))
-                        return hr;
+                     switch (item.kind)
+                     {
+                     case LoadKind::GameItem:
+                     {
+                        ULONG read;
+                        ItemTypeEnum type;
+                        stream.Read(&type, sizeof(int), &read);
+                        IEditable *const piedit = EditableRegistry::Create(type);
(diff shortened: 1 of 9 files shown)


toxieainc:
In principle also okay with me, but as mentioned in the other PR: is there a way to have a memory efficient (=no pre-read) path, too (without making things too complicated, of course!)?

toxieainc:
Oh, just saw your other PR..

gitfool:
**Memory follow-up on the load phase.**

I profiled this more carefully and the original "126 MB" figure was both wrong and misleading. Here's what's actually happening.

**The RSS delta is mostly dead pages, not live allocations.** Using `vmmap` and `heap` to snapshot both builds at the extraction phase boundary: the live heap delta between master and this PR is only +15 MB. The other ~130 MB is macOS libmalloc retaining freed pages in its large-allocation cache (`MALLOC_LARGE (empty)`). These pages are reclaimable under memory pressure. The kernel evicts them before killing the process, so they don't represent real cost on constrained devices.

**I tightened the cap so it actually bounds live memory.** Previously the producer read the next stream buffer before checking `bytesInFlight`, so two large buffers could coexist. Now it waits until `bytesInFlight + streamSize <= cap` before allocating, with an escape for streams larger than the cap (they wait until the queue is fully drained, then run alone). This guarantees peak live buffer memory is `max(32 MB, largestStreamInTable)`. No performance change (0.21s extraction vs 0.21s before).

**The real memory bound depends on the table's largest stream.** To understand how that plays out in practice, I surveyed 686 tables:

| largest stream per table | tables | % |
| --- | --- | --- |
| <= 8 MB | 222 | 32% |
| 8-16 MB | 208 | 30% |
| 16-32 MB | 184 | 27% |
| 32-64 MB | 59 | 9% |
| 64+ MB | 13 | 2% |

Percentiles: median 12 MB, p75 21 MB, p90 32 MB, p95 42 MB.

For 90% of tables the 32 MB cap is effective and live buffer memory stays within it. The 10% that exceed it are already large tables (median total payload 295 MB, minimum 93 MB) unlikely to run on a memory-constrained device without rework. The extreme outliers (Flash Gordon 210 MB, Zodiac 124 MB) have that single stream because the table author embedded a very large uncompressed asset. Any loader has to hold that in memory to decode it.

**Summary of where things stand:**

- Live buffer cost is bounded at max(32 MB, largest stream). Typically 12 MB.
- RSS overhead beyond that is allocator page caching, transient, reclaimable under pressure.
- #3814 on its own reduces load-phase memory by 25 MB vs master.
- A no-pre-read path is possible but would forfeit the extraction speedup, since that speedup comes from replacing thousands of per-field IStream calls with a single sequential read per stream.
- Process peak is unchanged (texture decode dominates, well after extraction).

vbousquet:
I just merged some commits regarding this issue. The main changes I made are:
- merge your "Add streamOffset to POLE" commit
- moved the loading path to POLE library (so all platform follow the same loading path)
- perform all loading as a task list, ordering the task list by stream position

This covers a large part of these 2 PR (but not the internal POLE library optimization). So far, I tried ot keep the change inside POLE library minimal. Performance seems ok on my end (but my setup is fairly different). It would be interesting to see how the perf compares so far and if it is needed to go further.

gitfool:
Thanks for merging the streamOffset commit and moving all platforms onto POLE with the ordered task list. One code path for every platform is a nicer shape than what I had, and folding the ordering into a single sorted queue is clean.

I benchmarked current master (`193928030`) against a small POLE follow-up. Extract phase (`PinTable Data loaded` to the images/sounds/fonts/parts marker), 3 interleaved reps, AC power, medians. Same harness as the earlier numbers on this PR.

**Cold, over SMB (WiFi), `purge` before every load:**

| build | Fish Tales | Dark Chaos | Airborne |
| --- | --- | --- | --- |
| master `193928030` (current) | 137.0s | 99.3s | 73.5s |
| current + POLE read coalescing | 20.1s | 16.8s | 11.1s |

**Warm, local SSD:**

| build | Fish Tales | Dark Chaos | Airborne |
| --- | --- | --- | --- |
| master `193928030` (current) | 2.9s | 5.5s | 4.6s |
| current + POLE read coalescing | 0.20s | 0.37s | 0.16s |

Over SMB the current per-block reads cost a network round trip each, so a stream-heavy table takes a while. `StreamIO::read` issues one `seekg` plus one `read` per block, with no coalescing of contiguous blocks and no read-ahead on the bulk path. `.vpx` chains are almost entirely contiguous, so there's a lot to collapse. The two commits are already on this PR's branch:

- coalesce block reads and add a per-stream read-ahead window, gitfool/vpinball@3f10c176c
- stop re-reading the same container block for every mini-block, gitfool/vpinball@e1eeb76f5

They cut cold-SMB load about 6-7x and warm-local about 15x. Both are contained in POLE, need no change to your loader, cherry-pick cleanly onto current master, and produce identical content fingerprints on all three tables.

One thing the SMB numbers make clear: even coalesced, reads still serialise on POLE's read mutex, so the remaining time is one network round trip per read shared across all workers. That's the next thing worth looking at (finer-grained locking or per-stream file handles), but it's separate from these two commits.

Unrelated, but while benchmarking I hit the VNI shutdown crash again (#3849) with a VNI-colorized table. It reproduces on current master, so the recent VNI change didn't touch that path. Worked around locally to keep the harness stable.

*Edit: corrected the cold table above. The figures I first posted there were cold local-SSD, not SMB, mislabeled on my end. The real SMB numbers are higher in absolute terms but the speedup ratio is about the same.*

gitfool:
I re-tested the pre-rewrite master (`42ec8999f`) over SMB to be sure the comparison was fair, and it confirms current master has a network regression worth flagging.

**Cold, over SMB (WiFi), `purge` before every load, extract phase, medians of 3:**

| build | Fish Tales | Dark Chaos | Airborne |
| --- | --- | --- | --- |
| master `42ec8999f` (pre-rewrite) | 38.1s | 32.1s | 21.4s |
| master `193928030` (current) | 137.0s | 99.3s | 73.5s |
| current + POLE read coalescing | 20.1s | 16.8s | 11.1s |

So over SMB, current master is about 3-3.6x slower than before the rewrite. The offset-ordered task list is a clear win, but moving reads onto POLE's block-by-block path costs more than it gives back on a network share: each block is its own `seekg`+`read`, so a stream is N round trips instead of one. The old Win32 path read each stream in fewer, larger chunks, which hid that.

The two POLE commits on this branch fix it and then some, they land under the pre-rewrite master too (20 vs 38s on Fish Tales), because coalescing contiguous blocks plus a read-ahead window turns those N round trips back into roughly one per stream.

- coalesce block reads and add a per-stream read-ahead window, gitfool/vpinball@3f10c176c
- stop re-reading the same container block for every mini-block, gitfool/vpinball@e1eeb76f5

One caveat: even with these, cold SMB is still 11-20s, because reads serialise on POLE's single read mutex, so it's one network round trip at a time across all workers. Worth looking at separately (finer-grained locking or per-stream handles), but the two commits above are the part that recovers the regression.

(For reference, warm local SSD is unaffected by the network cost: current master 2.9 / 5.5 / 4.6s, with the commits 0.20 / 0.37 / 0.16s.)

vbousquet:

Issue #2942, opened 2025-12-02
Standalone: Please enable OpenXR in Linux build


andy-d1969:
I have a Quest 3 connected to my Linux Desktop via ALVR and I already play Zaccaria VR and PinballFX VR with it. I can also play the Windows version of VPX with WINE in VR, but only the 32-Bit version works. Most newer tables need 64-Bit support, so please enable OpenXR in the Linux build. Thanks in advance and thanks for all the good work.

cadecomposer:
Seconded, would be nice to see this feature. In the meantime I suppose I will try to experiment with Wine/Proton

omigeot:
Especially with the upcoming Valve Frame, but in any case it would be interesting to test it. At least if it's in a testable(-ish) state.

PR #3837, opened 2026-08-26
Linux: enable OpenXR compilation for standalone BGFX.

omigeot:
Warning : this is AI-based. I was just willing to see what was missing for openxr support on Linux, and I was very pleased to see very little was actually missing.

Build the OpenXR loader in linux-x64/aarch64 external.sh, gate ENABLE_XR on Linux BGFX, and teach VRDevice/XRVulkanBackend about BX_PLATFORM_LINUX.

Wine's basetsd.h / minwindef.h define _WIN64 on 64-bit Unix, which made bx/platform.h report BX_PLATFORM_WINDOWS (and pull in the D3D headers). VRDevice.h now wraps the bx/platform.h include with the same push_macro / undef _WIN64 / pop_macro guard already used in Shader.h and typedefs3D.h, so bx detects Linux and Android on its own and the previous Android override is dropped too. The disabled preview-window path in XRVulkanBackend.h uses the Vulkan WSI extension name literals directly instead of pulling in the xlib / xcb / wayland Vulkan headers.

Testing instructions : 
* run platforms/linux-x64/external.sh
* cmake -DCMAKE_BUILD_TYPE=Release -B build
* cmake --build build -- -j$(nproc)
* ensure your VR runtime is running (if using wivrn like me, make sure the wivrn client is connected)
* export XR_RUNTIME_JSON="$HOME/.config/openxr/1/active_runtime.json"
* run VPinballX_BGFX with a VR-compatible ini file and table

### DIFF
--- a/.github/workflows/vpinball.yml
+++ b/.github/workflows/vpinball.yml
@@ -165,7 +165,7 @@ jobs:
       - if: (matrix.platform == 'linux')
         run: |
           sudo apt-get update
-          sudo apt install cmake nasm bison zlib1g-dev libdrm-dev libgbm-dev libglu1-mesa-dev libegl-dev libgl1-mesa-dev libwayland-dev libwayland-egl-backend-dev libudev-dev libx11-dev libxcursor-dev libxi-dev libxss-dev libxtst-dev libxkbcommon-dev libxrandr-dev libasound2-dev libpipewire-0.3-dev
+          sudo apt install cmake nasm bison zlib1g-dev libdrm-dev libgbm-dev libglu1-mesa-dev libegl-dev libgl1-mesa-dev libwayland-dev libwayland-egl-backend-dev libudev-dev libx11-dev libxcursor-dev libxi-dev libxss-dev libxtst-dev libxkbcommon-dev libxrandr-dev libasound2-dev libpipewire-0.3-dev libvulkan-dev
       - uses: actions/checkout@v6
         with:
           ref: ${{ needs.context.outputs.checkout_ref }}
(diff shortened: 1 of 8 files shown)


omigeot:
(It would be a step towards fixing https://github.com/vpinball/vpinball/issues/2942 )

christian-georg:
I tried this and  got it to work. It already looks much better than Pinball FX VR, but there are some glitches. The playfield, backbox as well as some elements in the VRRoom are completely black. Also when launching the controller and autoconfig popped up the program crashed. And when using any PinMame dependent table it also crashed. I hope someone from the core team can look into this. If someone needs help with debugging I would be willing to help out.

faethon:
I also got this running, and it looks promising! I have visuals and sound, but no form of control whatsoever. Using a HTC Vive (original).

EDIT: No matter what I try, I seem to not get any input from either keyboard or controller... Can any of you who have been able to get this to run point me in the right direction? :)

omigeot:
> EDIT: No matter what I try, I seem to not get any input from either keyboard or controller... Can any of you who have been able to get this to run point me in the right direction? :)

From what I gathered (not sure if it's true for the windows version as well), the desktop VPX window needs to have focus to relay input. Which is  bit confusing, since it's not rendering anything. At least for keyboard input, it's like that. Maybe a game controller wouldn't have that issue and could process inputs while the desktop window is minimized.

Not sure how it helps.

Chickenzilla:
This PR gets you pretty close, but as the others have stated, can't actually play any games.  I only have a Oculus Rift DK2 (which amazingly works at all under Linux with monado and the custom OpenHMD fork), so I don't have any VR controllers/hand sensing, but I hardcoded a forced VR path (no idea what a VR-compatible .ini file looks like, couldn't get it working without hardcoding) and also had Qwen3.8-27B write me up a nasty fix for the input window, attached.
[wayland-vr-preview-input-fix.patch](https://github.com/user-attachments/files/32411082/wayland-vr-preview-input-fix.patch)

Curiously, I can also play PinMAME tables now.  Opening the menu doesn't work/crashes the player, however, and as discussed some tables don't seem to render all their parts correctly (may be a table issue and not an engine issue, this is my first time in VR).

ruetalipmoc:
I should have mentionned I've only tested it on Xorg, not Wayland. Sorry about that.

PR #3866, opened 2026-08-31
Optimize POLE stream reads for faster table loading

gitfool:
> Draft, and AI-assisted. Opened as a draft per the project's policy. Every claim here I measured myself, and I have tried to be clear about what is not measured.

Follow-up to #3817. That PR's ordered task list and cross-platform POLE move are merged. This picks up the two POLE read optimizations mentioned there but not taken, plus one more that turns out to matter more than either: sizing the load thread pool by where the table lives.

Cold over SMB, tables load 23-32x faster than current master. Local loading is unchanged. The only added memory is a small per-stream read-ahead buffer (see below), nowhere near the whole-stream buffering that raised memory concerns before.

## Why

Current master reads each stream on a pool worker, so all workers read at once. On local storage that is a win, the disk does not care about interleaving and the parsing runs in parallel. Over a network share it is a big loss. The client's read-ahead only works on a sequential run of requests, and a dozen workers reading a dozen different regions is not sequential. Every interleaved jump stalls the prefetch and costs a round trip.

Two things follow from that, and this PR does both. Make each stream's own reads contiguous, and stop workers reading different streams at the same time when the table is on a network share.

## The commits

**`POLE: coalesce block reads and add a per-stream readahead window`.** `StreamIO::read` did one `seekg` plus one `read` per block, and `BiffReader` reads a handful of bytes at a time, so a single stream turned into hundreds of block-sized requests. This merges physically contiguous blocks into one request (`.vpx` chains are almost always contiguous) and serves `read` from a 256 KB per-stream window, so a run of small reads no longer triggers a block read each time. The window belongs to the `StreamIO`, so it stays safe under the existing read mutex.

**`POLE: stop re-reading the same block for every mini-block`.** Streams under 4096 bytes live in the mini-stream container, and reading each 64-byte mini-block pulled its whole enclosing 4 KB block again. This keeps the last enclosing block so consecutive mini-blocks in it come from one read.

**`PinTable: size the load thread pool by storage type`.** The loader now picks its pool size from where the table lives. One thread for a network or SMB path, all cores for local storage. Streams are still visited in offset order in both cases (already in current master); the only thing this changes is how many threads read at once. On a network share one reader keeps the requests sequential, which is what the read-ahead needs. On local storage all cores parse in parallel with no interleaving penalty. Detection is `IsNetworkPath()` in `def.cpp`: `MNT_LOCAL` on macOS, filesystem magic on Linux, UNC plus `GetDriveType` on Windows, best-effort and treating anything it cannot classify as local. A new `Editor / LoadThreadPoolMode` setting (Auto, Single Threaded, Multi Threaded) overrides it if the guess is ever wrong.

The only memory this adds is the read-ahead windows from the first two commits: 256 KB per stream being read, so at most one per pool thread. Master's approach of buffering whole streams, and the memory concern raised on #3817 and #3767, does not apply here.

## How the design was chosen

The useful part was ruling out the obvious answers. All measured on a real SMB share, purge before every load so every read is cold, medians of 3.

Reads are not starved by request size. A single-threaded sequential read of a 428 MB table is about 3.2s at ~130 MB/s, and that holds whether the read size is 512 bytes or 1 MB. The macOS SMB client prefetches well, so small sequential reads are already fast.

Concurrency does not help the reads. The same file read with 1, 4, and 12 threads gave 133, 102, and 114 MB/s. More readers is flat to worse, one already saturates the link. So dropping POLE's internal read lock to allow concurrent reads would buy nothing, and I ruled it out.

The cost is the seek pattern, not the reads. If 3.2s of reading sits inside a 20s or 137s load, the rest is stalls. Forcing the current pool to a single thread over SMB took one table from ~20s at 12 threads to ~4.4s at 1 thread, near the sequential floor. That points straight at interleaving between concurrent workers as the dominant cost.

So the fix is to read sequentially over the network, not to read in parallel and not to touch the lock. Coalescing makes each stream's reads contiguous. Single-threaded loading keeps the order across streams monotonic. A producer/consumer split, one reader feeding parallel parsers, would gain a little more but brings back per-stream buffering and its memory cost, so I left it out. Adaptive sizing gets most of the win with no added memory.

## Results

Extract phase (`PinTable Data loaded` to `Images, Sounds, Fonts and Parts loaded`), medians of 3, cold, purge before each load. Same machine and method as the #3817 numbers.

Cold, over SMB (WiFi):

| build | Fish Tales | Dark Chaos | Airborne |
| --- | --- | --- | --- |
| master `42ec8999f` (pre-rewrite) | 38.1s | 32.1s | 21.4s |
| master `193928030` (current) | 137.0s | 99.3s | 73.5s |
| this PR | 4.4s | 4.4s | 2.6s |

Warm, local SSD, where adaptive keeps all cores:

| build | Dark Chaos |
| --- | --- |
| master `193928030` | 5.5s |
| this PR | 0.37s |

The local win comes from the two POLE commits alone; they help every transport. The SMB win is those plus the single reader. Content fingerprints and error counts match across builds on all three tables, so what gets loaded is unchanged.

## Testing

macOS arm64 BGFX, tables on both local SSD and a real SMB NAS. Auto-detection checked both ways: SMB paths load on 1 thread, local paths on all cores, both logged. Both override modes checked. The Windows and Linux detection paths are written but not run here; each falls back to local on any failure, so the worst case is today's behavior.

### DIFF
--- a/src/core/Settings_properties.inl
+++ b/src/core/Settings_properties.inl
@@ -4,6 +4,9 @@ PropString(Version, VPinball, "VPX Version"s, "VPX version that saved this file"
 // General Application settings
 PropBool(Editor, EnableLog, "Enable Log"s, "Enable general logging to the vinball.log file"s, true);
 PropBool(Editor, DisableHash, "Disable File Validation"s, "Disable file integrity validation (risky; but slightly faster loading)"s, false);
+PropEnum(Editor, LoadThreadPoolMode, "Table Load Concurrency"s,
+   "Thread pool size used to read and parse table data during load.\nAuto reads with a single thread for network/SMB paths (keeps reads sequential, far faster over the network) and uses multi threaded for local storage.\nSingle Threaded forces one thread; Multi Threaded forces all cores."s,
+   int, 0, "Auto"s, "Single Threaded"s, "Multi Threaded"s);
 
 // Audio settings
 PropInt(Player, MusicVolume, "Backglass Volume"s, "Main volume for music and sound played from the backglass speakers"s, 0, 100, 100);
--- a/src/core/def.cpp
+++ b/src/core/def.cpp
@@ -8,6 +8,12 @@
 #include <dirent.h>
 #include <sys/stat.h>
 #include "standalone/PoleStorage.h"
+#if defined(__APPLE__)
+#include <sys/param.h>
+#include <sys/mount.h>
+#elif defined(__linux__) || defined(__ANDROID__)
+#include <sys/vfs.h>
+#endif
 #endif
 
 #include "core/VPApp.h"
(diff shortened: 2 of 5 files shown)


vbousquet:
Just merged the 2 POLE optimization commits. I rewrote the commit that selects the number of thread to remove AI comments (useful to review, but excessive for codebase) and to avoid having one more setting (opaque and not exposed to user). The changes give quite a boost to loading time. Thanks a lot!

PR #3871, opened 2026-08-31
Ignore empty display frames in DMDUtil plugin

francisdb:
Exiting a PinMAME table crashed with a SIGSEGV in the DMDUtil plugin's update thread.

When the emulation stops, PinMAME's display source keeps being advertised for a short while but no longer has a frame to hand out. DMDUtil did not account for that and tried to read a frame that was not there. It now ignores empty frames, like the other display consumers do.

Only visible once #3870 is in, which removed an earlier abort in the same teardown sequence.


vbousquet:
The bug is the provider pushing a null frame and must be fixed there. An advertised frame as the contract to actually exist. So the fix is not the right one and would hide a real bug (here in libpinmame)

Issue #3849, opened 2026-08-28
macOS/BGFX: SIGABRT on exit — VNI plugin static destructor throws during __cxa_finalize


gitfool:
**Describe the bug**
On macOS (Apple Silicon, BGFX), VPX aborts with `SIGABRT` when the process exits after playing a table with a VNI-colorized DMD. A destructor throws while the process is already tearing down, so C++ calls `std::terminate()`, which aborts.

Reproducible with Fish Tales, which loads a VNI colorization set, so the VNI plugin is active.

**To Reproduce**
1. Build current `master` and run on macOS (arm64, BGFX).
2. Load a table with a VNI-colorized DMD (e.g. Fish Tales).
3. Quit the app (Cmd+Q / quit event).
4. Process aborts (`Abort trap: 6`) instead of exiting cleanly.

**Root cause**
The crash happens during static-destructor teardown at process `exit()`, not during `Player::~Player()`. The player log has no `Closing player` line, so `~Player` never ran. The quit event goes straight to `-[NSApplication terminate:]` → `exit()` → `__cxa_finalize`.

Faulting (main) thread, top of stack downward:

```
exit
  __cxa_finalize_ranges
  cxa_at_exit_callback_installed_at
  std::unique_ptr<Vni::VNIColorizer>::~unique_ptr()
  Vni::VNIColorizer::~VNIColorizer()
  PinballPlugin::Controller::CtrlItemConsumer<DisplaySrcId>::~CtrlItemConsumer()
  std::terminate() -> abort
```

The owners are two file-scope statics in `plugins/vni/vni.cpp`:

```cpp
static std::unique_ptr<CtrlItemConsumer<ControllerDef>> controllers;
static std::unique_ptr<class VNIColorizer> colorizer;
```

At `exit()`, `__cxa_finalize` destroys these on the main thread. `~CtrlItemConsumer` (in `plugins/plugins/ControllerPlugin.h`) then calls back into the plugin message API: `m_msgApi->UnsubscribeMsg(...)`, `m_onItemsAboutToChange()`, `m_msgApi->ReleaseMsgID(...)`. By the time static destructors run, that subsystem is already gone or invalid, and the constructor recorded a thread id that the destructor asserts against. One of those calls throws, and a throwing destructor on the terminate path aborts.

**Correction to the original report**
[...]

PR #3872, opened 2026-08-31
Fix use-after-destroy in VNI and Serum colorizer teardown

gitfool:
Fixes the macOS SIGABRT on exit reported in #3849, seen after playing a table with a VNI-colorized DMD (e.g. Fish Tales).

## The bug

`VNIColorizer` declares its members in this order:

```cpp
CtrlItemConsumer<DisplaySrcId> m_dmdSource;     // declared first, so destroyed last
CtrlItemProvider<DisplaySrcId> m_colorizedDmd;  // declared second, so destroyed first
...
std::thread m_colorizeThread;
std::vector<uint8_t> m_colorFrame;
```

`m_dmdSource`'s destructor fires its `onItemsAboutToChange` callback, which is `StopColorizeThread()`. That callback joins `m_colorizeThread` and touches `m_colorizedDmd` and the frame state. Because members are destroyed in reverse declaration order, `m_dmdSource` is destroyed last, after everything its callback reads is already gone. The callback then touches freed members, and the throw escapes the `noexcept` destructor, so the runtime calls `std::terminate()`.

The crash shows up two ways: at process `exit()` via `__cxa_finalize` destroying the file-scope statics (the stack in #3849), and during `Player::~Player()` when the table exit script broadcasts a controller change that resets the colorizer. Same defect, both paths end in `~CtrlItemConsumer<DisplaySrcId>` calling `std::terminate()`.

Same class of bug as #3849's `ResURIResolver` fix (98c424c), where consumers were destroyed after the caches their callbacks clear.

## The fix

Move `m_dmdSource` to be the last member in both `VNIColorizer` and `SerumColorizer`, so it is destroyed first while everything its teardown touches is still alive. Init lists reordered to match.

`VNIPluginUnload` also now resets `colorizer` before `controllers` before nulling `msgApi`, so the members that call back into the message API are torn down in order while the API is alive. `SerumPluginUnload` already did this.

Serum has the identical member-order bug, so it gets the same reorder.

## Testing

Built `VNIPlugin` and `SerumPlugin`. Local testing on macOS (arm64, BGFX) no longer reproduces the VNI exit crash.

## Not covered

If the app quits without running `UnloadPlugins()`, the statics are still destroyed at `__cxa_finalize` against a dead `msgApi`. That path did not reproduce for me here, so I left it alone. Closing it fully would mean either running `UnloadPlugins()` on the quit path or making `~CtrlItemConsumer` skip its API calls when the API is gone, which touches the shared template used by every plugin.

### DIFF
--- a/plugins/serum/serum.cpp
+++ b/plugins/serum/serum.cpp
@@ -50,11 +50,11 @@ public:
    SerumColorizer(const std::filesystem::path& serumPath, const string& currentGameId, uint32_t controllerEndpointId)
       : m_pSerum(Serum_Load(serumPath.string().c_str(), currentGameId.c_str(), FLAG_REQUEST_32P_FRAMES | FLAG_REQUEST_64P_FRAMES))
       , m_controllerEndpointId(controllerEndpointId)
+      , m_colorizedDmd(msgApi, endpointId, CTLPI_DISPLAY_GET_SRC_MSG, CTLPI_DISPLAY_ON_SRC_CHG_MSG)
+      , m_colorizedframeId(std_rand())
       , m_dmdSource(
            msgApi, endpointId, CTLPI_DISPLAY_GET_SRC_MSG, CTLPI_DISPLAY_ON_SRC_CHG_MSG, [this](std::vector<DisplaySrcId>& items) { FilterDmdSource(items); },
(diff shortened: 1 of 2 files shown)


vbousquet:
Thanks !

DanRoid:
Many thanks, fixed crashes on table exit in linux BGFX for me.

PR #3882, opened 2026-09-01
Retract advertised displays before releasing display providers

francisdb:
Exiting a table with Serum active crashed in `GetRenderFrameSerumV1`. Serum, VNI, AlphaDMD and UpscaleDMD release their provider object by resetting a global pointer in the controller change hook; the pointer is nulled before the destructor retracts the advertised displays, so a consumer finishing a frame during that broadcast dereferenced a null provider.

The providers now retract their displays first and are released afterwards, so nothing is advertised past the point where it can be served.

### DIFF
--- a/plugins/alphadmd/alphadmd.cpp
+++ b/plugins/alphadmd/alphadmd.cpp
@@ -132,7 +132,9 @@ public:
          .GetIdentifyFrame = &GetIdentifyFrame });
    }
 
-   ~AlphaDMDRenderer()
+   ~AlphaDMDRenderer() { Stop(); }
+
+   void Stop()
    {
       StopRenderThread();
       m_dmdProvider.ClearItems();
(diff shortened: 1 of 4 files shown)


vbousquet:
This was cleanly fixed with the callContext refactor, so closing

francisdb:
@vbousquet prefer me to create issues instead of pr's?

for reference: 5bd6fe54c9f85d35649d03b56dd9d07910789fa0

vbousquet:
PR are fine. Ideally PR should be made against an issue (so open PR+issue), but if the issue is simple enough it can also be just described inside the PR.

Opening only issue is usually less interesting than issue+PR or PR including the issue description in the PR text, as the PR adds a few things:
- a fix proposal may be simply merged and the definitive solution to the problem
- otherwise it may be used as the base for the final fix/refactor (or merged as an intermediate hack fix waiting for a proper refactor)
- an additional way of explaining the issue: issues are usually on the symptoms while fixes are usually on the cause, they are both useful

So, in the end, PR are a really useful way of contributing, like issues. The real problem we are facing is that we have limited time for this hobby project (still multiples commits per day!) and the project has moved a huge lot, including merging some draft code which need refactor more than shallow fixes (for example, more or less all the plugin system that quickly drafted to avoid keeping standalone as a fork, but is being cleaned up atm)

Issue #3874, opened 2026-08-31
RumbleMode: is there a plan for the "N/A yet" modes, and would this approach fit?


kara2010:
**Is your feature request related to a problem? Please describe.**

`Settings_properties.inl` defines `RumbleMode` with four values, two of which
are marked as not yet available:

> Off / Table only (N/A yet) / Generic only (N/A yet) / Table with generic fallback

The generic rumble that exists today is three fixed calls - bumper, slingshot
and plunger - at hardcoded strengths. For desktop players without any feedback
hardware, the gamepad is the only haptic channel there is, and those three
events leave most of what happens on the playfield unfelt.

**Describe the solution you'd like**

I have an implementation of the generic side running on a 10.8.0 fork, and
before porting it to master I would like to know whether it fits what you had
in mind for those two modes.

What it adds, all through the existing `PlayRumble()`:

- flipper rumble on actual ball contact, scaled by impact speed (not on button
  press)
- ball drain, and nudging, scaled by the table velocity change
- plunger release, kicker/scoop capture and eject, drop target, spinner
- ball-to-ball collisions, scaled by closing speed - the one thing that carries
  multiball
- a strength slider per event, persisted in settings, adjustable while playing

The call sites are the obvious ones and all still exist in master:
`InvokeBallBallCollisionCallback` in `hitball.cpp`, `HitSpinner::Collide`,
`Kicker::KickXYZ` and the kicker hit event, `DISPID_TargetEvents_Dropped`, and
the existing plunger call in `hitplunger.cpp`.

**Describe alternatives you've considered**

- Leaving it in a fork. That works, but it means desktop players need a custom
  executable for something the engine could do on its own.
- Doing it through DOF instead. That covers ROM-driven effects well, but not
  these: ball contact, ball-to-ball and nudge happen in the physics engine and
  DOF cannot see them. The two approaches complement each other rather than
  compete.

**Additional context**

My main question is about direction, not code. If you already have a design in
mind for "Table only" and "Generic only" - particularly how a table script would
drive rumble - then I would rather build towards that than submit something that
has to be reshaped afterwards.

If the direction is right, I would split it into small pull requests rather than
one large one, starting with flipper ball contact alone, so the approach can be
judged before the rest follows.

**Disclosure:** this was built with AI assistance, under my direction, and I
would rather say so now than in a pull request footnote. I am not a C++
developer - my background is web development and scripting, some years back -
so I cannot claim I could defend every line to a reviewer. What I can vouch for
[...]

PR #3886, opened 2026-09-02
Rumble on flipper/ball contact, scaled by impact speed

kara2010:
First slice of the generic rumble work discussed in #3874, kept deliberately
small as suggested there.

**What it does:** plays a rumble pulse when a ball collides with a flipper,
scaled by the relative normal velocity of the impact (`bnv` in
`HitFlipper::Collide`). A soft catch barely registers, a hard slap kicks.
This complements the existing fixed-strength calls for bumpers, slingshots
and the plunger, and goes through the existing `PlayRumble()`, so it obeys
`RumbleMode` like they do.

**Changes** (+25 lines, no deletions):
- `Settings_properties.inl`: new `Player/RumbleFlipperContact` float
  (0..3, default 1, 0 disables), right below `RumbleMode`
- `InputManager`: cached member read in the constructor,
  `PlayFlipperContactRumble()` next to `PlayRumble()`, inline accessors
  for the live UI
- `hitflipper.cpp`: one call inside the existing hit-event gate (real hit, 250 ms
  idle), so a ball held on a raised flipper stays silent
- `InputSettingsPage`: slider below the RumbleMode entry, following the
  existing PropId item pattern (framework handles persistence)

**Scaling:** ~17 velocity units map to full strength; both motors are driven
because short high-frequency-only pulses are barely noticeable on many pads.
Values come from several weeks of by-hand testing across a large table
collection on a 10.8.0 fork of this.

**Testing:** built and played on Windows x64 (BGFX). Verified: no rumble with
RumbleMode=Off, strength 0 disables it, slider takes effect immediately,
setting persists across restarts, holding a ball on a raised flipper is silent.

If this shape fits, drain/nudge and the remaining events from #3874 would
follow as equally small PRs.

**Disclosure**, consistent with the linked issue: built with AI assistance
under my direction; tested by hand.

### DIFF
--- a/src/core/Settings_properties.inl
+++ b/src/core/Settings_properties.inl
@@ -202,6 +202,7 @@ PropInt(Player, NumberOfTimesToShowTouchMessage, "NumberOfTimesToShowTouchMessag
 PropBool(Player, Mirror, "Mirror"s, "Mirror the table (left <-> right)"s, false);
 PropEnum(Player, CacheMode, "Cache Mode"s, "Use cache to limit stutters and speedup loading"s, int, 1, "Disabled"s, "Preload Textures"s);
 PropEnum(Player, RumbleMode, "RumbleMode"s, "Use rumble motor(s) in attached input devices"s, int, 3, "Off"s, "Table only (N/A yet)"s, "Generic only (N/A yet)"s, "Table with generic fallback"s);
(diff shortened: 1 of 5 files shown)


kara2010:
Force-pushed a fix: the rumble call now sits inside the existing hit-event gate (real hit + 250 ms idle) instead of before it. Before, a ball held on a raised flipper produced a faint continuous rumble from the resting micro-collisions. Same +25 lines otherwise; retested by hand including the held-ball case.

PR #3892, opened 2026-09-03
docs: add pinmame cfg directory

dekay:
Standalone does not have a UI for changing things like per audio channel volumes or dipswitch settings. However, these *can* be changed in pinmame / xpinmame and the `.cfg` files created as a result can be used by libpinmame in VPX. I've documented the process for doing so on this wiki page: https://github.com/dekay/vpinball-wiki/wiki/Linux#using-pinmame-to-set-individual-channel-volumes-and-dipswitch-settings

### DIFF
--- a/docs/FileLayout.md
+++ b/docs/FileLayout.md
@@ -87,13 +87,17 @@ Table Name (Manufacturer Year)/              <= We created a dedicated folder to
 ├── music/                                   <= Folder from which music are loaded when script use the PlayMusic command
 │   ├── Multiball Theme.ogg
 │   └── ...
-├── pinmame/                                 <= PinMAME plugin will look here for rom, nvram, and alias files
+├── pinmame/                                 <= PinMAME plugin will look here for rom, nvram, config, and alias files
 │   ├── roms/
 │   │   ├── xxx.zip
 │   │   └── yyy.zip
 │   ├── nvram/
 │   │   ├── xxx.nv
 │   │   └── yyy.nv
+│   ├── cfg/
+│   │   ├── default.cfg
+│   │   ├── xxx.cfg
+│   │   └── yyy.cfg
 │   └── alias.txt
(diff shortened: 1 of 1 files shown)


vbousquet:
Thanks!

Note that direct setup through config is a temporary (and bad) way of adjusting the settings. It is indeed needed for the time being but this should be solved quickly with proper LiveUI integration:
- pinmame volume is now already supported and the cfg file should not been used for this
- dipswitches are in the process of being integrated in the LiveUI as well (through TomLogic's mepmap for very broad support)

vbousquet:
Just had a look to https://github.com/dekay/vpinball-wiki/wiki/Linux#using-pinmame-to-set-individual-channel-volumes-and-dipswitch-settings and I'm not sure to understand the use-case regarding adjusting volume through xpinmame. With latest version of VPX & its PinMAME plugin, you have already 3 integrated setups:
- the main volume (all channels on backglass / playfield or both at the same time)
- the per source mixer volume (allow to adjust PinMAME backglass volume, including boosting it)
- the ROM adjustment which allow to adjust the output level like on the real machine

Is there another setting need that I don't get ?

dekay:
@vbousquet Thanks for the review! The xpinmame approach allows for finer grained control than the three options you mention. For the case of Black Knight, the `~` key in xpinmame gains access to the following volume settings

- master Volume settings (now exposed by VPX as you point out, and I agree it shouldn't be changed in the cfg file)
- two YM2151 channels
- two DAC channels
- two HC55536 channels
- six Mech sound channels

This all came about when a Standalone user wanted to boost *just* the music volume on Black Knight, as it was too quiet for his taste relative to all the other pinmame sounds. They were able to do that by boosting (I think) *just* the two YM2151 channels and leaving the rest alone. 

Perhaps in the future tje VPX LiveUI will support this fine grained audio adjustment, but for now this was the only way we could think of to accomplish it. That dipswitch support is coming to the LiveUI is excellent news!

Issue #3909, opened 2026-09-06
BGFX Win64 - Crash On-Start 5575+


DJOMaverick:
Starting with build 5575, BGFX Win64 crashes and terminates immediately upon loading any table.

[vpinball.log](https://github.com/user-attachments/files/31873737/vpinball.log)

DJOMaverick:
[dump.zip](https://github.com/user-attachments/files/31873949/dump.zip)

gitfool:
I loaded the `dump.zip` in WinDbg and read it against the `vpinball.log`. Here's what the crash is.

It's an access violation reading address `0x0D`. The fault is in an inlined `strlen`-style scan at `VPinballX_BGFX64+0x6efbb0`. The routine got a bad pointer, `rcx = 0xFFFFFFFFFFFFFFF3` (that's `-13`), and scanned forward a byte at a time until it hit unmapped memory. Faulting thread is Main, about 7 seconds after the table starts loading.

The log lines up with that timing. It ends at:

```
22:34:49.444 INFO [Main] [MsgPI::MsgPlugin::Load@540] Plugin PinMAME loaded (...plugin-pinmame64.dll) [id=12]
```

No shutdown, no error line, just a hard stop mid-startup. So the process died during play-mode plugin loading, right after PinMAME loaded and before the next plugin finished. That matches "crashes immediately when loading any table."

On the cause: something in plugin init hands a string pointer to `strlen` without checking it, and the pointer holds `-13` (a sentinel or error code) instead of a real `const char*`. In the dump, the caller is VPX startup code at `VPinballX_BGFX64+0xf5197` (then `+0xf5160`), with a `plugin-pinup-events64.dll` frame and an `ntdll!RtlFreeHeap` return address on the stack next to it. Both the source pointer and the target buffer at `0x1cb8dcb0` are unmapped.

The log never mentions pinup-events, though, loaded or disabled. The only thing pointing at it is that stack frame in the dump, and it sits right beside an `RtlFreeHeap` frame, so it might be the plugin that crashed while loading (the one after PinMAME), or it might just be a stale frame rather than the live caller. I can't prove it from the log alone.

Test that would settle it: disable pinup-events in the plugin settings and try the table again. The log shows disabled plugins get skipped before loading, so that keeps its init code from ever running. If it still crashes, disable the other enabled plugins one at a time (AltSound, B2SLegacy, DOF, FlexDMD, PinMAME) until it loads. That pins down the trigger and gives a workaround.

[...]

PR #3913, opened 2026-09-06
build: emit debug symbols for Windows crash dumps

gitfool:
## What

Windows release builds ship without any debug symbols, so a crash dump from a user can't be resolved to function names or source lines. macOS gets away with it because its release binary is left unstripped and keeps its symbol table; the Windows builds emit nothing usable. This adds symbols for both Windows toolchains.

- MSVC (`PLATFORM=windows`, x64 and x86, all renderers including DX9): compile with `/Zi` and link with `/DEBUG`, producing a matching `.pdb`.
- MinGW (`PLATFORM=windows-mingw`, x64): compile Release with `-g1`, then split the symbols into a companion `.debug` with `objcopy` and strip the exe, leaving a `.gnu_debuglink` so a debugger finds the symbols on its own.

The CI staging steps bundle the symbol file next to the exe in the artifact zip, so whoever analyzes a dump has both halves.

Issue #3909 is the kind of report this helps with: a crash on start with only a log to go on. With symbols, the dump resolves to a real call stack.

## Why these choices

MSVC and MinGW carry symbols differently, so the two paths don't match.

MSVC never embeds debug info in the exe. `/DEBUG` writes a separate `.pdb`, and it stays separate by design, so there's no strip step to do. The one trap is that `/DEBUG` turns off `/OPT:REF` and `/OPT:ICF` unless you pass them explicitly. They were already in the link options, so the exe is the same size it was before (22.6 MB on x64). I kept the full `.pdb` rather than a stripped one: MSVC's `/PDBSTRIPPED` drops source line numbers, and at 82 MB the full pdb is small enough that giving up line info isn't worth it.

MinGW (GCC) embeds DWARF in the exe, but only with `-g`, and full `-g` produced a 486 MB `.debug`. `-g1` keeps line tables and function boundaries without per-variable and type data, which is all a backtrace needs, and drops the companion file to 63 MB. The exe is stripped and carries a `.gnu_debuglink`, so it stays the size it was.

## Sizes

| build | exe | symbols |
|-------|-----|---------|
| MSVC x64   | 22.6 MB | `VPinballX_BGFX64.pdb` 82.0 MB |
| MSVC x86   | 21.2 MB | `VPinballX_BGFX.pdb` 81.6 MB |
| MinGW x64  | 34.2 MB | `VPinballX_BGFX64.debug` 62.7 MB |

These are uncompressed. Symbol files compress well, so the zip grows by a good deal less.

## Verified

Built all three locally against VS2026 (MSVC 19.51) and UCRT64 GCC, then resolved a real address in each.

- MSVC x64, in WinDbg: `x VPinballX_BGFX64!HitBall::Contact*` resolves to `HitBall::Contact(CollisionEvent*, float)`. `dumpbin` confirms the exe's CodeView entry matches the pdb, and that `/LTCG` still coexists with `/DEBUG`.
- MSVC x86: `dumpbin /PDBPATH:VERBOSE` reports the pdb found and matched.
- MinGW x64, with `addr2line`: the same address resolves to `HitBall::Contact` at `hitball.h:48`. Without the `.debug` file present, it resolves to `??`, which is what ships today.

## Out of scope

External dependencies (bgfx, SDL3, ffmpeg, and the rest) still ship without symbols, so a frame inside those modules won't resolve. bgfx is the most likely place to want this since it links statically into the exe. That's a larger change (it touches `external.sh` and invalidates the CI dep cache), so I left it for a follow-up.

### DIFF
--- a/.github/workflows/vpinball.yml
+++ b/.github/workflows/vpinball.yml
@@ -289,6 +289,11 @@ jobs:
             else
               cp build/${{ matrix.config }}/VPinballX_${{ matrix.type }}.exe stage
             fi
+            if [[ "${{ matrix.platform }}" == "windows" ]]; then
+              cp build/${{ matrix.config }}/VPinballX_${{ matrix.type }}*.pdb stage
+            else
+              cp build/${{ matrix.config }}/VPinballX_${{ matrix.type }}*.debug stage
+            fi
             cp build/${{ matrix.config }}/*.dll stage
             if [[ "${{ matrix.type }}" == "GL" ]]; then
               cp -r build/${{ matrix.config }}/shaders-${{ needs.version.outputs.version_short }} stage
@@ -391,6 +396,7 @@ jobs:
           else
             cp build/${{ matrix.config }}/VPinballX.exe stage
           fi
+          cp build/${{ matrix.config }}/VPinballX*.pdb stage
           cp build/${{ matrix.config }}/*.dll stage
           cp -r build/${{ matrix.config }}/assets stage
           cp -r build/${{ matrix.config }}/scripts stage
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -288,6 +288,7 @@ if(PLATFORM STREQUAL "windows")
       /w44005
       /Zc:__cplusplus
       /std:c++20
+      /Zi
       $<$<NOT:$<STREQUAL:${ARCH},x64>>:/arch:SSE2>
    )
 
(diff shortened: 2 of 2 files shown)


toxieainc:
@vbousquet What do you think? Might be helpful, and nowadays HDD space and dload size is not that much of an issue anymore..

jsm174:
I have some nitpicks with the styling of these changes, as they don't match anything we've done in any cmake for this project. Reviewing...

gitfool:
Happy to match `if` style.

jsm174:
It wasn't so much that, but all the _var stuff. 

@toxieainc, @vbousquet - Do we even need the debug builds anymore, ie, if we just had the pdb's for windows, it should be good enough right?

gitfool:
I could inline the `_var` stuff and maybe we initially leave the debug builds in the makefiles but remove from ci?

gitfool:
@jsm174 I cherry picked your upstream branch to replace my commit with yours since it's better than mine 😉

gitfool:
As for debug builds, my 2 cents is that they should be kept for local builds only; i.e. removed from ci.

vbousquet:
I agree that the CI should only build release builds, distributing with symbols to get meaningful stacktrace. I'm really not in my skill zone here, so please base on @jsm174 review to merge this nice improvement

PR #3916, opened 2026-09-06
Rumble: mix overlapping pulses, own flipper trigger, measured motor model, plunger strike, nudge and ball collision

kara2010:
Follow-up to #3886. Seven commits, best read in order; each carries the measurement that
motivated it. All pulses were logged from the collision down to `SDL_RumbleJoystick` and its
return value, before and after each step, on an Xbox controller, Windows 11, BGFX/D3D11, with
B2S/DOF removed so nothing else drives the pad. The motor response was measured with an
accelerometer on the pad.

**1. Mix overlapping pulses, drive the motors as measured.** `SDL_RumbleJoystick` cancels
whatever is playing, and `PlayRumble()` forwarded calls as they came, so the last caller won
however weak: over 269 logged calls the plunger release pulse lived 2 ms before the spring
rebounds replaced it (91 % cut short), bumpers 49 %, slingshots 33 %. Pulses are now collected
(eight slots, a mutex since collisions come from the physics thread and buttons from the OS
thread), the device plays the strongest per motor and steps down when it ends. Measured, the
motors do nothing below 0.3, give half at 0.5, full at 0.86, and need 110..190 ms from rest;
so every level is mapped onto the usable range and a step up of the mix gets twice the level
for 80 ms. Rumble mode Off stops what is playing.

**2. Adjustable.** Bumper, slingshot, plunger and the flipper button get a strength setting
each (`Player/Rumble*`, 0..3) with sliders next to the flipper contact one. The plunger stops
passing its signed fire speed (negative on the forward stroke, clamped to silence).

**3. Flipper contact: own trigger.** The pulse from #3886 sat inside the flipper's script
hit-event gate (250 ms since the last contact of any kind). A ball resting on a raised flipper
touches it every few milliseconds, so the slap that follows never qualified: 49 of 70 slaps
lost. Contacts less than 80 ms apart form one sequence: its first contact plays, later ones
only when they exceed the sequence's strongest impact by a unit. The script event is untouched.

**4. Levels that are felt.** Bumper 0.10/0.05 and slingshot 0.15/0.10 for 100 ms are below
what the motors render; now 0.60/0.35 and 0.80/0.50 for 150 ms. The flipper button pulse moves
to `Flipper::RotateToEnd` (no buzz in attract mode) and becomes a solenoid thump on both
motors, kept below the kick level so the ball hit that follows stands out. Default 0.5.

**5. Plunger.** `Fire()` is not the launch: a plunger following an analog stick never calls
it. The strike now plays in `HitPlunger::Collide` when the tip hits the ball, scaled by the
closing speed. Many tables park the tip behind the end of the shooter lane (logged: park
position 1/6 of the stroke, lane wall 8 units in front of the resting tip), so a ball rolling
back lands on that wall and never touches the tip; a hit there, within the plunger width and
less than a ball diameter in front of the resting tip, plays the same contact. One pulse per
landing (a landing bounces 4.4, 1.2, 0.5, 0.5 within 330 ms).

**6. Nudge.** From the cabinet acceleration, 1 m/s² to 0.5 g, 80 ms, 200 ms cooldown.
`Player/RumbleNudge`.

**7. Ball/ball collision.** Next to the script callback in `HitBall::Collide`, 70 ms, mostly
on the high frequency motor. `Player/RumbleBallBall`.

DOF/B2S and the plugin path are not touched; the only output is `SDL_RumbleJoystick`, as before.

Disclosure, as in #3886: built with AI assistance under my direction; tested by hand.


jsm174:
Thanks! Can you rebase this on the latest as I added https://github.com/vpinball/vpinball/pull/3918.

It adopts your previous updates for `libvpinball` on mobile devices (iOS/android).

Can you make sure this all supports `libvpinball` too?

kara2010:
Rebased on master with #3918. The mixed output (strongest pulse per motor, motor floor
and start kick applied) now feeds your new path as well: Android through the SDL handler's
haptic effect, iOS through VPinballLib::PlayRumble, both with the same low/high values the
gamepad gets. Nothing platform specific in the diff; Windows BGFX build clean.

One thing to check on a device: the mixer maps every level onto the range the pad motors
render (0.3 + 0.7 x level) and doubles a step up for 80 ms, and the Android path then
applies cbrt on top. Both curves together may make light touches strong on a phone. If
that is the case I can apply the floor and kick to gamepad outputs only; I have no mobile
device to test it here.

toxieainc:
Looks good to me.. Thanks

jsm174:
I rebased with the latest and updated to support the new XR rumble logic.

PR #3917, opened 2026-09-06
Update libserum, libdmdutil, libzedmd

mkalkbrenner:
Fixed finding the colorization if serum plugin is used by PPUC, not VPX. Also added two missing Serum options.


plugins/plugins/ControllerPlugin.h
```diff
@@ -193,6 +193,7 @@ typedef struct DisplaySrcId
 {
    CtlResId id;                                                             // Unique Id of the display
    CtlResId overrideId;                                                     // If this source overrides another source, id of the overriden source, 0 otherwise
+   uint32_t controllerId;                                                   // Endpoint of the controller this display derives from, 0 if it is the controller's own (i.e. same as id.endpointId)
```
vbousquet: I don't understand why we would need this. The point of overrideId is to define a unique Id that answer the question of what are we overriding. I don't see any use of this new field

plugins/plugins/ControllerPlugin.h
```diff
@@ -214,6 +215,17 @@ typedef struct DisplaySrcId
    DisplayFrame(MSGPIAPI* GetIdentifyFrame)(void* callContext);             // Get the last identify frame. Thread safe. Returned value is not null, owned by the source, in the format defined by identifyFormat
 } DisplaySrcId;
 
+// The controller a display belongs to. A renderer that turns a controller's
+// state into a display -- alphadmd building a DMD out of segment displays -- is
+// not overriding anything, so overrideId cannot say which controller it came
+// from, and on a segment-only machine there is no parent display to point at
+// either. Colorizers need the answer regardless, to decide whether a display is
+// one they should colorize for the controller they selected.
+static inline uint32_t CtlDisplayControllerId(const DisplaySrcId* display)
```
vbousquet: See above, but I would not add this as I don't get why we double the overrideId field

plugins/plugins/MsgPluginManager.h
```diff
@@ -156,7 +157,14 @@ class MsgPluginManager final
...
    };
-   std::vector<MsgEntry> m_msgs;
+   // deque, not vector: message ids are indices into this container, and
+   // BroadcastMsg/SendMsg iterate m_msgs[msgId].callbacks while invoking
+   // subscribers. A subscriber that resolves a new message id from inside its
+   // own callback -- which is ordinary, plugins do it to answer a query -- can
+   // grow this container mid-iteration. With a vector that reallocates and the
+   // range-for's bound reference dangles. deque never invalidates references to
+   // existing elements when growing at the back.
+   std::deque<MsgEntry> m_msgs;
```
vbousquet: I think we should clarify what we cover here. During BroadcastMsg/SendMsg processing (iteration over callbacks):
- register/unregister message is legal (including the processed one ?), therefore maybe causing a std::move of the callbacks vector
- register/unregister callback is legal, therefore maybe causing an invalidation of the callbacks iterator

There is a latent bug here, but I don't think a dequeue would fully fix it

plugins/upscaledmd/UpscaleDMD.cpp
```diff
@@ -177,6 +177,7 @@ class DMDUpscaler
       const DisplaySrcId displayId = { //
          .id = { { endpointId, 0 } },
          .overrideId = m_dmdSrc.id,
+         .controllerId = CtlDisplayControllerId(&m_dmdSrc),
```
vbousquet: See comment above. I think this should be removed

plugins/vni/vni.cpp
```diff
@@ -67,7 +67,7 @@ class VNIColorizer
       // Only keep dmd corresponding to selected controller (or overrides to support alphanumeric rendered DMD for example)
       const std::function<bool(const DisplaySrcId&)> isFromController = [&](const DisplaySrcId& src)
       {
-         if (src.id.endpointId == m_controllerEndpointId)
+         if (CtlDisplayControllerId(&src) == m_controllerEndpointId)
```
vbousquet: See comment above.

vbousquet (commented):
I'm away from keyboard, so just add a quick review (not building the code, just reading). I do not understand the proposed change to the plugin API so far. Could you give deeper explanation ?

mkalkbrenner:
I created a new PR to get rid of the commit history that might confuse people if it gets merged without squash:
https://github.com/vpinball/vpinball/pull/3929

Issue #3915, opened 2026-09-06
Spinners crashing VPX on obj export


Sixtoe:
**Describe the bug**
Spinners crashing VPX on obj export

**To Reproduce**
Have a table, add a spinner, go to export obj mesh, vpx will crash

**Versions & Settings**
 - VPX: 10.8.1 Beta 5577 BGFX 64bit
 - Windows 11

gitfool:
Since you have a crash could you try latest and post the crash report (`crash.txt`) contents?

Sixtoe:
`Crash report VPX - v10.8.1 Beta (Rev. 5582 (d922d6dab), windows BGFX 64bits)
============
Process: VPinballX_BGFX64.exe
Reason: 0xC0000005 - EXCEPTION_ACCESS_VIOLATION at 0033:00000001402E249C
Attempt to read from 0x00000144
Thread ID: 0x30F4 [12532]

Call stack
==========
00000001402E249C VPinballX_BGFX64.exe (0x0015D200 0x02C0C458 0x00000000 0x00002f4c)
00000001402E11FB VPinballX_BGFX64.exe (0x0015CED8 0x0B2A3758 0x0015CF10 0x0015d090)
0000000140453389 VPinballX_BGFX64.exe (0x00000001 0x00000000 0x002E0533 0x00000000)
000000014045713F VPinballX_BGFX64.exe (0x0015E240 0x00000000 0x00000000 0x0007063e)
00000001401CF7FC VPinballX_BGFX64.exe (0x0015E240 0x00000000 0x00000111 0x00000000)
000000014046B431 VPinballX_BGFX64.exe (0x00000000 0x00000000 0x00000000 0x00000000)
000000014045778C VPinballX_BGFX64.exe (0x00000000 0x0015E2B0 0x00000000 0x00000001)
00007FFF9D1DC396 USER32.dll CallWindowProcW + 0x6A6 (0x0015E170 0x140192C70 0x000606D8 0x00000000)
00007FFF9D1DA7ED USER32.dll IsWindowUnicode + 0x3CD (0x0015E170 0x00000001 0x00000000 0x0015f8d0)
00000001401FCFBF VPinballX_BGFX64.exe (0x141165DE8 0x000000C0 0x000000C0 0x0052a780)
0000000140183DE2 VPinballX_BGFX64.exe (0x001B0002 0x0015F1D0 0x00000000 0x0052a780)
[...]

PR #3924, opened 2026-09-07
Windows: fix spinner crash on mesh export

gitfool:
Fixes the crash in #3915: adding a spinner and exporting the table to OBJ crashes VPX in the editor.

`Spinner::ExportMesh` calls `Spinner::UpdatePlate` to generate the plate geometry, and `UpdatePlate` read the spinner's current angle straight off `m_phitspinner->m_spinnerMover.m_angle`. `m_phitspinner` is the play-time physics object, it only exists while a table is running and is null in the editor. So the export path dereferenced a null pointer and faulted (`spinner.cpp:403`).

Every other accessor on the spinner already guards this pointer and falls back to the design value (`m_phitspinner ? ... : m_d....`). `UpdatePlate` was the one place that didn't. The fix reads the angle through the same guard, using the rest angle (0) when there's no live physics object, which is the correct neutral geometry for an export. This matches how the flipper exports its base mesh from the design start angle rather than the live mover.

```cpp
const float angle = m_phitspinner ? m_phitspinner->m_spinnerMover.m_angle : 0.0f;
```

I checked the other parts with an `ExportMesh` and a play-time hit object (gate, trigger, flipper, kicker, bumper, hittarget, ramp, rubber, surface, primitive). Spinner is the only one that dereferenced its play-time pointer on the export path, the rest build from design data or static meshes.

This is a minimal guard, not the real cleanup. As @vbousquet noted on the issue, the spinner (and trigger) mesh generation is tangled up with the play-time update and should be pulled apart so export never depends on the physics object at all. That's a larger refactor. This fix stops the crash and unblocks export in the meantime.

Tested by exporting a spinner to OBJ from the editor: before, it crashed at `Spinner::UpdatePlate`; after, the export completes and writes the mesh.

### DIFF
--- a/src/parts/spinner.cpp
+++ b/src/parts/spinner.cpp
@@ -388,7 +388,8 @@ void Spinner::Render(const unsigned int renderMask)
       m_renderer->m_renderDevice->DrawMesh(m_renderer->m_renderDevice->m_basicShader, false, pos, 0.f, m_bracketMeshBuffer, RenderDevice::TRIANGLELIST, 0, spinnerBracketNumFaces);
    }
 
-   if (m_phitspinner->m_spinnerMover.m_visible && !isStaticOnly)
+   const bool plateVisible = m_phitspinner ? m_phitspinner->m_spinnerMover.m_visible : m_d.m_visible;
+   if (plateVisible && !isStaticOnly)
    {
       UpdatePlate(nullptr);
       Vertex3Ds pos(m_d.m_vCenter.x, m_d.m_vCenter.y, m_posZ);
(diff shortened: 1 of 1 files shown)


gitfool:
@jsm174 updated per the review. Two correctness fixes on top of the original guard:

- `Spinner::Render` had the same unguarded `m_phitspinner` deref, so a spinner in a non-playfield PartGroup would still fault on the first frame. Now guarded with the design-value fallback: `m_phitspinner ? m_phitspinner->m_spinnerMover.m_visible : m_d.m_visible`, matching `get_Visible`.
- The export fallback angle was `0.0f`, which is wrong for a spinner whose angle range excludes 0. Now `clamp(0.f, ANGTORAD(min(m_d.m_angleMin, m_d.m_angleMax)), ANGTORAD(max(...)))`, matching what `HitSpinner` initializes the mover to (`collideex.cpp:362`).

Left as separate follow-ups: passing the angle into the plate generator (cleaner, but relocates the null check rather than removing it, so it doesn't replace the `Render` guard), and the identical `Gate::Render` deref (pre-existing, different part).

Re-tested: spinner OBJ export from the editor completes, no crash.

PR #3925, opened 2026-09-08
macOS: in-process crash reports with named call stack

gitfool:
Gives the macOS standalone build the same crash-reporting quality as the Windows builds: on a fatal signal it writes `crash.txt` with a symbolicated call stack (function names + file:line), resolved in-process, starting at the faulting frame.

## Cause / approach

The MSVC build uses dbghelp/PDB and the MinGW build uses libbacktrace/DWARF; neither works on macOS. libbacktrace specifically fails on Mach-O/dSYM (reads the dSYM but returns null, and doesn't handle the PIE slide), which is why the Rust ecosystem abandoned it there too. The fix uses CoreSymbolication (Apple's own symbolication engine, same as `atos`/Xcode) to resolve our own frames to file:line, loading the app's `.dSYM` by an explicit path and de-sliding addresses manually. System-library frames are named with `dladdr` (module + symbol + offset, no line info), mirroring dbghelp naming system DLL frames on Windows.

## Fault-first stack

A `SIGSEGV`/`SIGBUS`/`SIGILL`/`SIGFPE`/`SIGABRT` handler (installed via `sigaction` with `SA_SIGINFO` on an alternate stack) walks the crashing thread's stack so frame 0 is the faulting instruction, not our handler. It takes a `backtrace()`, drops the leading frames through `_sigtramp` (the kernel signal trampoline, above which are only our own frames), and prepends the `ucontext` PC to recover a leaf that faulted without making a call. This mirrors the Windows path starting from the exception `CONTEXT`. The report is written from the signal handler using only `write(2)` and stack buffers.

## dSYM shipping (the macOS PDB equivalent)

The `.dSYM` is generated into `Contents/Resources/` (a sealed resource, not `Contents/MacOS/` where it would count as nested code and break codesign/notarization) and ships inside the app. `InitSymbols()` computes the path relative to the running executable and loads it via `CSSymbolicatorCreateWithURLAndArchitecture`, so resolution is deterministic and does not depend on Spotlight indexing the dSYM by UUID (which isn't true on an end-user machine). Built with `-g1` for line tables at roughly half the dSYM size of `-g`. `dsymutil` runs as a POST_BUILD step (it harvests DWARF from the `.o` files, so it must run after link while they exist); the app is signed after, sealing the dSYM.

## Testing

Verified on arm64, both a startup crash (main thread) and an in-play crash (player thread), each producing a single call stack that begins at the fault, with our frames carrying file:line and system frames named.

Startup crash (main thread):

```
Call stack
==========
0x102d37254 VPApp::VPApp() VPApp.cpp(153)
0x102c4769c WinMain main.cpp(195)
0x1033fb824 -[VPXAppDelegate makeWindowExit] main.mm(70)
0x1845999ec CoreFoundation __CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__ + 0x94
...
0x1889cd13c AppKit -[NSApplication run] + 0x170
0x1033fb978 main main.mm(109)
0x18411c4e4 dyld start + 0x1B50
```

In-play crash (player thread):

```
Call stack
==========
0x100c177ec PhysicsEngine::PhysicsSimulateCycle(float) PhysicsEngine.cpp(498)
0x100c176a4 PhysicsEngine::UpdatePhysics(unsigned long long) PhysicsEngine.cpp(472)
0x100a19a90 Player::UpdateGameLogic() player.cpp(1848)
0x100a1a424 Player::MultithreadedGameLoop() player.cpp(1918)
0x1009eb8d8 PlayTableCommand::Execute() AppCommands.cpp(140)
0x100a0b6c4 WinMain main.cpp(206)
0x1011bf824 -[VPXAppDelegate makeWindowExit] main.mm(70)
0x1845999ec CoreFoundation __CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__ + 0x94
...
0x1889cd13c AppKit -[NSApplication run] + 0x170
0x1011bf978 main main.mm(109)
0x18411c4e4 dyld start + 0x1B50
```

The crash sites (`VPApp.cpp(153)`, `PhysicsEngine.cpp(498)`) are injected test faults (on a companion test branch, not this PR), pinpointed to the exact line. The player-thread case confirms cross-thread resolution: the symbolicator is created on the main thread at startup and queried from the player thread at crash time.

PTAL @jsm174

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -550,6 +550,11 @@ elseif(PLATFORM STREQUAL "macos")
    add_executable(vpinball
       ${VPX_STANDALONE_SOURCES}
 
+      src/utils/StackTrace.cpp
+      src/utils/StackTrace.h
+      src/utils/CrashHandler.cpp
+      src/utils/CrashHandler.h
+
       standalone/macos/main.mm
    )
 
@@ -571,6 +576,8 @@ elseif(PLATFORM STREQUAL "macos")
       WINE_UNICODE_NATIVE
 
       "__forceinline=__attribute__((always_inline)) inline"
+
+      VPX_STANDALONE_CRASH_HANDLER
    )
 
    if(RENDERER STREQUAL "BGFX")
@@ -597,6 +604,7 @@ elseif(PLATFORM STREQUAL "macos")
       -ffp-contract=off
       -Wno-shorten-64-to-32
       $<$<AND:$<STREQUAL:${ARCH},x64>,$<STREQUAL:${RENDERER},GL>>:-mssse3>
+      $<$<CONFIG:Release>:-g1>
    )
 
    set(RUNTIME_LIBS_DIR "${CMAKE_SOURCE_DIR}/third-party/runtime-libs/macos-${ARCH}")
@@ -613,6 +621,8 @@ elseif(PLATFORM STREQUAL "macos")
       hidapi
       winevbs
       "-framework AppKit"
+      "-framework CoreFoundation"
+      "-F /System/Library/PrivateFrameworks" "-framework CoreSymbolication"
    )
 
    target_link_libraries(vpinball PUBLIC $<IF:$<STREQUAL:${RENDERER},BGFX>,bgfx,glad>)
@@ -651,6 +661,10 @@ elseif(PLATFORM STREQUAL "macos")
       COMMAND "${CMAKE_COMMAND}" -E copy "${CMAKE_SOURCE_DIR}/standalone/macos/VPinballX.icns" "$<TARGET_FILE_DIR:vpinball>/../Resources"
       COMMAND /usr/libexec/PlistBuddy -c "Set :CFBundleShortVersionString ${CMAKE_PROJECT_VERSION}" $<TARGET_FILE_DIR:vpinball>/../Info.plist
       COMMAND /usr/libexec/PlistBuddy -c "Set :CFBundleVersion ${CMAKE_PROJECT_VERSION}" $<TARGET_FILE_DIR:vpinball>/../Info.plist
+   )
+
+   add_custom_command(TARGET vpinball POST_BUILD
+      COMMAND dsymutil "$<TARGET_FILE:vpinball>" -o "$<TARGET_FILE_DIR:vpinball>/../Resources/$<TARGET_FILE_NAME:vpinball>.dSYM"
    )
 
 # Linux build (BGFX / GL)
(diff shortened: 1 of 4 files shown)


jsm174:
Is there anyway we could avoid private frameworks? 

`-F /System/Library/PrivateFrameworks" "-framework CoreSymbolication"`

gitfool:
Good question, and worth digging into properly, especially if MAS distribution is desired. The issue is specifically App Store review (Guideline 2.5.1): it scans the binary and rejects non-public framework links / `CS*` symbols with `ITMS-90338`. (For the current Developer-ID + notarization path it wouldn't be flagged, notarization is a malware scan, not an API check, but that doesn't help if MAS is a goal.)

Ideally we hit both goals at once: keep file:line resolution and stay fully public-API/MAS-safe. The line detail is worth preserving, it makes a crash report most easily actionable (whoever's debugging sees the exact source line without spelunking), matching the Windows PDB experience.

Here's the landscape as I found it, framed against those two goals:

- **CoreSymbolication (current).** In-process, name + file:line, deterministic. But it's a private framework, so not MAS-safe. This is the one thing that has to change for MAS.
- **dladdr only.** Public/MAS-safe, in-process, but symbol names only, no file:line (it reads the Mach-O symbol table, not DWARF). It's the Rust ecosystem's own fallback layer. Safe and simple, but gives up file:line.
- **libbacktrace** (like the MinGW build). I tried this first; it's broken on macOS, it reads the dSYM but returns null and doesn't handle the PIE slide. The Rust project hit the same wall and dropped it for Apple platforms. Dead end.
- **atos shell-out.** Public, gives file:line, but spawning a subprocess (fork/exec) from inside a signal handler is fragile, and it's a heavier runtime dependency. I'd avoided shell-out to match the MinGW in-process approach.
- **Vendor a DWARF reader** (LLVM's symbolizer / a gimli-equivalent). This is the one option that hits both goals: public API, in-process, name + file:line, handles the slide. The cost is a real new dependency to vendor and wire up (more code than the private-framework link it replaces), so it needs weighing.

Reading those, the only route that keeps file:line and is MAS-safe is vendoring a DWARF reader; everything else trades away one goal or the other. Before I go down that path, do you have a preference on which reader (or a lighter approach I've missed)? And if the vendoring turns out heavier than it's worth, would you be OK landing dladdr-only (names, MAS-safe, small) first and treating file:line as a follow-up?

PR #3926, opened 2026-09-08
docs: add .cromc as supported file format for Serum

dekay:
VPX Serum code supports both `.crc` (old) and `.cromc` (new) formats. The latter is preferred [for lower resource usage](https://vpuniverse.com/forums/topic/14058-serum-has-swapped-to-cromc-files-why/).

Found this while writing [some docs for the wiki](https://github.com/dekay/vpinball-wiki/wiki/File-Layout) that expands on this document with more details, examples, troubleshooting notes, etc.

### DIFF
--- a/docs/FileLayout.md
+++ b/docs/FileLayout.md
@@ -45,7 +45,7 @@ The settings system is based on 2 files:
 
 The table override file allows to tweak a few settings for playing a given table by overriding the settings defined in the global settings.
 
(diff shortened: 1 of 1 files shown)


vbousquet:
The idea of this filelayout page is not to exposed all possible files with informations about each of them, but to give an overall layout informations. So we may update to the new preferred Serum file format but the information are in fact already correct. 

The additional information about crz against cromc is out of context here. Could you remove them ?

dekay:
I found another small change to the docs I'd like to make so I'm going to close this PR and combine them in another, incorporating your suggested change at that time. Thanks!

PR #3929, opened 2026-09-09
Serum plugin: new serum version, settings, scene triggers, plugin dir for embedding hosts

mkalkbrenner:
Four changes to the Serum plugin, plus the two small pieces of build plumbing a host outside VPX needs to use it. PPUC drives real pinball machines and embeds the plugin system without building the vpinball app, which is where all of this was found.

serum: reach SerumPath in hosts with no VPX API. GetColorization returned an empty path whenever vpxApi was null, which is every non-VPX host, and SelectController reads an empty path as "this controller has no colorization" -- so Serum never loaded at all outside VPX. The table-relative probes are skipped in that case and the global setting is reached as intended.

serum: expose IgnoreUnknownFramesTimeout and MaximumUnknownFramesToSkip. Both are colorization- and ROM-dependent, libserum defaults them to 0/0, and a host had no way to set them.

serum: add a Resolution setting. A v2 colorization can carry a 32 and a 64 row output, and both were requested and published unconditionally. Consumers then disagree about which to take -- DMDUtilPlugin takes the largest colour depth, ResURIResolver breaks on the first match -- so a 128x32 panel can end up downscaling a colorization upscaled to 256x64. libserum also computes both every frame, which on a Raspberry Pi driving an SD panel is half the work discarded. Setting 32 or 64 requests and publishes only that size; anything else keeps today's behaviour of producing both, which remains the default.

serum: add a PupTriggers setting, and trigger colorization scenes from the shared event stream. A colorization can carry PUP triggers, forwarded on Serum/OnDmdTrigger:1; whether that is wanted depends on the setup, and libserum has the switch for it, so the plugin now applies it. In the other direction, a colorization can contain scenes the ROM's own frames never select -- Flash is the case, shipping flash_l1.cROMc alongside flash_l1.pup.csv -- and Serum_Scene_Trigger had no way to be reached. Rather than adding a Serum specific message, 'D' events on B2S/OnStateChange:1 are consumed: they are already documented as DMD triggers for PUP and Serum both, and anything can emit one, so a colorization scene is triggered exactly the way a PUP scene is. Ids outside 50000..62000 are ignored, matching the window libdmdutil applies, and only the press acts since these are pulses. Triggers are queued and applied on the colorize thread under m_stateMutex, because Serum_Scene_Trigger mutates libserum's animation state and has to run where Serum_Colorize does.

VPINBALL_PLUGIN_PACKAGE_DIR lets a host that has no vpinball target say where the plugins are packaged. Kept separate from the rpath setting so overriding the directory cannot change how plugins are linked.

libdmdutil moves to 4b46ee47 for the libserum and libzedmd versions the Resolution work depends on.

Verified against real ROMs through PPUC.


plugins/serum/serum.cpp
```diff
@@ -38,21 +38,79 @@ static const MsgPluginAPI* msgApi = nullptr;
...
+// Serum skips frames it cannot identify. How long to keep showing the last
+// known-good colorized frame before giving up on the unknown run, and how many
+// unknown frames to skip within it, are colorization- and ROM-dependent, so
+// they have to be tunable by the host. libserum defaults to 0/0, which means
+// "no timeout, skip nothing".
+MSGPI_INT_VAL_SETTING(serumIgnoreUnknownFramesTimeoutProp, "IgnoreUnknownFramesTimeout", "Ignore unknown frames timeout",
+   "Milliseconds to keep the last colorized frame while frames cannot be identified (0 disables)", true, 0, 65535, 0);
+MSGPI_INT_VAL_SETTING(serumMaxUnknownFramesToSkipProp, "MaximumUnknownFramesToSkip", "Maximum unknown frames to skip",
+   "How many consecutive unidentified frames may be skipped (0 disables)", true, 0, 255, 0);
+// A Serum v2 colorization can carry a 32 row and a 64 row output, and by
```
vbousquet, line 57: I would say that this goes the overall design by reverting the setup from the consumer to the producer. Serum should advertise what it can produce (and actually produce it if requested). This is the task of the consumer to only request what is used. Eventually Serum plugin could be optimized to only request colorization of requested format, but putting this option should really be avoided.

If the problem is that DMDUtil does not select a DMD matching the hardware it drives resolution, then DMDUtil must be changed, not Serum.

plugins/serum/serum.cpp
```diff
@@ -38,21 +38,79 @@ static const MsgPluginAPI* msgApi = nullptr;
...
+// on "Serum"/"OnDmdTrigger:1" for PUP and DOF to act on. Whether that is wanted
+// depends on the setup rather than on the colorization: a pack author may be
+// driving PUP from a .pup.csv instead and want the colorization's own triggers
+// out of the way, and a host with no PUP at all has no use for them.
+//
+// libserum reports them unless told otherwise (keepTriggersInternal defaults to
+// false), so the default here keeps what the plugin has always done. Note that
+// libserum's own switch is global rather than per-instance, which is why this
+// is applied on every load rather than once.
+MSGPI_BOOL_VAL_SETTING(serumPupTriggersProp, "PupTriggers", "Report colorization PUP triggers",
```
vbousquet, line 85: The problem here seems to be that the Serum triggers:
- are coalesced with other triggers (into a 'D' event)
- and use teh same trigger ID than the ones in the PUP pack

Adding this option is a UX anti pattern. The problem should be solved at the design, not relying on the user to understand deeply how things work and tweak it accordingly. So I would really avoid to expose this option. A simple hack, like adding 1000 to Serum trigger would solve this without needing users to dive into the pinball system communication.

plugins/serum/serum.cpp
```diff
@@ -38,21 +38,79 @@ static const MsgPluginAPI* msgApi = nullptr;
...
 
 static std::unique_ptr<CtrlItemConsumer<ControllerDef>> controllers;
 static std::unique_ptr<class SerumColorizer> colorizer;
 
 MSGPI_STRING_VAL_SETTING(serumPathProp, "SerumPath", "Serum Path", "Folder that cotains Serum colorization files (cROMc, cRZ)", true, "", 1024);
+// Serum skips frames it cannot identify. How long to keep showing the last
+// known-good colorized frame before giving up on the unknown run, and how many
+// unknown frames to skip within it, are colorization- and ROM-dependent, so
+// they have to be tunable by the host. libserum defaults to 0/0, which means
+// "no timeout, skip nothing".
```
vbousquet, line 52: I don't understand why this is part of the colorizer and not part of the colorization. The colorization author is supposed to know (and he is the only one) if he did colorized everything or not. This behavior should be part of the colorization, not needing the user to dive into this and have to tweak each single colorization. I would definitely not expose this

plugins/serum/serum.cpp
```diff
@@ -84,10 +149,22 @@ class SerumColorizer
    void FilterDmdSource(std::vector<DisplaySrcId>& items)
    {
       // Only keep dmd corresponding to selected controller (or overrides to support alphanumeric rendered DMD for example)
+      // A display the controller owns outright, or one derived from a resource it
```
vbousquet: The right way to select is to move up the override chain. The problem here is that we do it but only on display resources, which lead to miss alphaseg. The correct fix is to cleanly follow the complete override chain (this could be a template function provided by the Controller API header helper block).

The fix is not entirely correct and just hack to support a shallow override leading to a source which is not a display. THis should be marked with a `// FIXME` as the bug is still there

plugins/serum/serum.cpp
```diff
@@ -105,6 +182,20 @@ class SerumColorizer
          items.push_back(selected);
    }
 
+public:
+   // A scene trigger from the shared event stream. Deliberately not a Serum
```
vbousquet, line 193: There is no 'shared event stream' (the plugin design is made to move away from the locked we had before). Here you are listening for the backglass server publishing a DMD identification event. I don't think B2S server publish that. I guess that what you want is to listen for Pup DMD identification event ? Or maybe you want to recreate the legacy B2S plugin event stream (there is a class for that) ?

plugins/serum/serum.cpp
```diff
@@ -271,7 +403,7 @@ class SerumColorizer
                         .GetRenderFrame = &Trampoline<&SerumColorizer::GetRenderFrameSerumV2_32>::Call //
                      });
                   }
-                  if (colorizer->m_advertisedWidth64 > 0)
+                  if (colorizer->m_advertisedWidth64 > 0 && IsResolutionRequested(64))
```
vbousquet, line 413: THis against the overall design (see above)

plugins/serum/serum.cpp
```diff
@@ -371,7 +517,16 @@ static void SelectController(std::vector<ControllerDef>& items)
    for (const ControllerDef& controller : items)
    {
       const bool hasIdentifiableDmd = std::any_of(displays.begin(), displays.end(),
-         [&controller](const DisplaySrcId& display) { return display.id.endpointId == controller.endpointId && IsColorizableDmd(display); });
+         [&controller](const DisplaySrcId& display)
+         {
+            // Same reasoning as FilterDmdSource's isFromControllerEndpoint: a
```
vbousquet: See remark above as the code has the same problem

plugins/serum/serum.cpp
```diff
@@ -401,6 +556,34 @@ static void OnControllerChanged()
       });
 }
 
+// Scene triggers arrive on the shared B2S event stream, the same 'D' events PUP
```
vbousquet: I don't think this is true. There is nothing as a shared B2S event stream here.

plugins/serum/serum.cpp
```diff
@@ -401,6 +556,34 @@ static void OnControllerChanged()
...
+      uint8_t type;
+      int32_t index;
+      int32_t value;
+   };
+   const B2SPluginEvent* event = static_cast<const B2SPluginEvent*>(eventData);
+   if (event == nullptr || !colorizer)
+      return;
+   // value == 1 is the press: these are pulses, and acting on the release too
+   // would run every scene twice.
+   if (event->type != 'D' || event->value != 1)
```
vbousquet: THis will never fire. You want PUP events, not B2S state change

make/CMakeLists_plugins.txt
```diff
@@ -11,8 +11,19 @@ if(PLATFORM STREQUAL "ios" OR PLATFORM STREQUAL "ios-simulator" OR PLATFORM STRE
 endif()
 
 if(PLATFORM STREQUAL "macos")
-   set(PLUGINS_DIR "$<TARGET_FILE_DIR:vpinball>/../PlugIns")
    set(CMAKE_INSTALL_RPATH "@loader_path/../../Frameworks;@loader_path;@executable_path")
+endif()
+
+# Hosts that embed the plugin system without building the vpinball app -- PPUC
```
vbousquet, line 17: @jsm174 Can you have a look ? I haven't yet developped the skills for the overall build system we have

plugins/serum/serum.cpp
mkalkbrenner: Serum MUST emit the identical triggers. It is meant to replace libpupdmd. That has been discussed with nailbuster long time ago and is implemented in DMDext already.

plugins/serum/serum.cpp
mkalkbrenner: Same here. This is in libserum for a long time. In the Serum editor you can just delete a frame if you don't want to have it colorized. In an unfinished colorization that would black out all not yet colorized frames as well. There are colorizations out there that are already released but are incomplete. This is the way to make them playable.
We can't discuss bad design decisions in Serum from log time ago here. A serum plugin must support serum.

plugins/serum/serum.cpp
mkalkbrenner: See my comment above. That made it work.

plugins/serum/serum.cpp
mkalkbrenner: I asked myself the same question two days ago. Now it turned out that the code anlaysis found synthetic ones in the B2S code and that the test cases and test code is wrong.

plugins/serum/serum.cpp
mkalkbrenner: Serum_Load takes the format flags at load time, so "only request what is used" needs the plugin to know which of its sources a consumer will take, which is the same negotiation that doesn't exist yet. Until it does, both outputs get computed regardless. On slow devices, creating a needless HD frame must be avoided.
Changing that would require changing Serum's API. But maybe it is the right time to do it ...

mkalkbrenner: It is nearly impossible. The loader loads a cROMc as a stream. It skips parts of the colorization which are not relevant for the target. Otherwise it is impossible to load some colorizations on a low memory device.
For VPX on a high power machine, that might not be relevant, but for PPUC running on a 512MB Raspberry Pi, it is.

plugins/serum/serum.cpp
mkalkbrenner: done

plugins/serum/serum.cpp
mkalkbrenner: you're right. Changed it.

plugins/serum/serum.cpp
mkalkbrenner: I checked the old conversations. This is the solution to deal with "transitional frames" in fast animations, especially for WPC. Authors are not able to create dumps that contain any frame that could ever happen. So it is impossible to put the decision not to display a specific frame into the Serum format.

plugins/serum/serum.cpp
mkalkbrenner: Same as above. I think we must accept this if the plugin infrastructure should be re-usable.

plugins/serum/serum.cpp
mkalkbrenner: There is already one colorization that drives a PUP Pack. I think it is an original table. But I can't remember which one.

But the same is already planned for STTNG, TZ and DH.

plugins/serum/serum.cpp
mkalkbrenner: B2s Is now gone, we consume PUP DMD triggers.

plugins/serum/serum.cpp
vbousquet: > Serum MUST emit the identical triggers. It is meant to replace libpupdmd. That has been discussed with nailbuster long time ago and is implemented in DMDext already.

Serum emits trigger, but PUP uses them. Overlaying the same event from 2 different sources would be very hacky. As I understand, what is planned is that PUP uses Serum trigger instead of its own. But I would say that's a consumer thing, not a producer. So having an option here seems wrong to me. The hadling is on PUP side, not Serum

plugins/serum/serum.cpp
vbousquet: > I checked the old conversations. This is the solution to deal with "transitional frames" in fast animations, especially for WPC. Authors are not able to create dumps that contain any frame that could ever happen. So it is impossible to put the decision not to display a specific frame into the Serum format.

I understand that but then it should be the author, who has a deep knowledge and skills (and knows if he has finshed or not), who define this and store it inside the colorization. This looks to me like we are filling a design hole in Serum (legitimate as no tool is perfect from its first release) by requesting skills and hacks from the user instead of updating Serum (and having a sensible default for old colorization missing the needed field).

mkalkbrenner: You didn't understand the issue. The author doesn't have the knowledge and doesn't have the dumps.
User's report issues like the settings aren't displayed, the tournament mode is black, a wizard mode turns the DMD black, ...
This setting is the only way to display these frames in monochrome until the author took care about them.
But you need the timings, otherwise you will see lots of transitional frames in fast animations that author wants to hide.
I understand your concerns. But the question is wether VPX wants to support Serum as it is or not.

plugins/serum/serum.cpp
mkalkbrenner: How should we implement that? A per game setting for PUP wether to accept the events or not? What about other consumers?

make/CMakeLists_plugins.txt
jsm174: @vbousquet, @mkalkbrenner  - Instead of a package dir flag, I wonder if we should somehow have an option to only build plugins, instead of main application. Since VPX is generic, we probably don't need the added comments either. I'll take a stab at this in another commit, so this part could just be rebased.

PR #3932, opened 2026-09-09
Detect duplicate sound and image names case insensitively

francisdb:
The duplicate check when loading a table compared names exactly, while `GetSound()` and `GetImage()` look names up case insensitively and return the first match. A table with sounds named "drain" and "Drain" (Torpedo Alley 2.2 - JPJ - DOZER - TeamPP) therefore loaded both without a warning while the second could never be played. This compares the same way the lookups do, so the unreachable duplicate is dropped and logged like an exact one.


vbousquet:
Thanks

PR #3944, opened 2026-09-12
Fix windows msvc external build isolation

gitfool:
## Summary

The `windows-x64` and `windows-x86` external dependency builds could fail on a developer machine that has MSYS2 installed, even though those builds compile the main libraries with MSVC. Two independent problems, both fixed here:

1. The OpenXR build picked up MSYS2's `jsoncpp` and injected ABI-incompatible MinGW headers into the MSVC compile.
2. `MSYS2_PATH` was hardcoded to `/c/msys64`, so any MSYS2 installed elsewhere (for example via scoop) broke the ffmpeg/libzip steps and the runtime DLL copies.

Both Windows MSVC scripts are updated. The `windows-mingw-x64` and non-Windows platforms are untouched.

## Problem 1: OpenXR pulls in MSYS2 jsoncpp under MSVC

OpenXR's `src/CMakeLists.txt` calls `find_package(JsonCpp)` unconditionally. On a machine where MSYS2 has `mingw-w64-ucrt-x86_64-jsoncpp` installed (it arrives as a dependency of `mingw-w64-ucrt-x86_64-cmake`, so it is present on most MSYS2 setups), cmake's config-mode search finds `ucrt64/lib/cmake/jsoncpp` and adds its `ucrt64/include` directory to the build. Under the Visual Studio generator, MSVC then parses MinGW's `corecrt.h` / `corecrt_wstdio.h`, which use GCC builtins (`__UINTPTR_TYPE__`, `__asm__`), producing hundreds of errors like `'uintptr_t': undeclared identifier` and `'__asm__': unknown override specifier`.

Fix on the OpenXR configure step only:

- `-DCMAKE_FIND_USE_SYSTEM_ENVIRONMENT_PATH=OFF` and `-DCMAKE_FIND_USE_CMAKE_SYSTEM_PATH=OFF`. Both are required. jsoncpp is reachable through two independent search vectors (PATH-derived prefixes and cmake's system prefixes); disabling only one still finds it. With both off, cmake reports `Could NOT find JsonCpp` and OpenXR falls back to its vendored jsoncpp source, matching how CI behaves.
- `-DPython3_EXECUTABLE="$(which python3)"`. The environment-path flag above also disables PATH-based discovery of the `python3` interpreter, which OpenXR needs for dispatch-table generation. Passing it explicitly restores that.
- `-DBUILD_API_LAYERS=OFF`. The API layers are the only consumer of jsoncpp, and only `openxr_loader64` is shipped, so the layers are never used. Disabling them also cuts about 54s off the OpenXR build (81s to 28s locally). This is an optimization, not the fix: disabling the layers alone does not prevent the contamination, because `find_package(JsonCpp)` still runs and poisons the include path before the layers are even configured. The find-isolation flags are what actually resolve it. This mirrors the `-DBUILD_API_LAYERS=OFF` already present in `windows-mingw-x64` (where it is safe without isolation flags because that build is entirely MinGW and ABI-consistent).

## Problem 2: MSYS2 root hardcoded to /c/msys64

`MSYS2_PATH` defaulted to `/c/msys64`, used for the MinGW-toolchain builds (ffmpeg, libzip) and for copying MinGW runtime DLLs. An MSYS2 installed elsewhere failed with `/c/msys64/usr/bin/bash.exe: No such file or directory`, and the failure propagated into the nested `libdmdutil` / `libdof` sub-scripts, which share the same default.

Fix:

- Keep honoring an explicit `MSYS2_PATH` override, and keep `/c/msys64` as the default.
- Only when that default is absent, derive the root with `cygpath -m /` and accept it only if it contains the toolchain tree this build consumes (`ucrt64` for x64, `mingw32` for x86). That tree is unique to MSYS2; git-bash and Cygwin ship `mingw64` at most, never `ucrt64`/`mingw32`, so a non-MSYS2 shell correctly falls back rather than mis-detecting its own root. This matters because git-bash sets `MSYSTEM`/`MSYSTEM_PREFIX` and has a `mingw64/bin`, so a weaker check would wrongly accept it.
- `export MSYS2_PATH` so the nested `libdmdutil` / `libdof` sub-scripts inherit the resolved value.

Because the derivation only runs when `/c/msys64` is missing, the standard install and CI (GitHub runners have MSYS2 at `/c/msys64`, and the plain `windows` job runs under git-bash) never enter the new path and are unaffected.

## Testing

Built locally on Windows with MSYS2 installed via scoop (outside `/c/msys64`), MSVC (VS 2026 Insiders build tools), from an MSYS2 UCRT64 shell:

- `windows-x64`: full external build. OpenXR configures with `Could NOT find JsonCpp`, `openxr_loader64.dll` links, and the rest of the tree (SDL3, SDL3_image, SDL3_ttf, FreeImage, bgfx, PinMAME, libdmdutil, libaltsound, libdof, ffmpeg, libzip) builds.
- `windows-x86`: full external build, including OpenXR and ffmpeg, completed the same way.
- Verified `MSYS2_PATH` auto-resolves to the scoop location from an MSYS2 shell, and falls back to `/c/msys64` under git-bash (where the `ucrt64`/`mingw32` tree check correctly rejects git-bash's root).

## Notes

- Scope is limited to the two Windows MSVC scripts. `windows-mingw-x64`, macOS, Linux, iOS, and Android are unchanged.
- The isolation flags are deliberately applied only to the OpenXR step, the one build that does `find_package` for a library MSYS2 ships as a cmake config. The other MSVC builds were left as-is since they do not hit this.


gitfool:
Thanks @jsm174 for the offline review. I pushed changes. This is a better fix than what I had, so I've pivoted to your suggestion.

The OpenXR build now just adds `-DBUILD_WITH_SYSTEM_JSONCPP=OFF` next to the existing `-DBUILD_API_LAYERS=OFF`, and I dropped the rest: the two `CMAKE_FIND_USE_*` find-path flags, the `-DPython3_EXECUTABLE` workaround (only needed because the env-path flag had disabled Python discovery, Python resolves fine now), and the `MSYS2_PATH` cygpath detection (back to the `/c/msys64` default with the env-var override, like the other scripts).

`BUILD_WITH_SYSTEM_JSONCPP=OFF` does the job cleanly. OpenXR still finds the MSYS2 jsoncpp but no longer uses it, it compiles the vendored source, so the ucrt64 headers never reach the MSVC compile. Applied to all three Windows scripts.

Verified on Windows x64 and x86 (MSVC, VS 2026, scoop MSYS2 with `MSYS2_PATH` set): OpenXR configures with jsoncpp found-but-unused, the loader links clean, and the full external build completes.

PR #3945, opened 2026-09-12
Honor ResetLogOnPlay setting for all builds

gitfool:
## Summary

The `[Standalone] ResetLogOnPlay` setting was only acted on by the iOS and Android apps. The desktop editor and the CLI ignored it, so setting it did nothing there. This moves the reset into the engine so every build honors it, tied to table play rather than app launch. The default is unchanged: on for mobile, off elsewhere.

## When it resets

The log resets only when a table is played, and only if the setting is on:

- CLI `-Play`. The reset runs before the startup banner is written, so the fresh log keeps the version, paths, and CPU banner.
- Editor Play button, not the edit or POV modes.
- Mobile table load.

Opening the editor and non-play commands (`-Pov`, `-Audit`, script export, tournament validation) do not reset.

## Banner

On the CLI path the reset runs before the startup banner, so the fresh log opens with it and the console shows it once. On the editor and mobile paths the reset happens per play, after the banner was logged at startup, so those logs do not carry the banner for that play. This matches the prior mobile behavior, which also dropped the banner on reset.

## Truncation

Truncation previously opened a separate stream on the log path. On Windows the appender holds the file open with share-deny-write, so that second handle would be refused and the truncate would do nothing. On POSIX it would leave the appender's cached size stale and could race a concurrent write.

`Logger::Truncate()` now goes through the appender. A new `RollingFileAppender::truncate()` closes, unlinks, and reopens the file under the appender's own mutex, mirroring the existing `rollLogFiles()`. This fixes the Windows case and removes the race.

## plog change

Adds one public `truncate()` method to `third-party/include/plog/Appenders/RollingFileAppender.h`. It runs only when called, so no other plog user is affected.

## Removed code

With the engine doing the reset, the `VPinballResetLog` C API has no remaining callers and was the only unused export in the libvpinball C API. The whole chain is removed: the Swift binding, the Java native declaration, the JNI wrapper, the C API declaration and implementation, and `VPinballLib::ResetLog()`. The setting and its "Reset Log on Play" toggle in both mobile settings screens are unchanged and still drive the reset.

## Testing

- macOS standalone: `-Play` with the setting on resets the log and keeps the banner with no console duplicate; off does not reset; `-Pov` does not reset.
- Windows MinGW and Windows x64 editor: CLI `-Play` verified.

The mobile and editor Play button reset paths are covered by inspection, not run.

## Note

`ResetLogOnPlay` lives in the `[Standalone]` INI section. That section is read by the standalone build family. This change also reads it from the MSVC editor build via the CLI `-Play` path, the one consumer that is not itself a standalone build. It works. If you would prefer this setting in a more general section, I can move it. I left it in `[Standalone]` to avoid an INI-key break for existing users during beta.


jsm174:
If it becomes available to all platforms then it should not be in the [Standalone] section. 

Also the mobile apps have a clear button on the lower right hand side:

image

I’m away from computer so I can’t look at the code but just make sure that still works. I thought the UIs called reset via jni and the swift bridge.

gitfool:
@jsm174 on the clear button: it should still work. Both mobile clear actions truncate the log file directly rather than going through the reset bridge. iOS `CodeView.handleClear()` does `fileHandle.truncate(atOffset: 0)`, and Android `CodeWebViewDialog` writes an empty byte array over the file. Neither calls `VPinballResetLog`, so removing that C API doesn't affect them. The only callers of `VPinballResetLog` were the reset-on-play launch paths, which this PR replaces.

On the section: agreed, it shouldn't stay in `[Standalone]` now that all builds honor it. Two candidates:

- `[Editor]`, alongside `EnableLog` and `LogScriptOutput`. Those are already general all-builds logging settings that happen to live there (`EnableLog` gates the log appender on every build via `SetupLogger`), so this groups the log settings together. The section name is a bit misleading for non-editor builds, but the precedent is there.
- `[Player]`, which applies to all builds and matches "on play". Which begs the question of whether `EnableLog` (and `LogScriptOutput`) should move there too, since they're all-builds settings sitting under `[Editor]` for seemingly historical reasons.

Thoughts?

jsm174:
Thanks. I wrote that clear forever ago. 

I would think `[Player]` might be a better place? Any thoughts @vbousquet, @toxieainc ?

vbousquet:
I would say that if we do a (little) breaking change, this would be nice to move to something fully clean, so a new common section may eb the best here. Something like [Global] would look better as it applies to any flavor of the application.

jsm174:
I like it, sold!

gitfool:
@jsm174 @vbousquet done, pushed as a follow-up commit. Moved all three log settings (`EnableLog`, `LogScriptOutput`, `ResetLogOnPlay`) to a new `[Global]` section, written between `[Version]` and `[Editor]` in the INI.

Verified all three read from `[Global]` and work on macOS: `EnableLog` toggles the log file, `LogScriptOutput` toggles `Debug.Print` output, and `ResetLogOnPlay` truncates on play.

Issue #2263, opened 2025-03-20
Reducing the amount of vbscript required


francisdb:
This is just a thought experiment and a place to keep track of possible ideas.
All these things would move from vbs to configuration, but with the option to tweak in the script.
Once this is implemented it would be easier to move to a different scripting language.

* max balls, size, mass
* ball rolling sounds
* ball collision sounds
* nudge sounds
* flipper sounds
* tilt handling
* lut selection
* player tracking
* score tracking (Original and PinMAME)
* score saving
* default backglass
* what primitives should be visible in what mode
* PinMAME use (eg store gamename)
* implement everything in the vpinball scripts directory in c++

This also makes that vpinball itself knows more about the table and can optimize things, integrate more, eg a score tracking plugin that works for any table using these new features.

Some of these things are already being integrated by vpinball standalone and/or the new plugin system.

Would be nice if the main devs could give some feedback on:
* what is doable
* what needs to happen first
* what would make things impossible that are currently done by vbs

Feel free to add more in the comments, I will try to manage them in the description.

related to #568

PR #3959, opened 2026-09-14
JavaScript table scripts (QuickJS-NG)

### LINKED ISSUE #568: Consider a more modern scripting language
**Is your feature request related to a problem? Please describe.**

`VBScript` is ancient technology and hard to get 100% compatible cross-platform

**Describe the solution you'd like**

A more modern scripting language like `JS` that most devs/ppl these days have experience with. 

**Describe alternatives you've considered**

Fixing all existing and upcoming issues in wine vbscript. Might fix the `cross-platform` issues but will not improve the part where not many ppl know `vbs` these days.

**Additional context**

Of course this can't be a big bang migration but one needs to start somewhere so there needs to be some marker for JS scripts allowing to pick a different runtime.

francisdb:
Draft to gather feedback on running table scripts in JavaScript next to VBScript. Closes #568.

A table script whose first line is `// vpx-language: js` runs on an embedded QuickJS-NG engine (MIT, about 1 MB, interpreter only so fine for iOS); without the directive nothing changes for existing tables. A `.js` sidecar next to the table works like a `.vbs` one. The script is an ES module, table parts are globals, event handlers are exported functions named like the VBScript subs:

```js
// vpx-language: js
import { SoundFX, DOFContactors, DOFFlippers } from "controller.js";

export function Table1_KeyDown(keycode) {
   if (keycode === LeftFlipperKey) {
      LeftFlipper.RotateToEnd();
      PlaySound(SoundFX("fx_flipperup", DOFFlippers), 0, 0.67, AudioPan(LeftFlipper), 0.05, 0, false, true, AudioFade(LeftFlipper));
   }
}

export function Bumper1_Hit() {
   PlaySound(SoundFX("fx_bumper4", DOFContactors), 0, 1, AudioPan(Bumper1), 0, 0, false, true, AudioFade(Bumper1));
   b1l1.State = 1;
   Bumper1.TimerEnabled = true;
}

for (const light of GI)
   light.State = 1;
```

Errors report the JS stack, also through imported modules, and a runtime error in a handler does not stop the game:

```
Runtime error on line 179, col 59: TypeError: not a function
   at SoundFX (controller.js:179:59)
   at Plunger_Init (<table>:122:17)
```

The same mistake in VBScript reports `Runtime error on line 697, col 4: Object doesn't support this property or method`.

Why:

- VBScript is deprecated on Windows: Feature on Demand since 24H2, no longer enabled by default around 2027, removal after that. Standalone already carries its own Wine engine.
- One embedded engine behaves the same on every platform instead of `vbscript.dll` on Windows and Wine's implementation elsewhere.
- Actionable errors: typed messages with a stack, strict mode instead of silently created variables.
- Tooling exists: editors, linters, formatters; type declarations generated from the IDL are a possible follow up.
- Speed: 2M iterations of sqrt math take 913 ms in Wine VBScript, 547 ms in Windows `vbscript.dll` and 91 ms in QuickJS on the same CPU.

Not done yet, and needed before this could merge: porting the remaining scripts (core.vbs and the ROM family files), `[out]` parameters, the editor lexer, `TournamentFile` hashing of `core.js`. I plan to do these but wanted some initial feedback on the direction first. Tests: there is no unit test target upstream to add them to (#3544 was an attempt), so none here for now. The branch compiles on every platform of the CI matrix (run on my fork), but only the Linux standalone build was run; the Windows connection point sink is compiled but untested and the Visual Studio project files are not updated.

Open question: vendoring the QuickJS amalgamation under `third-party/include` versus building it in `external.sh`.


vbousquet:
This is a very interesting development. We also had a preliminary discussion regarding upgrading script support in VPX which is an aim that we share. Nothing was really done nor decided or even seriously evaluated so far but the overall directions was:
- to open to new language, especially python,
- to favor doing it outside the core, so by exposing what is needed to plugin and move the script engine to plugins
- to evaluate what it would mean to support script attached to subpart of the table (this is to go with another feature that has been progressing but not yet released, which is to be able to import/export subsets of table, allowing to create asset library, a clever way to include script would allow these assets to include behavior)

Latest development have been done with this target (and others) in mind. So I already pushed the first steps that allow to be able to foresee this feature but there are still quite some road to cover:
- [x] Split script from Win32 UI
- [x] Setup a global plugin system
- [x] Setup a platform agnostic scripting API inside the plugin system (this in turn should allow to expose VPX internals without using COM dispatch)
- [x] Abstract table part loading/saving allowing to use the same loader/saver for OLE file or the prototype vpz format (zip or folder with json + DCC assets)
- [ ] Evaluate how to split script (script part ?), including how to contextualize the script (importing 2 time a target asset with its associated script snippet should be supported without mismatch between events or name conflicts)
- [ ] Cleanup and publish the vpz format
- [ ] Expose the VPX COM API through the plugin scripting system (limited to relevant functions, perhaps finishing the in progress code cleanup that will lead the data model to be splitted from the UI first)
- [ ] Extend the VPX plugin API to provide the needed anchors to allow a plugin to contribute a script engine
- [ ] Move the VBS script engine to a plugin, preparing for deprecation of Windows core VBS engine (so 2 plugins: WineVBS and CoreWindowVBS ?)
- [ ] Implement a Python plugin

francisdb:
Thanks, good to know the aim is shared. A few points where this PR and that direction meet:

Language: I compared Python, Lua and JavaScript before starting. JS won on the size of the author pool, on tooling (type declarations generated from the IDL give completion and type checking today), on how naturally a dynamic language maps onto the IDispatch object model, and on runtime cost: QuickJS is about 1 MB and needs no JIT, while embedding CPython means shipping the stdlib on iOS and Android. I can post the comparison if useful.

Architecture: the first commit is the seam. `IScriptEngine` is what a plugin-provided engine would implement; the JS engine sits behind it and talks IDispatch only because that is what exists today, and it can move behind the plugin anchors once the plugin scripting API exposes the VPX objects. One concern with opening that up to arbitrary engines: it recreates the fragmentation we already have with tables that depend on `WScript.Shell` or `FileSystemObject` and only run on Windows. A table written for a language plugin only runs where that plugin is shipped, and a library of reusable assets only works if every table can run every asset's script. I would keep it to one language next to VBScript, shipped with every build, with the same API everywhere and no way out to the OS; whether that engine is compiled in or packaged as a non optional plugin is then a build detail.

Sub-part scripts and script parts: this is where JS modules help. Handlers registered with `item.on("Hit", fn)` are bound to an object instance, not to a global name, and state lives in a factory or class instance, so the same asset with its script can be imported twice without name or event conflicts. Shared code carried inside the table as separate script parts, the way [vpx-glf](https://github.com/mpcarr/vpx-glf) copies VBS modules into each table, maps onto `import` directly: modules next to the table already resolve, and serving embedded script parts is one more lookup in the loader. The `<Item>_<Event>` exports are the compatibility path for ported tables.

Status: JP's Party Zone runs on the branch with `core.js` and `wpc.js` ports, so the VPM loop, trough, saucers, fast flips and GI all work through the bridge.

Scripts folder: do you want to move towards fewer scripts, with the controller integration (lamp and solenoid updates, switch pulsing, fast flips, ball stacks) as native API, the ROM family files as data or descriptors, and table side code modules for reuse and deduplication, or do you see a one on one translation of what exists? #2263 collects the table-side part of the same question (rolling sounds, shadows, positional audio, DOF routing). The answer decides whether `core.js` stays the small stopgap it is now or becomes a full port.

Before this goes further I need three decisions, because the remaining work only makes sense for one language and one direction:

1. Is JavaScript acceptable as the second script language, or is Python the decision? If it is Python, I would rather stop here than build on a direction that gets dropped.
2. Is an in-tree engine behind the `IScriptEngine` interface acceptable now, moving to the plugin anchors when they land, or should this wait for them?
3. Scripts folder: fewer scripts, native API and data, table side code modules for reuse and deduplication, or a one on one translation of what exists?

Also, is the roadmap written up somewhere it can be referenced? On the table format, [vpxtool](https://github.com/francisdb/vpxtool)'s extract/assemble folder layout (json plus assets) exists and is in use by table authors; it would be worth aligning the vpz format with it rather than ending up with two.

vbousquet:
After thinking this thoroughly, here are my feeling about this change:
- I share that adding other languages is an interesting feature, but this is a bonus, not a core need,
- adding them inside the core would downgrade the overall core quality and increase its maintainability effort, as this will add to VBS that we must keep to support anyway for the 9000+ existing tables, in turn making everything heavier,
- hence, as pointed in my previous post, I think the aim is to push VBS outside and not add additional languages inside (also considering that they will likely come with there core scripts)
- considering the previous points, the language choice is not the core point at the moment (I guess every coder has his own taste, beside the fact that AI makes the language choice less important nowadays that everyone use AI to code in any of them)
- adding new language should also fit into the broader import/export arch, still to be designed (which need a way to define script in an imported subpart with connected events), or we would had an heavy piece which we would have to redesign just afterward to fit in, eventually creating additional legacy content to support in a backward manner
- the timing is also quite not favorable as we have a nearly finished release, so a large change like this one should be avoided at the moment.

So, I think the best decision is to proceed with the release, stabilize the API and build the plugin anchor, design the way mutiple scripts are integrated in a table (eventually with a language property) before considering adding new languages to VPX.

francisdb:
I understand, closing this for now

PR #3962, opened 2026-09-14
standalone: close the player on SIGINT instead of exiting from the signal handler

francisdb:
Quitting a ROM table on Linux standalone with SIGINT (Ctrl+C, or a frontend stopping the table) crashed with a double free. The signal handler called `exit()`, which runs the static destructors of the whole process while the table is still running. libpinmame's static state broadcasts a controller change from its destructor, the Serum plugin answers it by calling `GetTableInfo`, and that function stores the path in a function-local static string that was already destroyed at that point.

The handler now asks the player to close, so plugins are unloaded and the emulation stopped before the process exits the normal way. A signal repeated after a few seconds still exits immediately, for the case where the shutdown itself is stuck.

The macOS handler in `standalone/macos/main.mm` has the same `exit()` call but cannot see the player from there; I left it as is.


francisdb:
@jsm174 you might be interested for macos

gitfool:
I've hit this when trying to automate testing.

jsm174:
@francisdb - Thanks for this. I updated for macos as well!

